PR #18, opened 2022-07-03
Slightly improve installation process

gahdritz:
Quick PR addressing two minor issues I encountered while installing `flash-attention`:

1. Adds the `einops` dependency to `setup.py` such that it's installed automatically along with the package.
2. Adds `__init__.py` to the `flash_attn` package such that `setup.py` properly detects it. Before, `flash_attn` could not be imported from outside of the `flash-attention` directory, even after the `setup.py` installation.


tridao:
This looks great, thanks Gustaf!

PR #29, opened 2022-08-03
test flash attention with model

lw921014:
hi, Tri Tao,

As we mentioned in [issue](https://github.com/HazyResearch/flash-attention/issues/26). I find if I use your [test](https://github.com/HazyResearch/flash-attention/blob/main/tests/test_flash_attn.py) to check the precise, I found that precise loss is so small. But this test is still located in kernel level. We need to add some test in model level. This pr is what I do to check precise loss with a model layer. **We find it still have some numerical error which is larger than 3% as shown  following.**

- About model layer

here we construct a layer similar to that in [swin-T](https://github.com/microsoft/Swin-Transformer/blob/main/models/swin_transformer.py#L77) as our ref model. This is what tests/tools/window_attention.py do. Here, we have  WindowAttention and  WindowAttentionFP16, WindowAttention is for apex(case 1), WindowAttentionFP16 is for not using apex(case2). These two model is very similar, the only difference is the linear layer weght and bias type . Tested model is our [fmha layer](https://github.com/HazyResearch/flash-attention/blob/main/flash_attn/flash_attention.py#L80). 

- about test method

Due to our kernel is all fp16, so I use two method to get fp16 data, 
case1:  is using apex, this is what tests/test_with_model_apex.py do, here all the weight and bias are float, apex will chenge it into fp16 automantically. 
case2: is not using apex, we set all weight and bias as fp16 type. This is what tests/test_with_model_no_apex.py do.

- about comparison method
this is what tests/tools/check_tool.py do. **The major function is is_same_matrix, it will compare two np.array, `abs_eps = 0.01, relative_rps = 0.03` means abs_eps = abs(a-b), relative_rps = abs(a-b) / abs(a). For each ele in a matrix, if the abs_eps of all ele less than 0.01 and the relative_rps of all ele less than 0.03, this test will pass, otherwise failed and output the pos which not satisfied the requirement.**  In tests/test_with_model_no_apex.py and tests/test_with_model_apex.py, we test the output and all weight and bias dgrad using this function.

- about run method
case 1: `python test_with_model_apex.py`
case 2: `python test_with_model_no_apex.py`

note: we test fwd and bwd both, output means the fwd output, dgrad mens the grad for all the parameters, here we have four parameter, it is qkv weight and bias, proj weight and bias.

- about test result

1. case 1: 

output 
image
dgrad
image

1. case 2:

output
image
dgrad
image


flash_attn/flash_attention.py
```diff
@@ -103,6 +103,15 @@ def __init__(self, embed_dim, num_heads, bias=True, batch_first=True, attention_
...
 
+        if 'qkv_weight' in init_para.keys():
+            self.Wqkv.weight = torch.nn.Parameter(init_para["qkv_weight"])
+        if 'qkv_bias' in init_para.keys():
+            self.Wqkv.bias = torch.nn.Parameter(init_para["qkv_bias"])
+        if 'proj_weight' in init_para.keys():
+            self.out_proj.weight = torch.nn.Parameter(init_para["proj_weight"])
+        if 'proj_bias' in init_para.keys():
+            self.out_proj.bias = torch.nn.Parameter(init_para["proj_bias"])
+
```
vchiley, line 114: wouldn't [loading a state_dict](https://pytorch.org/tutorials/beginner/saving_loading_models.html#save-load-state-dict-recommended) be a better solution?

PR #53, opened 2022-10-05
build wheel workflow

robotcator:


robotcator:
Hi, Tridao, I created a workflow to build a wheel for pip installation. We don't need to compile from scratch.

tridao:
Thanks, this is super helpful!

PR #55, opened 2022-10-06
remove numpy dependency

ajfadam:
According to the `setup.py` file, only dependencies are torch and einops. But the `bert_padding.py` file requires `numpy` only to multiply the elements of a `torch.Size` object. This change aims at allowing the use of FlashAttention without numpy.


tridao:
Thanks! I learned a new trick of calling `.numel()` on the shape 😃

PR #75, opened 2022-11-15
allow for uploading to pypi

lucidrains:
should allow you to upload this to pypi

```bash
$ pip install twine
```

then

```bash
$ make upload_package
```

it will then ask for username and password, which you need to acquire by making an account at https://pypi.org/

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,21 @@
+# Byte-compiled / optimized / DLL files
+__pycache__/
+*.py[cod]
+
+# C extensions
+*.so
+
+# Distribution / packaging
+bin/
+build/
+develop-eggs/
+dist/
+eggs/
+lib/
+lib64/
+parts/
+sdist/
+var/
+*.egg-info/
+.installed.cfg
+*.egg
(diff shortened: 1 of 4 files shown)


tridao:
Thank you Phil!

robotcator:
Hi, @lucidrains, can this distinguish which Torch version and which Cuda version? It seems that the extension will compile with this setting.

tridao:
I'm not certain but  I think `pip install flash-attn` is equivalent to cloning the repo and then run `python setup.py install`. So it would compile locally and use the local torch and cuda version?

lucidrains:
@robotcator yup, what Tri said

may also throw in a check at some point to make sure it only compiles for the correct graphic cards (non-volta)

robotcator:
@tridao @lucidrains I see, it still builds locally not uploading the builded wheel package to `pypi-server`.

Previously, I created the workflow to build the wheel package and then uploading to release assert. Due to the `name format`, the built wheel can not upload to the 'pypi-server'

PR #92, opened 2022-12-16
Fix slicing dimensions in rotary embeddings

ploshkin:


flash_attn/layers/rotary.py
```diff
@@ -43,13 +43,11 @@ def forward(ctx, x, cos, sin, inplace=False):
         rotary_dim *= 2
         assert rotary_dim <= headdim
         assert seqlen <= rotary_seqlen
-        assert cos.shape == (rotary_seqlen, rotary_dim // 2)
-        assert sin.shape == (rotary_seqlen, rotary_dim // 2)
```
tridao, line 47: Can we keep the shape checking for sin?

flash_attn/layers/rotary.py
```diff
@@ -92,14 +90,12 @@ def forward(ctx, qkv, cos, sin):
         rotary_dim *= 2
         assert rotary_dim <= headdim
         assert seqlen <= rotary_seqlen
-        assert cos.shape == (seqlen, rotary_dim // 2)
-        assert sin.shape == (seqlen, rotary_dim // 2)
```
tridao, line 96: Can we keep the shape checking for sin?

tridao (changes requested):
Thanks for the fix @ploshkin!

flash_attn/layers/rotary.py
ploshkin: Ah, sure

flash_attn/layers/rotary.py
ploshkin: Here I changed the assertion because length of sin might be greater than `seqlen` in general

tridao: Oh right, that's a great point.

tridao (approved):
Perfect, thanks again @ploshkin!

PR #95, opened 2022-12-28
Add gpt-neox adoption

Quentin-Anthony:
Now that flash attention is being used to train all future gpt-neox models, we should display it as an example adoption framework.


tridao:
Thanks @Quentin-Anthony! Much appreciated!

PR #140, opened 2023-03-15
Remove unused kwargs like device in FlashAttention

VikParuchuri:
I noticed that there were kwargs specified in the FlashAttention implementation to mirror PyTorch (device, dtype), but they weren't used.  This is a little confusing when you're first reading the code.

I can see an argument for leaving them in (to mirror PyTorch), but I can also see an argument for removing them, so they don't mislead users into thinking they do something.

I modified FlashMHA to not pass in the factory kwargs.  I removed the kwargs argument in this class as well, since it wasn't being used.

I'll leave it to you to decide if this makes sense or not.

### DIFF
--- a/flash_attn/flash_attention.py
+++ b/flash_attn/flash_attention.py
@@ -18,7 +18,7 @@ class FlashAttention(nn.Module):
         attention_dropout: The dropout rate to apply to the attention
                            (default: 0.0)
     """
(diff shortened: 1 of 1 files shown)


tridao:
This is a good idea, thanks @VikParuchuri!

PR #147, opened 2023-03-31
Add option for deterministic execution

ksivaman:
The `_flash_attn_backward` function is non-deterministic as noted [here](https://github.com/HazyResearch/flash-attention/issues/73#issuecomment-1312814265) previously, however setting `num_splits=1` ensures deterministic behavior at the cost of some performance. This PR adds a boolean argument (`deterministic`) to the Python APIs to optionally enable this. The default is still `False` in order to keep the current behavior and perf unchanged.

I have verified that all tests under `tests/test_flash_attn.py` are passing.


tridao:
Thanks @ksivaman!

PR #166, opened 2023-04-13
Enable CUDA graph capture

### LINKED ISSUE #165: CUDA graph capture not supported
Repro script:

```
import torch
from flash_attn.flash_attn_interface import flash_attn_unpadded_func

seq_len, batch_size, nheads, embed = 2048, 2, 12, 64
dtype = torch.float16
pdrop = 0.1

q, k, v = [torch.randn(seq_len*batch_size, nheads, embed, dtype=dtype, device="cuda") for _ in range(3)]
cu_seqlens = torch.arange(
    0,
    (batch_size + 1) * seq_len,
    step=seq_len,
    dtype=torch.int32,
    device=q.device)

output = None

def run_flash_attn():
    output = flash_attn_unpadded_func(q, k, v, cu_seqlens, cu_seqlens, seq_len, seq_len, pdrop)

# Pre graph capture warmup in a separate stream.
s = torch.cuda.Stream()
with torch.cuda.stream(s):
    for _ in range(3):
        run_flash_attn()
torch.cuda.current_stream().wait_stream(s)

# Capture.
g = torch.cuda.CUDAGraph()
with torch.cuda.graph(g):
    run_flash_attn()

# Run graph.
g.replay()
```

Fails with the error:

```
RuntimeError: Cannot call CUDAGeneratorImpl::current_seed during CUDA graph capture. If you need this call to be captured, please file an issue. Current cudaStreamCaptureStatus: cudaStreamCaptureStatusActive
```

This is because getting and setting CUDA RNG states is not allowed during graph capture and thus calls to `torch.cuda.get_rng_state` and `torch.cuda.set_rng_state` cause the failure.

ksivaman:
Fixes #165 

To ensure that the same dropout mask is drawn during backprop as in the fprop execution, currently, the cuda RNG state is saved during the forward pass and then reset during backward. This is done in PyTorch via the `torch.cuda.{get|set}_rng_state` APIs. However, this breaks graph execution since querying or setting the internal seed is prohibited, along with other CPU ops. To bypass these calls while maintaining correct behavior (and enable graph execution), we can save the seed and offset values as GPU tensors in the forward kernels after unpacking the philox RNG state. These can be reused in the backward pass to correctly initialize the `Philox` class with the correct seed and offset without needing to reset the RNG or calling `philox_cuda_state` on the generator (during bprop).

I have tested that this method of resetting RNG state yields bitwise identical results compared to the previous method while also allowing graph capture. I have also verified that all tests under `tests/test_flash_attn.py` are passing.

### DIFF
--- a/csrc/flash_attn/src/fmha_fprop_kernel_1xN.h
+++ b/csrc/flash_attn/src/fmha_fprop_kernel_1xN.h
@@ -678,6 +678,8 @@ inline __device__ void device_1xN_loop(const Params &params) {
     // the attention matrix. This way, as long as we have the batch, head, and the location of
     // the 16 x 16 block within the attention matrix, we can generate the exact same dropout pattern.
     auto seeds = at::cuda::philox::unpack(params.philox_args);
+    params.rng_state[0] = std::get<0>(seeds);
+    params.rng_state[1] = std::get<1>(seeds);
     Philox ph(std::get<0>(seeds), 0, std::get<1>(seeds) + (bidb * params.h + bidh) * 32 + tidx % 32);
     constexpr int M = Kernel_traits::Cta_tile_p::M;
     const int STEPS = (params.seqlen_q + M - 1) / M;
--- a/csrc/flash_attn/src/fmha_fprop_kernel_1xN.h
+++ b/csrc/flash_attn/src/fmha_fprop_kernel_1xN.h
@@ -667,6 +667,8 @@ inline __device__ void device_1xN_loop(const Params &params) {
     const int bidb = blockIdx.x;
     // The block index for the head.
     const int bidh = blockIdx.y;
+    // The block index.
+    const int bidx = gridDim.x * bidh + bidb;
     // The thread index.
     const int tidx = threadIdx.x;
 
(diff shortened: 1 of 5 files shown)


csrc/flash_attn/src/fmha_fprop_kernel_1xN.h
```diff
@@ -678,6 +678,8 @@ inline __device__ void device_1xN_loop(const Params &params) {
     // the attention matrix. This way, as long as we have the batch, head, and the location of
     // the 16 x 16 block within the attention matrix, we can generate the exact same dropout pattern.
     auto seeds = at::cuda::philox::unpack(params.philox_args);
+    params.rng_state[0] = std::get<0>(seeds);
```
tridao: Should only 1 thread write this to global memory (e.g thread 0 of block (0, 0, 0))?

tridao (commented):
This looks great, thanks @ksivaman!
I just have a small question on how many threads should be writing the RNG seeds to global mem.

csrc/flash_attn/src/fmha_fprop_kernel_1xN.h
ksivaman: You're right that would be better. Made the changes and retested!

tridao:
Thanks @ksivaman!

PR #193, opened 2023-04-27
Use pyproject.toml to specify build dependencies

anthonyhu:
This PR introduces a `pyproject.toml` file to list the build dependencies (see https://snarky.ca/what-the-heck-is-pyproject-toml/).

Before, the build dependencies were listed in `setup.py`, but in order to read this file, you would need to have those dependencies installed (packaging, torch). By adding a `pyproject.toml` file, python knows which packages are needed to build flash-attention.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ pip install flash-attn
 
 Alternatively you can compile from source:
 ```
-python setup.py install
+python -m pip install .
 ```
 
 Interface: `src/flash_attention.py`
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["packaging", "setuptools", "torch", "wheel"]
+build-backend = "setuptools.build_meta"


ntoxeg:
I can confirm it fixes #209. Please merge ASAP.

tridao:
I had to remove pyproject.toml for now since I couldn't find a way to add torch as a build dependency that work for everyone.

kklemon:
@tridao That's really unfortunate. Removing the `pyproject.toml` breaks the PEP517 compliance and thus Poetry compatibility.

Handling the PyTorch dependency with that build system is indeed not trivial. In my own projects, I use one of the following solutions:

1. **Just don't declare torch as a dependency and let the user install the preferred version.** Probably only works for small projects where this is reasonable.
2. **Use a custom source definition with the `pyproject.toml` and point to a specific torch version.** This is my preferred solution at the moment. It does not solve the problem of exact CUDA version matching, but should still work for most users. If the user overwrites the torch version, e.g. with a specific CUDA version, the Poetry dependency resolver will typically respect it if the `pyproject.toml's` version declaration is still met.

For the second solution, an exemplary dependency configuration would look as follows:

```toml
[tool.poetry.dependencies]
torch = {version = "^2.0.1", source = "pytorch-gpu-src"}

[[tool.poetry.source]]
name = "pytorch-gpu-src"
url = "https://download.pytorch.org/whl/cu118"
priority = "explicit"
```

PR #202, opened 2023-05-06
[BugFix] avoid bug on ImportError

fedebotu:
Simple fixes to avoid `cannot unpack non-iterable NoneType object` upon `ImportError` - bug found during a debugging session with @massastrello


tridao:
Thanks @fedebotu and @massastrello!

PR #241, opened 2023-05-25
Fix compilation time

ksivaman:
There is a significant bump in compilation time going from `v1.0.4` &rarr; `v1.0.5`. It was ~2-3 minutes previously and now ranges from ~17-40 minutes. I did a bisect and narrowed it down to extra dependencies being required in the `pyproject.toml` file which are already set in the `install_requires` section during setup. I've left the `packaging` library as is in `[build-system]` since it's used by `setup.py` itself.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
-requires = ["ninja", "packaging", "setuptools", "torch", "wheel"]
+requires = ["packaging", "setuptools", "torch", "wheel"]
 build-backend = "setuptools.build_meta"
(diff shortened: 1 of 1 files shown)


tridao:
Thanks @ksivaman.
I don't have much experience with compiling and packaging. I thought adding `ninja` in `pyproject.toml` would help compilation time. I think `ninja` is already in the nvidia pytorch container, but some users report that if their setup doesn't already have `ninja` installed, compilation is much slower. 
For your case do you think it's because `ninja` is already there and somehow a different version of `ninja` is downloaded if we put it in `pyproject.toml`?

ksivaman:
The issue in this case is not with `ninja` but actually `torch`. I have re-adding ninja since it **does** indeed speedup compilation if not already installed! @tridao

PR #243, opened 2023-05-27
bump to v1.0.6

ksivaman:
**Request** to bump version. Grabbing the current version from pip results in very slow compilation (see #241).

If this is not the best way to request and raising an issue is preferred, please LMK.


tridao:
Thanks @ksivaman. I've uploaded to PyPI as well.

PR #266, opened 2023-06-08
Add initial support for prebuilt wheels

piercefreeman:
As per the discussion in https://github.com/HazyResearch/flash-attention/issues/223, each flash-attention build currently has to start from scratch on install. In the best case with parallelism this might happen in a few minutes but worse case can take 30min+ on a lower powered machine. It's preferable to build the most common wheels once and download them when required.

For conventional wheels, pip uses file conventions that condition on architecture (arm vs. amd) and operating system. If it finds a match in the pypi index, it downloads and attempts to install it. Flash attention doesn't fit this paradigm because of its dependencies on system dependencies in the form of CUDA and torch versions.

Instead, I've taken the following approach:

1. Host built wheels for each supported cuda/torch/platform combination through github releases. The existing flash-attention CI system does something similar to this, but has been broken for the last several releases. And even then it didn't appear to be geared to comprehensively build for all torch and architecture versions. When fixed and upgraded this pipeline results in a release page [like this](https://github.com/piercefreeman/flash-attention/releases/tag/v1.0.8) where there is a central release with each installable prewheel.
2. Add a custom `_bdist_wheel` installation hook to the setup.py install script. This function determines what the expected wheel should be given the current operating system, checks if it's available at the remote github path, and swaps it into the position where pypi expects the local wheel. This effectively circumvents each client from having to build it separately while otherwise plugging and playing with pip's installation system.

As a proof of concept, I turned my fork into an installable pip package. You can try it out on a supported architecture by:

```sh
pip install "torch==1.13.0"
pip install flash-attn-wheels --no-build-isolation
```

You can also view some of the verbose logging via:

```sh
pip install flash-attn-wheels --no-build-isolation --verbose
```

This PR should be largely ready to merge but needs to swap out the hard-coded variables that make it work on my branch. Once the approach is given a thumbs up and fork development ceases, I'll make these changes.

_A few asides for the future reference:_

- I also investigated some other approaches for these custom wheels. The most promising was to write a separate build-backend as an alternative to setup tools. Since the introduction of pyproject.toml configurations, it's possible to specify different backend builders for each project (poetry does this for instance). It seemed like it would be reasonable to write a generic backend that does the same logic as _bdist_wheel but allows arbitrary clients to publish their platform wheels on github. This didn't quite work out in practice because the backends are supposed to be stateless of the actual build files, and we needed to extract more metadata like the cuda / torch versions that best live in setup.py
- Current Github Action limits dictate that a job matrix can generate a maximum of 256 jobs per workflow run. This limit applies to both GitHub-hosted and self-hosted runners. Since the build matrix is 4-dimensional, I can see this limit being reached at somepoint in the midterm. If this happens the easiest workaround might be to build a base image with a particular OS and cuda version, then iteratively install new python-versions and torch versions while building wheels in sequence. This will cut back on runners at the expense of longer build times - but is probably a worthwhile tradeoff considering building cuda version is the bulk of the environment preparation time.


.github/workflows/publish.yml
```diff
@@ -27,23 +30,40 @@ jobs:
...
-          os: [ubuntu-18.04]
+          os: [ubuntu-20.04, ubuntu-22.04]
           python-version: ['3.7', '3.8', '3.9', '3.10']
-          torch-version: [1.11.0, 1.12.0, 1.12.1]
-          cuda-version: ['113', '116']
+          torch-version: ['1.11.0', '1.12.0', '1.13.0']
+          cuda-version: ['113', '116', '117']
           exclude:
-            - torch-version: 1.11.0
+            # Nvidia only supports 11.7+ for ubuntu-22.04
```
piercefreeman, line 52: These explicit exclude cases seem to be the best method that github actions supports today to deal with these cuda-torch-os compatibility issues. However I could see this getting pretty messy in the future as the dependencies grow in complexity. A better approach would be some boolean relationships to define the matrix - but this would have to be implemented as some sort of external yml generator and adds to the complexity here a bit too much for an initial release.

Leaving it for future work.

.github/workflows/publish.yml
```diff
@@ -104,24 +124,60 @@ jobs:
...
         with:
           upload_url: ${{ steps.get_current_release.outputs.upload_url }}
-          asset_path: ./${{env.wheel_name}}
+          asset_path: ./dist/${{env.wheel_name}}
           asset_name: ${{env.wheel_name}}
-          asset_content_type: application/*
\ No newline at end of file
+          asset_content_type: application/*
+
+  publish_package:
```
piercefreeman: The new CI pipeline also deploys a package version to pypi once it builds successfully.

.github/workflows/publish.yml
```diff
@@ -104,24 +124,60 @@ jobs:
...
+
+      - name: Build core package
+        env:
+          FLASH_ATTENTION_SKIP_CUDA_BUILD: "TRUE"
+        run: |
+          python setup.py sdist --dist-dir=dist
+
+      - name: Deploy
+        env:
+          TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
```
piercefreeman: These will have to be updated to whatever tokens Hazy Research uses for your secrets.

setup.py
```diff
@@ -172,25 +201,88 @@ def get_package_version():
     else:
         return str(public_version)
 
+
+class CachedWheelsCommand(_bdist_wheel):
```
piercefreeman, line 205: This represents the bulk of the client side logic changes.

setup.py
```diff
@@ -90,77 +119,77 @@ def append_nvcc_threads(nvcc_extra_args):
...
-cc_flag = []
-_, bare_metal_version = get_cuda_bare_metal_version(CUDA_HOME)
-if bare_metal_version < Version("11.0"):
-    raise RuntimeError("FlashAttention is only supported on CUDA 11 and above")
-cc_flag.append("-gencode")
-cc_flag.append("arch=compute_75,code=sm_75")
-cc_flag.append("-gencode")
-cc_flag.append("arch=compute_80,code=sm_80")
-if bare_metal_version >= Version("11.8"):
+if not SKIP_CUDA_BUILD:
```
piercefreeman, line 125: There are a few environment variable flags that can get the setup.py side to revert to its previous behavior, which is useful in CI and I imagine some user edge cases.

piercefreeman:
@tridao Let me know if you have thoughts on the approach or need an env to test these changes.

tridao:
This is awesome, thanks so much @piercefreeman!
Sorry I've been busy with my PhD thesis.
I'm trying to figure out how to test this. Should I (temporarily) change the package name and push to PyPI to see if the wheels are workign?

piercefreeman:
@tridao Good question... I can update the code to make it easier to tag a `1.0.7.Xdev` version to pypi. This shouldn't be picked up by a regular "pip install" so it should be pretty safe to work on concurrently with other changes to main. I think it's still worth having CI drive the whole thing since a lot of the building need to happen before the pypi release.

On your end this should just require:
- Adding in pypi secrets
- Allowing this branch to run Github Actions

What do you think? Also happy to jump on a discord if this is easier to work through live.

tridao:
Thanks! I'm traveling now (graduation trip) but will have time in 1.5 weeks to test this.

piercefreeman:
@tridao Hope you had a nice trip! Were you able to give these changes a test?

tmm1:
Would love to see this pushed forward. I guess the first step would be to get these conflicts resolved?

tridao:
@piercefreeman Thanks again. Sorry I've been busy with graduation and v2 release. I'm fixing a few correctness issue with the v2 backward pass for some edge cases. Then I'll test the wheel building.

tmm1:
> I'm fixing a few correctness issue with the v2 backward pass for some edge cases.

hi @tridao is there an open issue about this or a link to details?

chigkim:
Any progress on this?
I don't know why, but the comman `pip install flash-attn-wheels --no-build-isolation` still builds it.
I just built on 1TB memory with 64 vcpu 8x a100 80gb, and it took about 25 minutes!
Every time when I run an instance without a persistent boot drive, that's 30 minutes of 8x a100 80gb idling.

piercefreeman:
@chigkim Can you share more about the specs on the environment you're using? What version of pytorch, what operating system, etc? The wheels are only built for specific versions so it's possible I'm not currently building the version on your box (but can likely easily add it).

chigkim:
I'm running an instance on Runpod:
Ubuntu 22.04.2 LTS
Kernel: 5.4.0-148-generic
Python: 3.10
Pytorch 2.0.1+cu118
Cuda 11.8

tridao:
> Any progress on this? I don't know why, but the command `pip install flash-attn-wheels --no-build-isolation` still builds it. I just built on 1TB memory with 64 vcpu 8x a100 80gb, and it took about 25 minutes! Every time when I run an instance without a persistent boot drive, that's 30 minutes of 8x a100 80gb idling.

It's ongoing work. I'm testing the wheels in a simplified repo to make sure it works.

piercefreeman:
@chigkim 2.0.1 / cuda118 are not currently supported in this PR. If you want to give the wheels a try you can try downgrading your torch version temporarily. See allowed matrix here:
https://github.com/Dao-AILab/flash-attention/pull/266/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7R44

I can take a crack at adding this config later this week, will just have to resolve some of the upstream conflicts first.

chigkim:
Thanks. Maybe I missed, but there's nothing for cu118?
I need to reboot after downgrading cuda, right?
Unfortunately Runpod Instances don't seem to have a persistent boot drive (only persistent data volume), so I can't reboot after downgrading cuda.

tridao:
I've been testing the wheel building. Pretty close to being done. I'm trying to figure out how to compile with CUDA 12 for torch 2.0 since there's no pre-built wheel for torch 2.0 with CUDA 12.

piercefreeman:
@tridao Is torch 2.0 & CUDA 12.0 exact versioning a requirement? Torch seems to be moving ahead with 2.1+12.1 support in their nightly releases:
https://download.pytorch.org/whl/nightly/cu121/torch_nightly.html

piercefreeman:
Hey @tridao, saw you merged into master. I think it picked up some of the tests I was doing around torch 2 & cuda12 (commits [565615c](https://github.com/Dao-AILab/flash-attention/pull/266/commits/565615c603bc83ff0215cf62bc4d907b27041215) through [6ef3bd8](https://github.com/Dao-AILab/flash-attention/pull/266/commits/6ef3bd800e8b8104537ffa0ba4ea10306da40f42)). You may want to revert back to [9af165c](https://github.com/Dao-AILab/flash-attention/pull/266/commits/9af165c38920bd18fc066e193383903e6ecff451) which was a more comprehensive version.

I can put up another PR for this as well; just let me know if torch2 / cuda 12 has to be in scope for the initial release.

tridao:
That's ok, I've been testing the wheel building (in a different repo) and I'll edit `publish.yaml` directly before tagging the next version (hopefully today) :D 
Thanks so much @piercefreeman! This has been a huge help. Not having CUDA wheels has been the biggest issue with the installation, and you helped unblocked this.

piercefreeman:
🎉 Happy to help and excited to see it live! Thanks for all the work on V2 and getting this to the finish line

PR #343, opened 2023-07-19
Fix compile error with `BOOL_SWITCH`

danthe3rd:
See also: https://stackoverflow.com/questions/55136414/constexpr-variable-captured-inside-lambda-loses-its-constexpr-ness

We had this sort of errors with nvcc 11.6:
```
$SRC_DIR/third_party/flash-attention/csrc/flash_attn/src/flash_fwd_launch_template.h(68): error: expression must have a constant value
Note #2767-D: the value of *this cannot be used as a constant
          detected during instantiation of "void run_mha_fwd_hdim64<T>(Flash_fwd_params &, cudaStream_t) [with T=cutlass::half_t]" 
$SRC_DIR/third_party/flash-attention/csrc/flash_attn/src/flash_fwd_hdim64_fp16_sm80.cu(25): here
```


tridao:
Thank you @danthe3rd!

PR #436, opened 2023-08-08
Support flash attention 2 with causal masking when KV's seq length is…

BoxiangW:
… longer than Q's seq length.

Support causal masking like:
111000
111100
111110
111111
where Q's seq length is 5 while KV's seq length is 6.

### DIFF
--- a/csrc/flash_attn/src/flash_bwd_kernel.h
+++ b/csrc/flash_attn/src/flash_bwd_kernel.h
@@ -638,7 +638,8 @@ inline __device__ void compute_dq_dk_dv_1colblock(const Params &params, const in
     tdQgdQaccum.data() = tdQgdQaccum.data() + kBlockM * params.d_rounded;
 
     int m_block = m_block_max - 1;
-    int m_block_min = !Is_causal ? 0 : (n_block * kBlockN) / kBlockM;
+    int m_block_min = !Is_causal ? 0 : (n_block * kBlockN - int(binfo.actual_seqlen_k - binfo.actual_seqlen_q)) / kBlockM;
+    m_block_min = m_block_min < 0 ? 0 : m_block_min;
 
     // We might need to exit early and write 0 to dK and dV.
     // Otherwise we get wrong result for the case where we don't enter the for loop.
@@ -799,7 +800,8 @@ inline __device__ void compute_dq_dk_dv_1colblock(const Params &params, const in
             // Putting this causal masking right after acc_s is *much* slower for some reason.
             if (m_block * kBlockM < (n_block + 1) * kBlockN) {
                 flash::apply_mask_causal(scores, n_block * kBlockN + (tidx / 32 / AtomLayoutMS) * MMA_N_SdP * 16,
-                                         binfo.actual_seqlen_k, m_block * kBlockM + get<0>(taccScS_row(0)),
+                                         binfo.actual_seqlen_q, binfo.actual_seqlen_k,
+                                         m_block * kBlockM + get<0>(taccScS_row(0)),
                                          // binfo.actual_seqlen_k, m_block * kBlockM + (tidx / 32) % AtomLayoutMS * 16 + (tidx % 32) / 4,
                                          AtomLayoutMS * 16);
             }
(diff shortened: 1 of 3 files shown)


tridao:
Thanks! This is only the forward pass right? 
I thought about it a bit and I think we'll need to be quite careful about correctness. There could be cases where we need causal masking on 2 blocks (currently because of the assumption of causal, we only need to check 1 block).
The other concern is performance: it should not affect performance for the common use case of causal attention when seqlen_q = seqlen_k.

BoxiangW:
I changed both forward pass and backward pass, and tried a few tests about the correctness. 
For the first concern are you talking about 2 blocks like 
|1110|0000|
|1111|1000|
    1       2     ?
I tested this situation and the result matrix was correct.
About the second concern, I think my implementation will not affect the performance since the algo was not affected.

sz128:
> … longer than Q's seq length.
> 
> Support causal masking like: 111000 111100 111110 111111 where Q's seq length is 5 while KV's seq length is 6.

Does it support arbitrary masking like:
[
100000
110000
001000
001100
000010
000011
,
100000
110000
111000
000100
000110
000111
]
, where batch_size=2, Q\K\V seq_length=6 ?

BoxiangW:
> > … longer than Q's seq length.
> 
> > 
> 
> > Support causal masking like: 111000 111100 111110 111111 where Q's seq length is 5 while KV's seq length is 6.
> 
> 
> 
> Does it support arbitrary masking like:
> 
> [
> 
> 100000
> 
> 110000
> 
> 001000
> 
> 001100
> 
> 000010
> 
> 000011
> 
> ,
> 
> 100000
> 
> 110000
> 
> 111000
> 
> 000100
> 
> 000110
> 
> 000111
> 
> ]
> 
> , where batch_size=2, Q\K\V seq_length=6 ?

For this PR no.

siddartha-RE:
Just wanted to let you know that I tested this branch against my version of the FastChat monkey patch. I confirmed it was working as expected:
https://github.com/siddartha-RE/FastChat.flashpatch/commit/6ad40e9f8aca571971859f480b7a5bc9183209c3

The change makes sense to me in that there is basically still one block that will be mixed (0 and 1) in terms of the causal mask. The position of the block just gets shifted because we have chopped off a set of trailing q vectors. It doesn't seem it impacts anything other than adjusting the calculation of where the cross over block falls.

tridao:
I'm working on this this week. I also want it to work for when seqlen_q > seqlen_k, so there quite a few edge cases.

tridao:
> For q_len > k_len and causal=True how will you handle the q positions that cannot see any of the k positions? For example the 0th position when q_len = k_len + 1. That will have to be a NaN output right?

Output will be 0 for that case.

 > Separately, just wanted to mention that I tested the change in the PR together with my change enabling flash_attn + use_cache = True on a real world dataset of 16K prompts. I saw an amazing speed up of at least 20x! We are very excited to get this into production.

That's some crazy speedup!

siddartha-RE:
> > For q_len > k_len and causal=True how will you handle the q positions that cannot see any of the k positions? For example the 0th position when q_len = k_len + 1. That will have to be a NaN output right?
> 
> Output will be 0 for that case.
> 

If thats the case (ie it will be constant 0) shouldn't it just be illegal to be q_len > k_len?

siddartha-RE:
> > > For q_len > k_len and causal=True how will you handle the q positions that cannot see any of the k positions? For example the 0th position when q_len = k_len + 1. That will have to be a NaN output right?
> > 
> > 
> > Output will be 0 for that case.
> 
> If thats the case (ie it will be constant 0) shouldn't it just be illegal to be q_len > k_len?

Meaning If is_causal then q_len > k_len is illegal since the output for [0, q_len - k_len] == 0.

TsingWei:
Does it support blocked causal mask like:
110000
111100
111111,
where tokens are grouped in causal?

tridao:
No it doesn't. What's the use case?

TsingWei:
Just an idea suddenly came to mind that, if we use a ViT encoder for a video sequence.  The tokens from one frame are in one group, and the ViT encoder can be combined with a group casual map for a video task.

jxjalsq:
why impl flash attention v2 , spilt q, k, v with block size =1,  it pricision is 10*e-2, the error is so big? when v is not fuse to onlinesftmax , is ok?

jxjalsq:
i impl flash attention in cpp code, and run in android, please resolve this problem， thank you.@

jxjalsq:
> No it doesn't. What's the use case?

hello

PR #446, opened 2023-08-14
enable loading hf llama checkpoints for training

lxuechen:


lxuechen:
@tridao This PR is ready to be reviewed. Aside, I have two questions:

- Have you tested the backward pass for llama with TP? I would like to run training soon. 
- How did you decide on some of the constants in the numerical tests? E.g., why is 3 is a reasonable number in this test`(logits - logits_ref).abs().max().item() < 3 * (logits_hf - logits_ref).abs().max().item()`? 
  - I'm asking mostly to learn about setting these tests, as I have become a little frustrated with numerical loss in half precision.

tridao:
> Do these numbers look reasonable to @tridao ?

Yes they look reasonable.

tridao:
> * Have you tested the backward pass for llama with TP? I would like to run training soon.

I didn't test the backward pass of llama specifically. I did test the backward pass of a generic GPT model with TP, to make sure the gradient is the same as if we didn't use TP (up to numerical precision).

> * How did you decide on some of the constants in the numerical tests? E.g., why is 3 is a reasonable number in this test`(logits - logits_ref).abs().max().item() < 3 * (logits_hf - logits_ref).abs().max().item()`?

No particular reason, I usually set this to 2 or 3. E.g. the definition of "reasonable numerical difference" is somewhat subjective. In my experience if two different implementations have numerical difference within a factor of 2 or 3 then they're most likely both doing correct things.

lxuechen:
> > * Have you tested the backward pass for llama with TP? I would like to run training soon.
> 
> I didn't test the backward pass of llama specifically. I did test the backward pass of a generic GPT model with TP, to make sure the gradient is the same as if we didn't use TP (up to numerical precision).

Thanks. Sounds like I should run a quick sanity check with llama just to make sure things are alright. 

> 
> > * How did you decide on some of the constants in the numerical tests? E.g., why is 3 is a reasonable number in this test`(logits - logits_ref).abs().max().item() < 3 * (logits_hf - logits_ref).abs().max().item()`?
> 
> No particular reason, I usually set this to 2 or 3. E.g. the definition of "reasonable numerical difference" is somewhat subjective. In my experience if two different implementations have numerical difference within a factor of 2 or 3 then they're most likely both doing correct things.

lxuechen:
For context all tests with the hugging face checkpoint format pass. I’m running a final sweep with the meta format checkpoints.

lxuechen:
> For context all tests with the hugging face checkpoint format pass. I’m running a final sweep with the meta format checkpoints.

I installed all the fused kernels, and now all tests pass.

tridao:
Thanks so much @lxuechen!

PR #480, opened 2023-08-24
enable checkpointing for gpt model def'n

lxuechen:


lxuechen:
Initial benchmarking shows that gradient checkpointing doesn't bring noticeable differences when flash-attention is enabled. This makes sense as flash-attention does activation recomputation.

tmm1:
How do you mean? Generally checkpoints will trade slower speed for much less memory usage, no?

tmm1:
I am interested in checkpointing support to reduce VRAM requirements. But I don't quite understand how this PR works. Is there something using `checkpointing` kwarg already?

tmm1:
This change was not enough for me to use with HF trainer:

>AttributeError: 'GPTLMHeadModel' object has no attribute 'gradient_checkpointing_enable'

lxuechen:
> This change was not enough for me to use with HF trainer:
> 
> > AttributeError: 'GPTLMHeadModel' object has no attribute 'gradient_checkpointing_enable'

This is because trainer calls `model.gradient_checkpointing_enable`, but `gradient_checkpointing_enable` is function of Tri's model definition. 

To answer your question above as to why I don't expect setting `checkpointing = True` further saves memory when flash-attention is turned on: This is because Tri is checkpointing the attention portion, but FlashAttention already avoids storing activations in the attention block and performs backwards recomputation (this is documented in [the original paper](https://arxiv.org/pdf/2205.14135.pdf)).

PR #486, opened 2023-08-25
minor model compat issues with hf trainer

tmm1:


flash_attn/models/gpt.py
```diff
@@ -574,6 +574,7 @@ def __init__(self, config: GPT2Config, process_group=None, device=None, dtype=No
         super().__init__(config)
         self.process_group = process_group
         self.transformer = GPTModel(config, process_group=process_group, **factory_kwargs)
+        self.device = torch.device(device)
```
tridao: What happens if we construct the model and then do `mode = model.to('cuda')`? Idk how hf keeps track of self.device?

tmm1: Good question. Looks like it uses this:

https://github.com/huggingface/transformers/blob/74081cb5fa52540bbdde620942bd3a657af85c8e/src/transformers/modeling_utils.py#L821

Maybe we can just re-use that mix-in.

I'm thinking instead of modifying `GPTLMHeadModel`, I should subclass it into a HF compat class that adds the required methods.

tridao: That sounds better to me. I personally don't use HF trainer. I didn't realize HF trainer doesn't just take an nn.Module but requires the class to implement specific methods.

tmm1:
I found this in the HF docs: https://huggingface.co/docs/transformers/main_classes/trainer

IMG_5423

lxuechen:
> I found this in the HF docs: https://huggingface.co/docs/transformers/main_classes/trainer
> 
> IMG_5423
> 
> I will rework this PR based on this and the other mixins.

happy to take a look if you need another pair of eyes when the rework is done!

tmm1:
I created a subclass. It was pretty simple.

However I noticed that `transformers.modeling_utils.PreTrainedModel` has a lot of stuff, specifically `resize_token_embeddings` and all its related code. Maybe I could setup multiple-inheritance and get it to work, but I'm not sure if its worth the hassle.

PR #490, opened 2023-08-25
Support MQA + MP for decoding

danthe3rd:


flash_attn/modules/mha.py
```diff
@@ -736,7 +736,7 @@ def __init__(
             self.num_heads, self.world_size, self.local_rank
         )
         self.num_heads_kv_per_rank = get_dim_for_local_rank(
-            self.num_heads, self.world_size, self.local_rank
```
lxuechen, line 738: this was a bug i wrote. thanks for fixing this!

flash_attn/models/gpt.py
```diff
@@ -628,18 +628,24 @@ def forward(self, input_ids, position_ids=None, inference_params=None, last_toke
         last_token_only: whether to return the logit for the last token only,
             of shape (batch_size, vocab_size)
         """
+        assert input_ids.ndim == 2
```
lxuechen: nit: since we're enforcing this, could we put the expected dimension in the docstring? also would be useful to have an assertion message.

flash_attn/models/gpt.py
```diff
@@ -628,18 +628,24 @@ def forward(self, input_ids, position_ids=None, inference_params=None, last_toke
...
+        b, slen = input_ids.shape
         hidden_states = self.transformer(
             input_ids, position_ids=position_ids, inference_params=inference_params
         )
-        if last_token_only:
-            hidden_states = hidden_states[:, -1]
         if self.project_out is not None:
             hidden_states = self.project_out(hidden_states)
+        # shape is either (b*slen/N, ft) in seqparallel
```
lxuechen: nit: if `self.project_out` is not None and `last_token_only` is True, this structure might waste some compute for projecting the un-returned tokens. any chance we could move the hidden state reshape and extraction to before project_out?

flash_attn/models/gpt.py
```diff
@@ -824,14 +832,6 @@ def combine_qkv_headdim(state_dicts, state_dict, key):
                 ]
                 state_dict[key] = rearrange(torch.cat(xs, dim=1), "three d ... -> (three d) ...")
             else:
-                xs = [
```
lxuechen, line 826: sorry i was a bit sloppy when i refactored this part. partially this was due to there not being a test. 

would you mind adding a test for the GQA and MQA case?

lxuechen: e.g., i had a test that i ran for standard MHA which i never pushed, which 1) shards a state_dict, 2) merges the shards back, and 3) compares the original vs recovered.

lxuechen (commented):
per discussions with @tridao, i'm helping out for some of these community contributions mostly through commenting for now. 

great fixes in this PR!

flash_attn/models/gpt.py
danthe3rd: I reverted the code to what it was previously. I believe the right fix is to disable sequence parallel during generation - I added an assert

flash_attn/models/gpt.py
danthe3rd: added test :)

lxuechen (approved):
awesome! thanks a lot again! really excited about this PR enables and trying out code-llama fine-tuning with TP soon! @tridao i think this PR is ready to merge.

PR #488, opened 2023-08-25
[ft_attention] Fix for seqlen=8136

danthe3rd:
When seqlen=8136, `smem_sz = 48840`, and apparently starting the kernel returns an `invalid argument` CUDA error.

`48840 < 48 * 1024` but apparently it's still above the limit somehow..? Tested on A100

PR #491, opened 2023-08-25
[decoding] Support LLaMa2 and CodeLLaMa

danthe3rd:
Requires the other PRs to work well:
https://github.com/Dao-AILab/flash-attention/pull/490
https://github.com/Dao-AILab/flash-attention/pull/488

LLaMa2 notably adds MQA

### DIFF
--- a/flash_attn/models/llama.py
+++ b/flash_attn/models/llama.py
@@ -10,6 +10,7 @@ from typing import Union
 
 import torch
 import torch.nn.functional as F
+from sentencepiece import SentencePieceProcessor
 from transformers import GPT2Config, LlamaConfig
 
 
@@ -308,7 +309,30 @@ def config_from_meta_checkpoint(
         num_attention_heads=params["n_heads"],
         num_hidden_layers=params["n_layers"],
         rms_norm_eps=params["norm_eps"],
+        num_key_value_heads=params.get("n_kv_heads", None),
     )
+    multiple_of = params.get("multiple_of", 1)
+    ffn_dim_multiplier = params.get("ffn_dim_multiplier", None)
+
+    # Compute the hidden dimension of the MLP
+    # https://github.com/facebookresearch/llama/blob/1a240688810f8036049e8da36b073f63d2ac552c/llama/model.py#L224
+    intermediate_size = 4 * config.hidden_size
+    # https://github.com/facebookresearch/llama/blob/1a240688810f8036049e8da36b073f63d2ac552c/llama/model.py#L195-L199
+    intermediate_size = int(2 * intermediate_size / 3)
+    # custom dim factor multiplier
+    if ffn_dim_multiplier is not None:
+        intermediate_size = int(ffn_dim_multiplier * intermediate_size)
+    intermediate_size = multiple_of * ((intermediate_size + multiple_of - 1) // multiple_of)
+
+    config.intermediate_size = intermediate_size
+    if "rope_theta" in params:
+        config.rotary_emb_base = params["rope_theta"]
+    config.vocab_size = 32000
+    # some CodeLLaMa have vocab_size 32000, some 32016
+    # Sadly it's not specified in the `params.json` file :(
+    tokenizer = Path(checkpoint_path) / model_name / "tokenizer.model"
+    if tokenizer.is_file():
+        config.vocab_size = SentencePieceProcessor(str(tokenizer)).vocab_size()
     return config
 
 
@@ -364,4 +388,6 @@ def llama_config_to_gpt2_config(llama_config: LlamaConfig) -> GPT2Config:
         out_proj_bias=False,
         mlp_fc1_bias=False,
         mlp_fc2_bias=False,
+        rotary_emb_base=getattr(llama_config, "rotary_emb_base", 10000.0),
+        n_head_kv=llama_config.num_key_value_heads,
     )


flash_attn/models/llama.py
```diff
@@ -308,7 +309,28 @@ def config_from_meta_checkpoint(
...
+    if ffn_dim_multiplier is not None:
+        intermediate_size = int(ffn_dim_multiplier * intermediate_size)
+    intermediate_size = multiple_of * ((intermediate_size + multiple_of - 1) // multiple_of)
+    config.intermediate_size = intermediate_size
+    if "rope_theta" in params:
+        config.rotary_emb_base = params["rope_theta"]
+    config.vocab_size = 32000
+    # CodeLLaMa
+    if "code" in model_name.lower():
+        # some have vocab_size 32000, some 32016
```
tmm1: Do you mean the vocab size varies by model size?

lxuechen: Code llama has a few more tokens, partly due to the [fill-in-the-middle](https://arxiv.org/abs/2207.14255) fine-tuning. In their paper:

> "We extend Llama 2’s tokenizer with four special tokens that mark the beginning of the prefix, the middle part or the suffix, and the end of the infilling span."

In addition, there's adjustments to the sentencepiece formulation.

flash_attn/models/llama.py
```diff
@@ -308,7 +309,28 @@ def config_from_meta_checkpoint(
...
+    intermediate_size = 4 * config.hidden_size
+    intermediate_size = int(2 * intermediate_size / 3)
+    # custom dim factor multiplier
+    if ffn_dim_multiplier is not None:
+        intermediate_size = int(ffn_dim_multiplier * intermediate_size)
+    intermediate_size = multiple_of * ((intermediate_size + multiple_of - 1) // multiple_of)
+    config.intermediate_size = intermediate_size
+    if "rope_theta" in params:
+        config.rotary_emb_base = params["rope_theta"]
+    config.vocab_size = 32000
```
lxuechen, line 330: being slightly pedantic here, but extracting the `vocab_size` based on the `model_name` can feel slightly ad-hoc. can we unify the two settings by always loading the `tokenizer.model` file and measuring the size of the resulting tokenizer? 

e.g., i might want to load another fine-tuned model from codellama, which has modelname `custom-my-model`, which doesn't contain the unigram `code` in the name.

flash_attn/models/llama.py
```diff
@@ -308,7 +309,28 @@ def config_from_meta_checkpoint(
         num_attention_heads=params["n_heads"],
         num_hidden_layers=params["n_layers"],
         rms_norm_eps=params["norm_eps"],
+        num_key_value_heads=params.get("n_kv_heads", None),
     )
+    multiple_of = params.get("multiple_of", 1)
+    ffn_dim_multiplier = params.get("ffn_dim_multiplier", None)
+
+    intermediate_size = 4 * config.hidden_size
```
lxuechen, line 319: this could be due to me missing context here, but could we get some clarity on this `intermediate_size` computation logic? 

fwiw, it seems this would not give the result i expect for the old 7b model (intermediate_size = )

flash_attn/models/llama.py
```diff
@@ -308,7 +309,28 @@ def config_from_meta_checkpoint(
...
+    intermediate_size = int(2 * intermediate_size / 3)
+    # custom dim factor multiplier
+    if ffn_dim_multiplier is not None:
+        intermediate_size = int(ffn_dim_multiplier * intermediate_size)
+    intermediate_size = multiple_of * ((intermediate_size + multiple_of - 1) // multiple_of)
```
lxuechen, line 325: this part is rounding `intermediate_size` to nearest multiple of `multiple_of`. could we do something like `intermediate_size = multiple_of * math.ceil(intermediate_size / multiple_of)`, and add a comment?

lxuechen (commented):
thanks for the great work! could you confirm if the old tests still pass?

flash_attn/models/llama.py
lxuechen: For clarity, I'm just saying there's no issues with this code in terms of being consistent with the recent Hugging Face integration.

lxuechen:
Hi @danthe3rd, would you mind following up with the comments? would also be great if you could apply the patches if that makes sense. i'm happy to help out here but i don't think i can edit your pr.

flash_attn/models/llama.py
danthe3rd: The vocab_size depends on the CodeLLaMa model yes. The 7b might have a different vocab size than the 34b for instance.

flash_attn/models/llama.py
danthe3rd: This is from the llama code release:
https://github.com/facebookresearch/llama/blob/1a240688810f8036049e8da36b073f63d2ac552c/llama/model.py#L224
https://github.com/facebookresearch/llama/blob/1a240688810f8036049e8da36b073f63d2ac552c/llama/model.py#L195-L199

I've added comments to the code

flash_attn/models/llama.py
danthe3rd: I've added a link to the original implementation, I believe it's good to keep the exact same code for readability?

flash_attn/models/llama.py
danthe3rd: Good point - I've changed the code. Note that the regular LLaMa models don't contain the tokenizer within the checkpoint folder, so we will naturally fallback to 32k which is correct.

lxuechen (approved):
great work! @tridao i think this one is also ready to merge. thanks again guys!

tridao:
Thanks @danthe3rd and @lxuechen

Issue #432, opened 2023-08-07
Will attention_mask be extended to 3D? (concatenate short samples for efficient training)


sz128:
It seems ```unpad_input``` only support 2-D attention_mask matrix, while it is also meaningful to support a 3-D attention_mask matrix (batch_size x seq_len x seq_len). 

There are several scenarios that we would like to use a 3-D attention_mask:
* graph-attention;
* _Concatenating short samples to reduce padding rate which can make training more efficient. [supported now]_

tridao:
Cool, thanks for the context. Attention mask will eventually be there (hopefully) but we have other priorities right now.
Btw we already support concatenating short samples to avoid unpadding (e.g. the "varlen" functions, see our BERT [implementation](https://github.com/Dao-AILab/flash-attention/blob/d30f2e1cd50185c98ed88c0684b4a603f15bee37/flash_attn/models/bert.py#L144)).

sz128:
Glad to hear that. For concatenating short samples, I would like to make real sequences in a batch to have closer lengths, since the training efficiency is dominated by the max length of the batch. I draw a figure to show the procedure.
<img width="943" alt="image" src="https://github.com/Dao-AILab/flash-attention/assets/16218783/c6b571e4-bd8d-4208-a54a-398cc4297eb9">
In my understanding, it is different with the "varlen" functions. Am I right?

tridao:
I see, that's different from varlen yeah.

lumosity4tpj:
In triton's code, it seems that we can use bias to represent attention mask if the attention mask we pass in is already a concatenated mask.

lumosity4tpj:
> In triton's code, it seems that we can use bias to represent attention mask if the attention mask we pass in is already a concatenated mask.

That is, when the flash_attn_func function is passed, the attention mask is similar to bias (with the same shape), and there are two possible cases for the spliced attention mask:
1. It is possible that the attention mask of this block is all the values that should be mask, so qk is all -inf when qk = qk * softmax_scale + bias, resulting in nan in the following calculation;
2. Due to the padding sequence, there may be cases where lse_i and m_i are -inf, resulting in the case of nan in the following calculation;

Solution:
The solution I tried was to convert -inf to the min value of float32 (like `torch.finfo(torch.float32).min=-3.40282e+38`) when in use, so that there is no such thing as `-inf - (-inf)=nan`, but `-inf - (min value) = -inf`:
[...]

PR #499, opened 2023-08-29
add unpad_input_for_concatenated_sequences

sz128:
A new `unpad_input_for_concatenated_sequences` function is created to replace `unpad_input` in `flash_attn/bert_padding.py`. 
It helps to support concatenating short samples in one for efficient training, which is discussed in https://github.com/Dao-AILab/flash-attention/issues/432#issuecomment-1668822286.

### DIFF
--- a/flash_attn/bert_padding.py
+++ b/flash_attn/bert_padding.py
@@ -124,46 +124,50 @@ def unpad_input(hidden_states, attention_mask):
 
 def unpad_input_for_concatenated_sequences(hidden_states, attention_mask_in_length):
     """
+    Supports concatenating short samples in one sequence. The attention_mask_in_length is utilized to mask other short samples. It helps efficient training of variant lengths-based samples (e.g., the supervised fine-tuning task in large language model).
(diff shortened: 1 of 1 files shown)


sz128:
It is discussed in https://github.com/Dao-AILab/flash-attention/issues/432.

flash_attn/bert_padding.py
```diff
@@ -122,6 +122,74 @@ def unpad_input(hidden_states, attention_mask):
     )
 
 
+def unpad_input_for_concatenated_sequences(hidden_states, attention_mask_in_length):
+    """
+    Arguments:
```
tridao, line 168: Can you add a reference to the github issue so that the motivation is clear?

sz128: Yes. It is mentioned in https://github.com/Dao-AILab/flash-attention/issues/432#issuecomment-1697022599. Is that ok?

tridao: I mean just adding a line in the code docstring that links to the github issue.
Sth like: "The motivation for this function is explained here: [link to github issue]".

tridao:
Thanks @sz128

PR #501, opened 2023-08-29
fix citation in README

jquesnelle:
Very minor, but the Flash Attention 2 citation was missing a closing brace on the title (resulting in an error when copy/pasting into a paper 🙂 )


tridao:
Thanks!

PR #502, opened 2023-08-30
[bugfix] handle_x not define when using checkpoint_lvl = 2

GGGGGGXY:
when using checkpoint_lvl=2, we all_gather_raw(x) without async_op=True. 
So we don't need to wait for handle. Just skip.


tridao:
Good catch, thanks!

PR #516, opened 2023-09-05
fix: add __init__.py at flash_attn.ops.triton

Kyeongpil:
flash_attn can not find ops.triton. I think this file is missing.


dittops:
Thanks @Kyeongpil. I was facing the same issue and this fixed it.

Birch-san:
I think this only solves the first problem. afterwards it would complain that ops.triton.rotary is not found.

what I did to get it working more completely, was instead of creating an `__init__.py`, I created the `ops/triton` folder and copied into it the scripts from here:  
https://github.com/Dao-AILab/flash-attention/tree/main/flash_attn/ops/triton

tridao:
Thanks! I've cut a release v2.2.0 (might take CI 9-10 hours to finish compiling all the CUDA wheels).

PR #527, opened 2023-09-08
Convert flash_attn BERT models back to HuggingFace format

kevinhu:
Adds an `inv_remap_state_dict` function that will convert flash_attn BERT models back to a HF-compatible state_dict


tridao:
Thanks @kevinhu!

PR #532, opened 2023-09-09
BigCode model converters

kevinhu:
Implements bidirectional conversion for the BigCode series of models (e.g. StarCoder, WizardLM) along with corresponding unit tests.

Also adds `gelu_pytorch_tanh` to the list of supported activation functions (see https://github.com/huggingface/transformers/blob/95b374952dc27d8511541d6f5a4e22c9ec11fb24/src/transformers/activations.py#L28-L46), since `gelu_new` and `gelu_fast` are equivalent to `nn.GELU(approximate='tanh')`


tridao:
This is great! Thanks @kevinhu!

PR #544, opened 2023-09-14
Support key padding masks in GPT models

kevinhu:
Implements key padding mask support for both flash and non-flash GPT models, along with tests.

Note that key padding masks will not work with rotary embeddings because the kernel does not support variable lengths.


sentialx:
@kevinhu Would PrefixLM be possible with your PR?

PR #546, opened 2023-09-15
Fix Llama GQA/MQA

kevinhu:
The current `permute` methods used in loading/unloading Llama weights do not map key weights correctly when MQA/GQA is used. This PR patches this bug and adds a corresponding test to validate.

### DIFF
--- a/flash_attn/models/llama.py
+++ b/flash_attn/models/llama.py
@@ -15,7 +15,7 @@ from transformers import GPT2Config, LlamaConfig
 
 
 def remap_state_dict_meta_llama(
-    state_dict: dict[str, torch.Tensor], config: GPT2Config
+    state_dict: dict[str, torch.Tensor], config: GPT2Config, multi_query: bool = False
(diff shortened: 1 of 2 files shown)


flash_attn/models/llama.py
```diff
@@ -15,7 +15,7 @@
 
 
 def remap_state_dict_meta_llama(
-    state_dict: dict[str, torch.Tensor], config: GPT2Config
+    state_dict: dict[str, torch.Tensor], config: GPT2Config, multi_query: bool = False
```
tridao, line 18: `multi_query` is an argument but it's not used in the function?

tridao: `multi_query` is an argument but it's not used in the function?

kevinhu: Fixed!

tridao:
Thanks @kevinhu, larger Llamas and CodeLlamas with MQA/GQA now works!

PR #563, opened 2023-09-21
Fix E1136

### LINKED ISSUE #560: PR #465 #446 breaks Python <= 3.8
## Description
In #465 and #446, subscription on `dict`/`list` are introduced in `state_dict: dict[str, torch.Tensor]` and `def state_dicts_from_checkpoint(checkpoint_path: Union[str, os.PathLike], model_name: str) -> list[dict]:`

However, this is a new feature in Python 3.9: [Type Hinting Generics in Standard Collections](https://docs.python.org/3/whatsnew/3.9.html#type-hinting-generics-in-standard-collections).

In Python 3.7 and 3.8, `TypeError` will be raised when encountering this file path.

## Repro
`python -c 'from flash_attn.models.llama import remap_state_dict_meta_llama'` gives:
```
  File "<string>", line 1, in <module>
  File "<REDACTED>/python3.8/site-packages/flash_attn/models/llama.py", line 18, in <module>
    state_dict: dict[str, torch.Tensor], config: GPT2Config
TypeError: 'type' object is not subscriptable
```

## Scope
This issue affects `flash-attn>=2.0.8` including `main` branch.

icyblade:
This PR will fix #560.

One can always identify this issue with: `pylint --disable=R,C,W,E,I --enable=E1136 flash_attn/` under Python <= 3.8.

Reference:
 - https://pylint.readthedocs.io/en/stable/user_guide/messages/error/unsubscriptable-object.html

### DIFF
--- a/flash_attn/models/gpt.py
+++ b/flash_attn/models/gpt.py
@@ -6,6 +6,7 @@ import re
 from collections import OrderedDict, namedtuple
 from collections.abc import Sequence
 from functools import partial
+from typing import Dict, List
 
 import torch
 import torch.nn as nn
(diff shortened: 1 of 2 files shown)


tridao:
Thanks!

PR #572, opened 2023-09-26
A100 FAv2 not working

ksivaman:
Running `import flash_attn` on A100 gives the following error:

```
ImportError: /usr/local/lib/python3.8/dist-packages/flash_attn_2_cuda.cpython-38-x86_64-linux-gnu.so: undefined symbol: _ZN3c104cuda9SetDeviceE
```

A bisect of the versions shows that [this](https://github.com/Dao-AILab/flash-attention/commit/f8dccfc90a115ce797fed7d5eaf8b066f2753d34) is the offending commit. I have confirmed that installing from source using the uploaded wheels (via `pip install .`) results in the error for all newer versions whereas locally building via `python setup.py install` works, indicating that this change results in erroneous wheels for ampere (h100 works, haven't looked at other devices).


tridao:
That's strange, I've been installing it on top of nvcr image (e.g. nvcr pytorch 23.09) and that works fine. Are you using a different image?

Syno8:
how to solve it ?

ksivaman:
@tridao Were the uploaded wheels modified since this PR? Seems to work now with the same setup

Issue #606, opened 2023-10-15
Request for Support of Flash Attention for Custom Attention Bias （Sample as Deberta）


fecet:
# Issue Description:

I would like to address a concern regarding the support for Flash Attention in deberta, specifically in relation to custom attention bias. Currently, deberta is considered a state-of-the-art model, especially for mid-scaled tasks with models of around 1 billion parameters. However, when it comes to the `DisentangledSelfAttention` mechanism in deberta, the attention_bias is not simply formed from the attention mask but is computed by adding relative position encodings to the query_state and key_state before applying them to the `attention_scores`(resulting from the dot product of `q@k.T`).

This computation requires exposing an interface for layer-wise and instance-wise attention_bias operations, which is not currently supported by most mainstream attention acceleration libraries.

# Issue Details:

- Model: deberta
- Attention Mechanism: DisentangledSelfAttention
- Concern: Lack of support for custom attention bias computation
- Expected Support: Interface for layer-wise and instance-wise attention_bias operations
Request:

I would like to inquire if you have any plans in the near future to support this type of computation for custom attention bias. This support is crucial for researchers and practitioners who wish to work with custom attention mechanisms in bert-like mask-lm models.

Thank you.

Issue #534, opened 2023-09-11
how to add bias to to attn


maidabu:
If original code is:
  ...
  attn = q @ k
  attn = attn.softmax(dim=-1)
  ...

now my code is:
  ...
  attn = q @ k
  attn = attn + bias
  attn = attn.softmax(dim=-1)
  ...
where the shape of bias is the same with attn, and the dtype of bias is float(or half).

How do I modify the compute_attn_1rowblock in the flash_fwd_kernel.h.
I tried to copy global bias to the acc_s, but never successfully.
Any one can help me.

tridao:
acc_s is stored in a permuted layout. E.g. thread 0 holds element at indices (0, 0), (0, 1), (0, 8), (0, 9), etc, since this is the layout of the accumulator returned by the mma instruction (tensor cores). 
The layout is documented by Nvidia [here](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#matrix-fragments-for-mma-m16n8k16-with-floating-point-type).

So you'd need to read the bias in such a way that each thread would have the bias at the indices corresponding to the indices of their acc_s.
The right way to do that is probably to use an intermediate buffer in shared memory: read from global memory to shared memory, then use the `ldmatrix` instruction to load from shared memory to registers of each thread. Cutlass has abstractions for these (we're using them to read in Q, K, V) but it'll take some trial and error.

maidabu:
> 

Thanks very much for replying.
I found that you copy result of softmax to global memory directly without usage of shared memory in write_softmax_to_gmem. Why not use shared memory here?
I tried to copy global bias to acc_s referred to the write_softmax_to_gmem in flash_fwd_kernel.h, but got incorrect result. Is the idea wrong?

tridao:
The softmax is written to gmem in a permuted layout, for simplicity. This softmax is only used for testing. So when we test, we "unpermute" this matrix using Pytorch [here](https://github.com/Dao-AILab/flash-attention/blob/08c295c043ed9463f0f3d19de67e03f469455fd2/tests/test_flash_attn.py#L327). This is slow, but it doesn't matter because it's only for testing.

maidabu:
[...]

Issue #342, opened 2023-07-19
does flash-attention support bias in attention layer? 


tiandiao123:
Hi:
   Thanks for amazing work! I am wondering whether this flash attention implementation supports bias in this case? Thank you!

tridao:
No it's not supported right now. Idk when I'll have time to add it due to other priorities.

tiandiao123:
> No it's not supported right now. Idk when I'll have time to add it due to other priorities.

thank you!

Issue #332, opened 2023-07-18
Relative positions, or Alibi, or attention_vias (T5) support


vince62s:
Hello,
At the moment it is impossible to use flash attention with one of those 3 above because it requires to alter the scores before the softmax.

Each of the 3 is a different way to encode positions but in the end for Relative positions or attention boas (T5) it is just a matter of adding a term to scores before the softmax and in the case of Alibi it's just applying a function.

Would there be an easy way to pass a function within flash_attn() so that it takes this into account and hence support those ?

Cheers. Great library btw.

tridao:
Right, none of those are supported right now. Eventually they should be supported, I just don't know when as there are a few other priorities.

cccc0der:
> Right, none of those are supported right now. Eventually they should be supported, I just don't know when as there are a few other priorities.

Wish bias too :)

PR #617, opened 2023-10-19
Custom attention bias (WIP)

b-albar:
This PR is an attempt to add custom (additive) attention biases. This is still very much a work in progress to say the least but I though to make my code available as there may be a lot of interest for this feature (#332, #342, #534, #606).

For now, there is no gradients for the bias matrix (I'm working on this) and it has been through very limited testing. But the forward pass as well as dq, dv, dk looks ok (tested only on A100). Also there has been no time spent on optimizing stuff.

I'm not familiar working with kernels so any comments or help for this PR would be very much welcome !

## Todo
- [ ] Return gradients for the bias matrix
- [ ] Test extensively (causal, head size, masking, ...)
- [ ] Benchmark the performance compared to vanilla FA
- [ ] Optimize

### DIFF
--- a/csrc/flash_attn/flash_api.cpp
+++ b/csrc/flash_attn/flash_api.cpp
@@ -41,7 +41,11 @@ void set_params_fprop(Flash_fwd_params &params,
                       float p_dropout,
                       float softmax_scale,
                       int window_size_left,
-                      int window_size_right) {
+                      int window_size_right,
+                      void* attn_bias,
+                      uint32_t attn_bias_batch_stride,
+                      uint32_t attn_bias_head_stride,
+                      uint32_t attn_bias_q_stride) {
 
     // Reset the parameters
     memset(&params, 0, sizeof(params));
@@ -62,6 +66,12 @@ void set_params_fprop(Flash_fwd_params &params,
     params.o_ptr = out.data_ptr();
     params.o_row_stride = out.stride(-3);
     params.o_head_stride = out.stride(-2);
+
+    // Attention biases
+    params.attn_bias_ptr = attn_bias;
+    params.attn_bias_batch_stride = attn_bias_batch_stride;
+    params.attn_bias_head_stride = attn_bias_head_stride;
+    params.attn_bias_q_stride = attn_bias_q_stride;
 
     if (cu_seqlens_q_d == nullptr) {
         params.q_batch_stride = q.stride(0);
@@ -148,7 +158,12 @@ void set_params_dgrad(Flash_bwd_params &params,
                       float p_dropout,
                       float softmax_scale,
                       int window_size_left,
-                      int window_size_right) {
+                      int window_size_right,
+                      void* attn_bias,
+                      void* attn_ds,
+                      uint32_t attn_bias_batch_stride,
+                      uint32_t attn_bias_head_stride,
+                      uint32_t attn_bias_q_stride) {
 
     set_params_fprop(params,
                      b, seqlen_q, seqlen_k, seqlen_q_rounded, seqlen_k_rounded, h, h_k, d, d_rounded,
@@ -160,7 +175,11 @@ void set_params_dgrad(Flash_bwd_params &params,
                      p_dropout,
                      softmax_scale,
                      window_size_left,
-                     window_size_right);
+                     window_size_right,
+                     attn_bias,
+                     attn_bias_batch_stride,
+                     attn_bias_head_stride,
+                     attn_bias_q_stride);
 
     // Set the pointers and strides.
     params.do_ptr = dout.data_ptr();
@@ -175,6 +194,9 @@ void set_params_dgrad(Flash_bwd_params &params,
     params.dq_head_stride = dq.stride(-2);
     params.dk_head_stride = dk.stride(-2);
     params.dv_head_stride = dv.stride(-2);
+
+    // Attention biases
+    params.attn_ds_ptr = attn_ds;
 
     if (cu_seqlens_q_d == nullptr) {
         params.do_batch_stride = dout.stride(0);
@@ -255,6 +277,7 @@ mha_fwd(at::Tensor &q,         // batch_size x seqlen_q x num_heads x head_size
         bool is_causal,
         const int window_size_left,
         int window_size_right,
+        const c10::optional<at::Tensor> &attn_bias, // batch_size x num_heads_k x seqlen_q x seqlen_k
         const bool return_softmax,
         c10::optional<at::Generator> gen_) {
 
@@ -309,6 +332,22 @@ mha_fwd(at::Tensor &q,         // batch_size x seqlen_q x num_heads x head_size
     CHECK_SHAPE(q, batch_size, seqlen_q, num_heads, head_size_og);
     CHECK_SHAPE(k, batch_size, seqlen_k, num_heads_k, head_size_og);
     CHECK_SHAPE(v, batch_size, seqlen_k, num_heads_k, head_size_og);
+
+    // Attention biases
+    uint32_t attn_bias_batch_stride = 0;
+    uint32_t attn_bias_head_stride = 0;
+    uint32_t attn_bias_q_stride = 0;
+
+    if (attn_bias.has_value()) {
+        TORCH_CHECK(attn_bias.value().is_cuda(), "Input tensor must be on CUDA device");
+        TORCH_CHECK(attn_bias.value().stride(-1) == 1, "Input tensor must have contiguous last dimension");
+        TORCH_CHECK(attn_bias.value().dtype() == q_dtype, "attention bias and query must have the same dtype");
+        CHECK_SHAPE(attn_bias.value(), batch_size, num_heads, seqlen_q, seqlen_k);
+
+        attn_bias_batch_stride = attn_bias.value().stride(0);
+        attn_bias_head_stride = attn_bias.value().stride(1);
+        attn_bias_q_stride = attn_bias.value().stride(2);
+    }
 
     at::Tensor q_padded, k_padded, v_padded;
     if (head_size_og % 8 != 0) {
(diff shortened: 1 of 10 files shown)


khj94:
An error occurred during an experiment with this code.
It was confirmed that the sequence length (512, 1024, 2048, 4096) used in `test_attn_bias.py` operates normally. However, when I tested various sequence lengths other than multiples of 512, I found that `RuntimeError: CUDA error: misaligned address` occurred. 
Can't this code be used in inference?

khj94:
@b-albar Thank you for responding quickly.
When I checked the newly committed code, I confirmed that forward was operating normally. However, this time, the same `RuntimeError: CUDA error: misaligned address` occurs in backwards.
To fix the above error in the backward pass, you need a pad code in the bias tensor like the forward pass.

b-albar:
Actually, the way its implemented for qkv is that the padded tensors are returned in the forward and reinjected through the python interface in the backward. So no need to pad it twice. I added this in the last commit. Also I encountered some nan in dq,dk,dv and had to implement some condition for boundaries on row and columns similar to what was implemented in the alibi patch. But it look like I ran into some race conditions (due to branching ?), and I had to add a syncthread(). It looks ok now (hopefully) but I'll have to investigate more on that later.

b-albar:
I added the gradients for the bias. The gradient is computed only in case of requires_grad in the bias. This avoid to copy a large matrix if it's not needed (like for alibi style bias or masking) as it's one of the main feature of FA to not have to copy this matrix.

khj94:
I want to check the last commit, but an error occurs during the build process.
After checking the code, there is no identifier in the two places below. 
Could you please check?

```
flash-attention/csrc/flash_attn/src/flash_bwd_kernel.h(1618): error: identifier "Is_even_MN" is undefined
                  if (Is_even_MN || get<0>(tdScdS(0, m, 0)) < binfo.actual_seqlen_q - m_block * kBlockM) {
                      ^
flash-attention/csrc/flash_attn/src/flash_bwd_kernel.h(1627): error: identifier "m_block_min" is undefined
              if (m_block > m_block_min) {
                            ^
```

b-albar:
@khj94 My bad, should be good now!

turboderp:
Is there any progress on this, and can I help?

b-albar:
@turboderp Most of the work remaining is finding the right parameters for the various head dim. I did it for head dim 64 [here](https://github.com/Dao-AILab/flash-attention/pull/617/commits/2b0885b71c4ed70c65c490c455f708c86add2494) as it's the one I use but the perfs are not optimal for the others. Also, I will have a bit of time this week to fix the latest conflicts.

Taytay:
Thank you for your work on this @b-albar! I look forward to seeing this merged to get T5 some FA2 love. :)

Taytay:
Just dropping in as an interested observer @b-albar to say: Thank you for your continued work on this. This PR is going to unlock some serious efficiency gains for T5 models, and I think it will lead to some great stuff. ❤️

c0nn3r:
Just echoing what others have said - extremely excited for this to land - let me know if there is anything I can do to help!

csrc/flash_attn/flash_api.cpp
```diff
@@ -383,6 +406,47 @@ mha_fwd(at::Tensor &q,         // batch_size x seqlen_q x num_heads x head_size
...
+    #endif
+
+    if (attn_bias.has_value()) {
+        TORCH_CHECK(attn_bias.value().is_cuda(), "Input tensor must be on CUDA device");
+        TORCH_CHECK(attn_bias.value().stride(-1) == 1, "Input tensor must have contiguous last dimension");
+        TORCH_CHECK(attn_bias.value().dtype() == q_dtype, "attention bias and query must have the same dtype");
+
+        auto bias_sizes = attn_bias.value().sizes();
+        TORCH_CHECK((bias_sizes[0] == batch_size) || (bias_sizes[0] == 1), "First dimension of the bias should be 1 or batch size");
+        TORCH_CHECK((bias_sizes[1] == num_heads) || (bias_sizes[1] == 1), "First dimension of the bias should be 1 or num_heads");
```
c0nn3r: Shouldn't this be that the second dimension of the bias should be `1` or `num_heads`?

c0nn3r:
I'll make a note here for anyone wanting to give this a try right now that all tests currently are failing.

b-albar:
@c0nn3r Thanks for noticing this. There is clearly a memory error somewhere. It is very odd that it didn't show up in my local version. I usually compile with alibi and some other stuff disabled as it is faster for debugging but this error show up only  when all options are enabled ! This is very odd, i'll investigate what's going on. In the mean time specifying "-DFLASHATTENTION_DISABLE_ALIBI" for compilation seems to do the trick (at least for me) as a temporary workaround.

csrc/flash_attn/flash_api.cpp
b-albar: Yes this is a typo, thanks !

c0nn3r:
Thanks for looking into it.

Hm, I just compiled it ([7fe76b6](https://github.com/Dao-AILab/flash-attention/pull/617/commits/7fe76b62fd74c0cbaeb6858ca601b143a5403816)) and I'm still seeing `RuntimeError: CUDA error: misaligned address`

b-albar:
I definitively didn't have a misaligned adress error previously. Could you post here the pytest logs here ? I'd like to see which seqlen are failing. Also which GPU and cuda version do you use ?

b-albar:
Ok after some experiments, I think we are dealing with a compiler issue here. That would explain why it work after disabling some options and why I didn't found it before. It look like the kernel is quite big and the compiler doesn't like it too much (just compiling the kernel with debug infos crash the compilation). I was using CUDA 12.3 update 1 previously and I could reproduce the memory error even while disabling alibi. But disabling alibi, uneven_k and local worked fine without any change in the code ! Updating CUDA to 12.3 update 2 works better (the changelog mention some register spilling issue and potential miscompilation - I don't know if it could be this) and now just only disabling alibi works again. I'll try to recompile the code with all the options to see. For now, I'll disable all options on this branch until it is fully clear what happens, at least it will speed up the compilation.

@c0nn3r did you use CUDA 12.3.1 to compile ?

c0nn3r:
@b-albar I used CUDA 12.1 to compile.

c0nn3r:
I get the same result as before using:
```
nvcc: NVIDIA (R) Cuda compiler driver
Copyright (c) 2005-2023 NVIDIA Corporation
Built on Wed_Nov_22_10:17:15_PST_2023
Cuda compilation tools, release 12.3, V12.3.107
Build cuda_12.3.r12.3/compiler.33567101_0
```

I don't think it is the compiler - both the flash attention and attention bias tests fail.

wehos:
Hi, there! I am willing to help with the testing. I tried on 4090 (built with CUDA12.1 and your latest branch). It ends up with half `111 failed, 177 passed in 46.93s`, from two errors `RuntimeError: CUDA error: invalid argument` and `RuntimeError: FlashAttention backward for head dim > 192 requires A100/A800 or H100/H800`. I tried it with my app, it continues to throw `RuntimeError: CUDA error: invalid argument`.

I'm going to test on A100 PCIE 40G now.

sentialx:
Finally someone did it! Does it also support negative bias (i.e mask)?

jlamprou:
I am currently testing this, i have implemented it for [switchtransformers](https://huggingface.co/docs/transformers/main/en/model_doc/switch_transformers#switchtransformers), and an NVIDIA A100 40GB. Training at the MLM task with an LR=1.5
e-4. Loss scaling is very unstable. Lowering the LR and increasing the warmup_ratio seems to help but at some point something explodes and the destroys loss scaling.
W B Chart 3_9_2024, 5 24 25 PM

jlamprou:
@b-albar Sorry i didn't clarify, im using bf16, but the Switch architecture is tricky too with mixed-precision (they use selective mixed precision so some layers are bf16 and others are fp32), tried it again with the new commit on a subset of starcoder data(7B tokens) and reached 0.17 loss on epoch 0.32 which is a bit ridiculous on an MLM task

b-albar:
> @b-albar Sorry i didn't clarify, im using bf16, but the Switch architecture is tricky too with mixed-precision (they use selective mixed precision so some layers are bf16 and others are fp32), tried it again with the new commit on a subset of starcoder data(7B tokens) and reached 0.17 loss on epoch 0.32 which is a bit ridiculous on an MLM task

Ok I think I see, you mean the loss going (too) quickly to zero ? In this case, maybe it's a causality issue. In this case it's better to force causality in the decoder using causal=True and not using the bias matrix for this. It look like the SwitchTransformers architecture is similar to T5, I'm working on a T5 with flash-attention, maybe it can help. The code is available [here](https://github.com/catie-aq/flashT5).

jlamprou:
> > @b-albar Sorry i didn't clarify, im using bf16, but the Switch architecture is tricky too with mixed-precision (they use selective mixed precision so some layers are bf16 and others are fp32), tried it again with the new commit on a subset of starcoder data(7B tokens) and reached 0.17 loss on epoch 0.32 which is a bit ridiculous on an MLM task
> 
> Ok I think I see, you mean the loss going (too) quickly to zero ? In this case, maybe it's a causality issue. In this case it's better to force causality in the decoder using causal=True and not using the bias matrix for this. It look like the SwitchTransformers architecture is similar to T5, I'm working on a T5 with flash-attention, maybe it can help. The code is available [here](https://github.com/catie-aq/flashT5).

Thanks a lot, i was actually looking for T5-like implementation for Flash, Switch are basically T5-MOE. I've also replaced the T5 RPE, with [FIRE](https://openreview.net/pdf?id=rR03qFesqk) for bias which gave me better performance on length interpolation.

FSSRepo:
@b-albar Could this work with infinite negatives (custom attention mask)?

PR #647, opened 2023-10-30
Allow varlen_fwd to take optional seqused_k

bottler:
Add a new argument `seqused_k` to varlen_fwd / mha_varlen_fwd . By supplying a tensor (instead of None/nullptr), a user can indicate how many of the keys in each batch element should actually be attended to. This allows k and v to contain padding.


tridao:
Thanks @bottler, this looks fine. I assume this is only for the forward pass?

bottler:
Yes, I have in mind inference with a kv-cache.

PR #666, opened 2023-11-09
Fix typo in RotaryEmbedding forward output type

antony-frolov:


tridao:
Great, thanks @antony-frolov!

PR #673, opened 2023-11-16
fix flash ce mp large vocab

simonJJJ:
Hi @tridao, I found and fixed a bug for large vocab size under mp settings. The original implementation would cause an all_gather bug.


tridao:
Thanks @simonJJJ !

Issue #672, opened 2023-11-16
Could we have a `flash_attn_varlen_func_with_kv_cache`?


garrett4wade:
Hi, I'm trying to implement [inflight batching](https://github.com/triton-inference-server/server/issues/6358) with flash-attn, i.e., during generation, when one sequence completes with the EOS token, we put a new prompt at this position such as the PAD token will not accupy computation bandwidth. This technique results in variable-length inputs during each generation step, (e.g., [1, 1, 3, 1] if the new prompt is length 3), so I'm wondering whether `flash_attn_varlen_func_with_kv_cache` can be preferred.

BTW the reason I'm trying to do this instead of using inference libraries like vLLM is that I need to train the model after the generation, so I must ensure the model runs generation with PyTorch, otherwise I must do parameter synchronization after each train step.

garrett4wade:
I have implemented this feature and opened a PR #685 .

PR #685, opened 2023-11-22
Add `flash_attn_varlen_func_with_kvcache`.

garrett4wade:
Implementing a new function called [`flash_attn_varlen_func_with_kvcache`](https://github.com/garrett4wade/flash-attention/blob/main/flash_attn/flash_attn_interface.py#L1055), which behaves similar to `flash_attn_with_kvcache` but allows variable-length q/k/v inputs.

## Aim for Implementing This Feature

This enables [inflight batching](https://github.com/triton-inference-server/server/issues/6358 )(#672) , i.e., during generation, when one sequence completes with the EOS token, we put a new prompt at this position such as the PAD token will not accupy computation bandwidth. This technique results in variable-length inputs during each generation step, (e.g., [1, 1, 3, 1] if the new prompt is length 3).

Enabling inflight batching in flash-attn benefits LLM [RLHF](https://en.wikipedia.org/wiki/Reinforcement_learning_from_human_feedback) which requires high generation throughput (no pad in this case) while updating model parameters at the same time (we can't use inference libraries like vLLM because we need to synchronize parameters after each train step, which is expensive).

## Usage

q/k/v are packed 1D tensors. `q` should be passed in together with a `cu_seqlens_q` and `max_seqlen_q` similar to `flash_attn_varlen_func`. `k` and `v` are optional arguments, which should be passed in with `cu_seqlens_k`. `max_seqlen_k` is determined by kv cache. kv cache will be updated in-place. Use kv cache only if `k` and `v` are not passed in. 

`k_cache` and `v_cache` still have shape (batch_size_cache, seqlen_cache, n_heads, head_dim).

## Major Changes:

+ Adding `mha_varlen_fwd_kvcache` in `csrc/flash_attn/flash_api.cpp`. This function is similar to `mha_fwd_kvcache` but sets forward params appropriately to deal with variable-length inputs. `block_info.h` is also changed accordingly.
+ Adding `flash_attn_varlen_func_with_kvcache` in `flash_attn/flash_attn_interface.py`.
+ Adding `test_flash_attn_varlen_kvcache` in `tests/test_flash_attn.py`.

## Minor Changes:

+ Add a `make clean` option to remove compilation cache. I found weird bugs when compile the code again without `make clean` .
+ Minor refactor of `mha_fwd_kvcache`. Set cache length to the `k_cache_seqlens` attribute of forward params instead of using `cu_seqlens_k` and `is_seqlens_k_cumulative`. Add a new attribute called `cu_seqlens_knew` to distinguish it from `cu_seqlens_k`. `block_info.h` and `flash.h` are changed accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,7 @@
 
+clean:
+	rm -rf build/ dist/ *.egg-info/ *.so
+
 clean_dist:
 	rm -rf dist/*
 
--- a/csrc/flash_attn/flash_api.cpp
+++ b/csrc/flash_attn/flash_api.cpp
@@ -72,6 +72,8 @@ void set_params_fprop(Flash_fwd_params &params,
 
     params.cu_seqlens_q = static_cast<int *>(cu_seqlens_q_d);
     params.cu_seqlens_k = static_cast<int *>(cu_seqlens_k_d);
+    params.k_cache_seqlens = nullptr;
+    params.cu_seqlens_knew = nullptr;
 
     // P = softmax(QK^T)
     params.p_ptr = p_d;
@@ -114,8 +116,6 @@ void set_params_fprop(Flash_fwd_params &params,
     if (window_size_left >= 0 && window_size_right < 0) { window_size_right = seqlen_k; }
     params.window_size_left = window_size_left;
     params.window_size_right = window_size_right;
-
-    params.is_seqlens_k_cumulative = true;
 }
 
 void set_params_dgrad(Flash_bwd_params &params,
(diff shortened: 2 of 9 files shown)


shcho1118:
@garrett4wade 
If this PR applies, could it be used in place of vllm's paged attention v2 kernel?
The problem I have right now is that I can't use the splitkv kernel in varlen funcs in the decoding phase.

garrett4wade:
@shcho1118 Hi cho, I don't think this PR can help because I forcely set num_splits=1 for varlen func with kv cache, although this can be potentially fixed in the future.

sgrigory:
> @shcho1118 Hi cho, I don't think this PR can help because I forcely set num_splits=1 for varlen func with kv cache, although this can be potentially fixed in the future.

@garrett4wade @shcho1118 FYI, I'm adding split-kv to `mha_varlen_fwd` for decoding in https://github.com/Dao-AILab/flash-attention/pull/754

Issue #429, opened 2023-08-07
The output result of _flash_attn_backward in flash_attn_triton.py is random


t90tank:
I use flash attention to train my model multiple times, and get different results when use same seed.
The reason is there will be different results each time when calling _flash_attn_backward.

Here is my code appending in flash_attn_triton.py to confirm it:

```
if __name__ == "__main__":
    batch_size = 32
    seqlen = 1024
    nheads = 32
    headdim = 128
    torch.manual_seed(233)
    q = torch.rand((batch_size, seqlen, nheads, headdim), dtype=torch.bfloat16, device="cuda")
    k = torch.rand_like(q)
    v = torch.rand_like(q)
    bias = torch.rand((1, nheads, seqlen, seqlen), dtype=torch.bfloat16, device="cuda")
    q, k, v = [x if x.stride(-1) == 1 else x.contiguous() for x in [q, k, v]]
    o, lse, softmax_scale = _flash_attn_forward(
        q, k, v, bias=bias, causal=True, softmax_scale=None)
    do = torch.rand_like(o)
    with torch.inference_mode():
        dq = torch.empty_like(q)
        dk = torch.empty_like(k)
        dv = torch.empty_like(v)
        _flash_attn_backward(do, q, k, v, o, lse, dq, dk, dv,
                                bias=bias, causal=True, softmax_scale=None)
    t = {
            "dq" : dq,
            "dk" : dk,
            "dv" : dv
        }
    print(t)
    if False: #True when run in first time
        torch.save(t, "t0.pt")
    t0 = torch.load("t0.pt")
    for key in ["dq", "dk", "dv"]:
        print(torch.abs(t0[key] - t[key]).max())
```
The reuslt: 
```
...
tensor(3.8147e-06, device='cuda:0', dtype=torch.bfloat16)
tensor(0., device='cuda:0', dtype=torch.bfloat16)
tensor(0., device='cuda:0', dtype=torch.bfloat16)
```

tridao:
The backward is non-deterministic because we use atomic adds. This is expected.

t90tank:
> The backward is non-deterministic because we use atomic adds. This is expected.

Thank you for reply. May I ask if there is any solution? The inability to reproduce the experiment has caused some difficulties for our team.

tridao:
[...]

PR #722, opened 2023-12-13
[feat] deterministic backward atomicAdd version

defei-coder:
FA2 Supports deterministic Computation Feature, fix this issue https://github.com/Dao-AILab/flash-attention/issues/429

- Add one deterministic flag like FA1 to control whether the results are deterministic.
- This feature needs extra worksapce for semaphores.
- For convenience, users are allowed not to allocate extra workspace, it will be allocated automatically and throws one warning message.
- We do not support deterministic=True and local=True at the same time, will workaround this limitation later.

Example(for usage and test):
 
        workspace = torch.zeros(flash_get_bwd_workspace_size_func(batch_size, nheads, seqlen_q, d), device=device, dtype=torch.int32)
        q = torch.randn(batch_size, seqlen_q, nheads, d, device=device, dtype=dtype, requires_grad=True)
        k = torch.randn(batch_size, seqlen_k, nheads, d, device=device, dtype=dtype, requires_grad=True)
        v = torch.randn(batch_size, seqlen_k, nheads, d, device=device, dtype=dtype, requires_grad=True)
        out = flash_attn_func(q, k, v, 0.0, causal=causal, window_size=window_size, workspace=workspace, deterministic=True)
        g = torch.randn_like(out)

        (
            dq,
            dk,
            dv,
         ) = torch.autograd.grad(out, (q, k, v), g, retain_graph=True)

        (
            dq1,
            dk1,
            dv1,
        ) = torch.autograd.grad(out, (q, k, v), g)
	assert (dq - dq1).abs().max().item() == 0

By this way, result of dq is deterministic. We run twice backward, get dq and dq1, The results of these two are completely identical.

### DIFF
--- a/csrc/flash_attn/flash_api.cpp
+++ b/csrc/flash_attn/flash_api.cpp
@@ -795,6 +795,7 @@ mha_bwd(const at::Tensor &dout,  // batch_size x seqlen_q x num_heads, x head_si
         c10::optional<at::Tensor> &dq_,   // batch_size x seqlen_q x num_heads x head_size
         c10::optional<at::Tensor> &dk_,   // batch_size x seqlen_k x num_heads_k x head_size
         c10::optional<at::Tensor> &dv_,   // batch_size x seqlen_k x num_heads_k x head_size
+        c10::optional<at::Tensor> &workspace_,   // b x h x s / blockM
         const float p_dropout,         // probability to drop
         const float softmax_scale,
         const bool is_causal,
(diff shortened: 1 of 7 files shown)


evanluyifan:
Since extra workspace could be reused during runtime, to avoid redundant GPU driver alloc/free at high frequence.
We may need to expose the python api for alloc extra workspace, let the framework allocator to handle gpu mem reuse.

hwu36:
https://research.colfax-intl.com/nvidia-hopper-flashattention-2/ replaced ampere `mma.sync` and `cp.async` with hopper `tma` and `wgmma` on hopper.  It needs to use smaller tile size to prevent reg spilling and use warp specialization to hide more data movement.

@jayhshah

evanluyifan:
Hi, @tridao  this PR has been opened for a while, could you help to do the code review?
BTW, this change has been tested in cases from Meituan, the bwd outputs are deterministic.

tridao:
Thanks, just got back from some travel, let me review it this week.

tridao:
Thanks, I've incorporated some of the idea here with a slightly different [approach](https://github.com/Dao-AILab/flash-attention/commit/732654583c2e640adc012ecb60e460bf19dcd9e3), and there's an option for deterministic bwd as of v2.4.1. I've acknowledged your contribution in the README.

PR #768, opened 2024-01-12
Add ability to return z_loss component of overall loss

cghawthorne:
If `return_z_loss` is True, return how much z_loss contributed to the overall loss.


tridao:
Thanks @cghawthorne !

PR #788, opened 2024-01-24
Preprocessor switches to control functionality

bottler:
For faster and smaller builds in some simple cases, provide switches to allow disabling

* backward
* alibi
* uneven k
* dropout
* local attention

Note that the error checking here depends on the C++ preprocessor definitions, and most of the implementation depends on the nvcc preprocessor definitions, so I am assuming users will set them the same in both places.

cc @drisspg @danthe3rd @sgrigory


tridao:
Thanks @bottler !

PR #795, opened 2024-01-27
Updated missing docstrings for args and returns in bert_padding.py

Avelina9X:
Updated docstring for `unpad_input()` to include missing `indices` return.
Updated docstring for `unpad_input_for_concatenated_sequences()` to include missing `indices` return.
Updated docstring for `pad_input()` to include more detailed description of `indices` argument and include missing `batch` and `seqlen` arguments.


tridao:
Thanks!

navvye:
I think the example of the 3D attention mask should be non-causal because the code is under bert_padding.py.

PR #793, opened 2024-01-27
[CI] Build wheels for Pytorch 2.3 (dev/nightly)

ckadner:
Add PyTorch 2.3 (dev/nightly) to CI build matrix.

Resolves #790

PR #800, opened 2024-01-30
[CI] Use torch 2.2.0 nightly build dev20231106

ckadner:
Tying up loose ends :-)

@tridao

> nvcr 24.01 is out and still uses torch2.2.20231106. I'll merge this and fix that later.

_Originally posted by @tridao in https://github.com/Dao-AILab/flash-attention/issues/793#issuecomment-1915919111_

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -44,7 +44,7 @@ jobs:
           # manylinux docker image, but I haven't figured out how to install CUDA on manylinux.
           os: [ubuntu-20.04]
           python-version: ['3.7', '3.8', '3.9', '3.10', '3.11']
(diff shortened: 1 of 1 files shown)


ckadner:
Hmm, I don't think `1106` is still available. I just tried a pip install and looked at the nightly build mirror

https://download.pytorch.org/whl/nightly/torch/

... only `1010` is still around before `1205` so `1130` should not work any longer either.

I suppose we wait with this change until just before the next release of `flash-attention` (2.5.1) to pick a nightly build that will still be around then?

tridao:
I think I figured out how to install torch 2.2.0.dev20231106. Let's see if flash-attn 2.5.1 builds successfully.

tridao:
flash-attn 2.5.1 wheels build successfully. Thanks @ckadner !

PR #804, opened 2024-01-30
Add arbitrary mask support to flash attention

c2mda:
- Change fwd and bwd kernels and API to support passing in a [B, H, Q, K] tensor representing arbitrary mask
- Score[b,h,q,k] is set to -INF if mask[b,h,q,k] < 0.5
- Add Python test parameterized over our typical configs
- Tests are passing on H100
- We diverged from upstream v2.3.3 as this is the latest one TE is using.

Things I am not 100% confident in because I don't fully understand them:
- correctness of the thread/copy layouts. I took them from existing code and existing PR bias.
- how to get actual q/k coordinates from thread index for flash::apply_attn_mask. I took them from existing code.
- hard to debug as code does not compile with debug symbols so can't use gdb. hacky way to identify mapping from threads to actual tensor coordinates is print thread coordinates and tensor values from all threads, then match values to tensor computed in pytorch

still TODO
- varlen support. We don't need "real" varlen support. TE uses the varlen version but we can just use the mask to hide padding

related upstream PR still in flight for additive attn bias:
[Dao-AILab/flash-attention#617](https://github.com/Dao-AILab/flash-attention/pull/617)

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -14,24 +14,6 @@ on:
 
 jobs:
 
-  setup_release:
-    name: Create Release
-    runs-on: ubuntu-latest
-    steps:
-      - name: Get the tag version
-        id: extract_branch
-        run: echo ::set-output name=branch::${GITHUB_REF#refs/tags/}
-        shell: bash
-
-      - name: Create Release
-        id: create_release
-        uses: actions/create-release@v1
-        env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        with:
-          tag_name: ${{ steps.extract_branch.outputs.branch }}
-          release_name: ${{ steps.extract_branch.outputs.branch }}
-
   build_wheels:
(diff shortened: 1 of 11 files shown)


nikita-petrashen:
Hey @c2mda !
Very important and most welcomed feature, thank you. May I ask why this PR was closed? Is the code in a usable state? I solely need `flash_attn_func` with `attn_mask` as a substitute for `F.scaled_dot_product_attention` in `nn.MultiheadAttention`. Can I do that with your code? Thanks!

c2mda:
Hi @nikita-petrashen 

It is not ready for review: 
- performance is not tested
- some crucial features are missing (eg mask is passed in bf16 instead of bool, wasting a lot of memory)
- I am not confident in some important details of thread layouts and tensor coordinate arithmetic

No ETA for these, sorry.  That said, tests are passing on Ampere and Hopper, so if you like to live dangerously feel free to use.

PR #810, opened 2024-02-01
Added support for building for C++ to use in LibTorch

shaltielshmid:
As a user of LibTorch (the C++ bindings) and TorchSharp (the C# bindings), the FlashAttention feature was sorely missing. 
I was able to compile the C++ API + Cuda Binaries into a dynamic link library both on windows and on linux, without making any real modifications. 
I am hoping to add bindings on top of the C++ binaries to C#, for easy usage in TorchSharp. The changes included here are what I needed to do to get it building - and I wanted to share with the world. 

The modifications include:
1] Adding CMakeLists.txt
2] Adding a header file `flash_api.h` on top of `flash_api.cpp`
3] Added preprocessor directives to disable the python functionality and to enable the dll export for windows. 

I added a basic template to README, but of course I am happy to make any changes as you see fit.

Would love to hear feedback and to see this merged in :)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,84 @@
+cmake_minimum_required(VERSION 3.18)
+
+# Set CMAKE_C_FLAGS_RELEASE without /DNDEBUG
+set(CMAKE_C_FLAGS_RELEASE "/MD /O2 /Ob2 /DDLL_EXPORT /DNOT_PY_BUILD" CACHE STRING "Release flags" FORCE)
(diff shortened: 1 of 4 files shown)


shaltielshmid:
I see that there are other packages included here like the FusedDense and LayerNorm. 
If the changes here are acceptable, should I extend the CMakeLists compilation to optionally include all of the other modules?

I've already written most of the TorchSharp wrapper which can provide access to FlashAttention for all the TorchSharp users, and I would love for these changes to merge in, so that future updates can be streamlined.

FSSRepo:
I am trying to see if I could compile it without the need for libtorch, just a C++ API, does this require compiling the entire cutlass library or just what is needed for flash attention?

Issue #814, opened 2024-02-04
Windows build is broken


AlpinDale:
It seems #754 has broken windows builds. Look [here](https://github.com/bdashore3/flash-attention/actions/runs/7760920034/job/21168254905) for more details on where/how it fails. Can confirm it fails on baremetal Windows as well.

CC @tridao @sgrigory

bdashore3:
This is reproducible on my home PC as well using the latest commit of FA2. I cannot build any version that's >= 2.4.3

Malrama:
Same here

PR #816, opened 2024-02-06
Fix Windows build

sgrigory:
Fix build error reported in https://github.com/Dao-AILab/flash-attention/issues/814

I don't have a way to test this on Windows, but @bottler pointed out that `long` keyword is the likely problem. Is there a way to run Windows CI on this PR? @tridao

### DIFF
--- a/csrc/flash_attn/flash_api.cpp
+++ b/csrc/flash_attn/flash_api.cpp
@@ -696,8 +696,8 @@ mha_varlen_fwd(at::Tensor &q,  // total_q x num_heads x head_size, total_q := \s
     }
 
     if (seqlenq_ngroups_swapped) {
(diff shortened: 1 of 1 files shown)


shaltielshmid:
I got it compiling on windows by changing to this:

image

bottler:
@shaltielshmid I think that your version would also be a valid fix, equally as good as this one.

bdashore3:
@sgrigory I successfully built FA2 on windows using this PR. The Github Actions log is located [here](https://github.com/bdashore3/flash-attention/actions/runs/7819237957) and the release is [here](https://github.com/bdashore3/flash-attention/releases/tag/v2.5.2)

Thanks for the PR!

PR #819, opened 2024-02-09
Add C++ build support for use with LibTorch

shaltielshmid:
1] Added a CMakeLists.txt file equivalent to setup.py for compiling in C++
2] Added a preprocessor directive when building in Python so the python headers and binding occur (I added it for the Python build since that requires adding the directive just in the setup.py, whereas for C++ would need to be added in every project that references the project). 

I've tested out using the C++ compiled library both both for window & linux, and it works great! I also managed to build the python library with the adjust code for windows. 

Happy to make any changes, as you see fit. 

If this PR is accepted, I'd be happy to add similar changes for the other sub-packages here (LayerNorm, FusedDense, etc.) .

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,92 @@
+cmake_minimum_required(VERSION 3.18)
+
+project(FlashAttention)
+
+find_package(CUDA REQUIRED)
+find_package(Torch REQUIRED PATHS ${LIBTORCH_PATH})
+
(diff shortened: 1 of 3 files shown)


shaltielshmid:
@tridao Using the built C++ binaries I added a wrapper for TorchSharp for all the TorchSharp users to have access to your brilliant library! You can view it here: https://github.com/shaltielshmid/TorchSharp.FlashAttention

Issue #845, opened 2024-02-21
ALiBi slopes are incorrectly cast to `torch.float16`


Markus28:
First of all, thanks to the developers for this incredible project and the ALiBi integration in particular! When trying to use ALiBi with the BERT implementation of this repository, I encountered an issue relating to the dtype of the ALiBi slopes. 

## Problem
Flash attention may only be used with `fp16` or `bf16`, while the ALiBi slopes need to be at `fp32` precision. Thus, it is my understanding that we need to cast our models to lower precision, which will also affect the `alibi_slopes` buffers in `FlashSelfAttention` and `FlashCrossAttention`. Maybe I'm misunderstanding how ALiBi is intended to be used with flash attention, but this casting causes issues for me.

Consider the following minimal example, executed on the main branch, commit 43950dda456e095969d842fca7a73c5bfe3cecd0 with `torch==2.2.0`:
```python
import torch
from flash_attn.modules.mha import MHA

USE_ALIBI = True

module = MHA(embed_dim=16, num_heads=4, use_flash_attn=True, use_alibi=USE_ALIBI).to('cuda', dtype=torch.float16)
print(module(torch.randn(8, 8, 16).to('cuda', dtype=torch.float16)))
```
This will give the following exception:
```
Traceback (most recent call last):
  File "/home/markus/flash-attention/reproducer.py", line 7, in <module>
    print(module(torch.randn(8, 8, 16).to('cuda', dtype=torch.float16)))
  File "/home/markus/flash-attention/venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/markus/flash-attention/venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/markus/flash-attention/flash_attn/modules/mha.py", line 666, in forward
    context = self.inner_attn(qkv, **kwargs)
  File "/home/markus/flash-attention/venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1511, in _wrapped_call_impl
    return self._call_impl(*args, **kwargs)
  File "/home/markus/flash-attention/venv/lib/python3.10/site-packages/torch/nn/modules/module.py", line 1520, in _call_impl
    return forward_call(*args, **kwargs)
  File "/home/markus/flash-attention/flash_attn/modules/mha.py", line 120, in forward
[...]

PR #846, opened 2024-02-21
Cast the alibi slopes to torch.float32

Markus28:
This PR addresses https://github.com/Dao-AILab/flash-attention/issues/845. Since the `MHA` module may be cast to fp16 or bf16, but the `alibi_slopes` buffer should be fp32, we cast them to that precision in `forward`.


tridao:
Thanks!

vadimkantorov:
Will alibi slopes still lose precision upon the module cast?

Markus28:
The slopes will temporarily be in a different precision before being cast back to full precision in the next forward pass. In particular, there will be some small floating point errors coming from these cast. However, they looked very much negligible in my experiments. I assume this is what you were referring to?

PR #852, opened 2024-02-25
Add in Macros  for defining __grid_constant__

drisspg:
# Summary
Bringing in some changes from: https://github.com/pytorch/pytorch/pull/118935
- Get a Werror from using memset
- In PyTorch we release wheels defined for a large number of compute architectures, example list can be found here:    `TORCH_CUDA_ARCH_LIST="3.5 5.2 6.0 6.1 7.0+PTX 8.0"`
- When attempting to compile the flash_attention kernels on architectures less than 8.0 NVCC will error. Since we guard against this at runtime, https://github.com/Dao-AILab/flash-attention/blob/87a1277653fc55cd615f5341255e00c69d5c00a1/csrc/flash_attn/flash_api.cpp#L334, this shouldn't be a problem

Feel free to merge these in if you would like

### DIFF
--- a/csrc/flash_attn/flash_api.cpp
+++ b/csrc/flash_attn/flash_api.cpp
@@ -46,7 +46,7 @@ void set_params_fprop(Flash_fwd_params &params,
                       bool seqlenq_ngroups_swapped=false) {
 
     // Reset the parameters
-    memset(&params, 0, sizeof(params));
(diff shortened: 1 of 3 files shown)


tridao:
Thanks @drisspg, sorry for the delay.

Issue #730, opened 2023-12-18
[release] wheel should be published with local version available in wheel `METADATA`


yundai424:
Hi team,

This is an issue following up this PR: https://github.com/Dao-AILab/flash-attention/pull/229

Right now the released wheels, though with local versions available in the wheel's filename, doesn't encode the proper local version in the `METADATA`:
```
wget https://github.com/Dao-AILab/flash-attention/releases/download/v2.3.6/flash_attn-2.3.6+cu118torch2.0cxx11abiFALSE-cp38-cp38-linux_x86_64.whl && unzip flash_attn-2.3.6+cu118torch2.0cxx11abiFALSE-cp38-cp38-linux_x86_64.whl && cat flash_attn-2.3.6.dist-info/METADATA | head -n 10
```
This will give
```
Metadata-Version: 2.1
Name: flash-attn
Version: 2.3.6
Summary: Flash Attention: Fast and Memory-Efficient Exact Attention
Home-page: https://github.com/Dao-AILab/flash-attention
Author: Tri Dao
Author-email: trid@cs.stanford.edu
Classifier: Programming Language :: Python :: 3
Classifier: License :: OSI Approved :: BSD License
Classifier: Operating System :: Unix
```

it would be useful to encode the local version in wheel metadata as well and I believe the only thing needed is to add `FLASH_ATTN_LOCAL_VERSION` env var into GH action flow. Would be happy to raise a tentative PR if this is something you agree upon. Thanks!

PR #856, opened 2024-02-28
Add local version identifier to package metadata for pre-built wheels

yundai424:
This attempts to fix https://github.com/Dao-AILab/flash-attention/issues/730 so we follow [PEP 440](https://peps.python.org/pep-0440/#local-version-identifiers) better.

Currently the filename of pre-built wheels (published by GH Action in each release) follows the `public+local` versioning scheme as suggested in PEP 440 [local version identifier](https://peps.python.org/pep-0440/#local-version-identifiers) section, but this is NOT reflected in the actual package metadata ([PEP 459](https://peps.python.org/pep-0459/)) and this causes discrepancy. For example, if these wheels are uploaded to a private python index, for instance a hosted python index within a company, people can't specify the exact version to install by simply using `pip install flash-attn===2.4.2+<local version>`

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -213,38 +213,40 @@ if not SKIP_CUDA_BUILD:
     )
 
 
-def get_package_version():
+def get_public_version():
     with open(Path(this_dir) / "flash_attn" / "__init__.py", "r") as f:
         version_match = re.search(r"^__version__\s*=\s*(.*)$", f.read(), re.MULTILINE)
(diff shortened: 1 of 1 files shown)


yundai424:
@tridao wondering if you would mind helping take a look at this change and see if it can be merged. Thanks!

yundai424:
@tridao @piercefreeman @ksivaman et al. would greatly appreciate your help to take a look at this. Thanks!

PR #885, opened 2024-03-13
Make nvcc threads configurable via environment variable

chiragjn:
My build environment kept using high cpu and was getting killed trying to compile fused_dense_lib.


tridao:
Can you use `os.getenv("NVCC_THREADS", 4)` instead?

chiragjn:
sure, I just kept it with `or` in case NVCC_THREADS is set but with empty value

tridao:
I see, that's fine then, don't need to change it.

PR #911, opened 2024-04-05
Fix spurious re-compilations of `rotary_kernel`

dfyz:
All integer parameters in Triton kernels are [specialized](https://github.com/openai/triton/blob/release/2.2.x/python/triton/runtime/jit.py#L180) by default, meaning that (for the `2.2` version of Triton) a different kernel will be generated for the following mutually exclusive classes of parameter values:
  * divisible by 16
  * divisible by 8, but not by 16
  * not divisible by either 8 or 16, and not equal to 1
  * equal to 1

`nheads` and `CACHE_KEY_SEQLEN` are not used at all in `rotary_kernel`, but they still trigger re-compilation whenever their values move between classes. This is doubly ironic given that the comment in the code indicates that `CACHE_KEY_SEQLEN` was meant to `limit number of compilations` (probably when it was used for autotuning?).

I removed the unused parameters and autotuning setup, and added a simple test to verify that changing the class of `nheads` and/or `CACHE_KEY_SEQLEN` (i.e., `seqlen//128`) doesn't lead to re-compilations. The test is a little hackish because it relies on undocumented Triton internals, but at least it shows the problem clearly (e.g., when you run the test with Triton 2.2 and without the fix applied, it fails with `FAILED tests/test_rotary.py::test_compilation_count - assert 8 == 2`).

I also strongly suspect that some other parameters that *are* used in the kernel (such as `seqlen`) don't get any benefit from being specialized; that is, we just do a re-compile to get essentially the same kernel. However, this is trickier to prove, and I wanted to make only safe changes in this PR.

### DIFF
--- a/flash_attn/ops/triton/rotary.py
+++ b/flash_attn/ops/triton/rotary.py
@@ -8,15 +8,6 @@ import triton
 import triton.language as tl
 
 
-# @triton.autotune(
-#     configs=[
-#         triton.Config({"BLOCK_M": 2}),
-#         triton.Config({"BLOCK_M": 4}),
-#         triton.Config({"BLOCK_M": 8}),
-#         triton.Config({"BLOCK_M": 16}),
-#     ],
-#     key=["CACHE_KEY_SEQLEN", "BLOCK_K", "INTERLEAVED"],
-# )
 @triton.jit
 def rotary_kernel(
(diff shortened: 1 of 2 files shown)


tridao:
You're right, thanks for the detailed investigation and fix!

PR #937, opened 2024-05-05
Move packaging and ninja from install_requires to setup_requires

weiji14:
Set `packaging` and `ninja` as build time dependencies rather than runtime dependencies.

Xref https://github.com/conda-forge/staged-recipes/pull/26239#discussion_r1590048101 where we're trying to package `flash-attn` on conda-forge, and would like the setup.py metadata to be correct so that running `pip check` would work.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -344,10 +344,10 @@ setup(
     install_requires=[
         "torch",
         "einops",
-        "packaging",
-        "ninja",
     ],
     setup_requires=[
-        "psutil"
+        "packaging",
+        "psutil",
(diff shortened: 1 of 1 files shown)


carterbox:
For clarity, it is highly unusual to require the ninja build system and packaging at runtime. setup_requires is the list of packages required to *build* this package. install_requires is the list of packages required to *run* this package.

dtrifiro:
Agreed, `ninja` and `packaging` should not be required when installing, but only when building it.

PR #956, opened 2024-05-14
Relative position encoding

b-albar:
This PR add fused relative positional encoding to the kernel. Contrary to #617, it doesn't pass the bias matrix to the kernel but only the bucket tensor allowing linear memory for T5 architecture and similar. For now, only the forward pass is supported (backward will come quickly).

## Todo
- [ ] Backward pass
- [ ] Test
- [ ] Benchmark


ywrt:
Is there an issue tracking this pull? is there some blocker to it being merged? (I'm looking for this functionality)

PR #958, opened 2024-05-17
add pyproject.toml with build dependencies

dhellmann:
Adding this metadata means it is possible to `pip install flash-attn`
without pre-installing the packages imported in setup.py. It is still
possible to follow the existing manual instructions, too, but by not
requiring pre-installation it _also_ makes it easier for someone to
build from the source dist.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "packaging", "psutil", "ninja", "torch", "wheel"]
+build-backend = "setuptools.build_meta"
--- a/Makefile
+++ b/Makefile
@@ -2,8 +2,10 @@
 clean_dist:
 	rm -rf dist/*
 
+# This step assumes there is already a virtualenv active.
 create_dist: clean_dist
-	python setup.py sdist
+	python -m pip install build
+	python -m build --sdist .
 
 upload_package: create_dist
 	twine upload dist/*


Skylion007:
Unfortunately, this also prevent src builds from compiling against the right version of PyTorch (due to build-isolation) and there is no standard way to turn it off. The only way to do it is to use a pre-pyproject.toml package and rely on legacy support for setup.py.

dhellmann:
> Unfortunately, this also prevent src builds from compiling against the right version of PyTorch (due to build-isolation) and there is no standard way to turn it off. The only way to do it is to use a pre-pyproject.toml package and rely on legacy support for setup.py.

This isn't correct. I still install torch ahead of time, and build without isolation using the `pip wheel` command and the `--no-build-isolation` option. `setup.py` is still run, and performs all of the validation it is doing now.

Ideally, when we have something like [wheel selectors](https://discuss.python.org/t/selecting-variant-wheels-according-to-a-semi-static-specification/53446/89) implemented, that will also help with this situation.

technillogue:
please use something like this, your package is impossible to use otherwise. does anyone have a fork that declares build dependencies correctly?

collinsleewyatt:
> Unfortunately, this also prevent src builds from compiling against the right version of PyTorch (due to build-isolation) and there is no standard way to turn it off. The only way to do it is to use a pre-pyproject.toml package and rely on legacy support for setup.py.

Does this mean that under the pyproject.toml implementation, flash-attention would always pull the latest version of torch and ignore any currently installed?

dhellmann:
> > Unfortunately, this also prevent src builds from compiling against the right version of PyTorch (due to build-isolation) and there is no standard way to turn it off. The only way to do it is to use a pre-pyproject.toml package and rely on legacy support for setup.py.
> 
> Does this mean that under the pyproject.toml implementation, flash-attention would always pull the latest version of torch and ignore any currently installed?

It depends on how the build is run.

In our build, we 

1. Apply this patch. 
2. Create a virtualenv with the version of torch we want along with the other build dependencies. 
3. Run `pip wheel` with the `--no-build-isolation` flag.

With `--no-build-isolation`, pip runs the build step using the environment where pip is installed. That gives us control over creating that virtualenv, including the torch version.

If `--no-build-isolation` is not used, then pip will create a virtualenv and install the build dependencies. With this patch, and without the flag, pip _will_ pull in the latest version of torch.

collinsleewyatt:
Are there other projects which target different pytorch versions that have gotten around the build isolation limitation?

The problem with setup.py is that it attempts to import torch and wheel, before calling `setup()` with torch and wheel in `install_requires`. Not sure, but maybe it's possible to modify setup.py to use dynamic imports or call setup() before said imports? Any thoughts on the feasibility of the former solution? Happy to work on a patch for this.

dhellmann:
For our project, we are also building

* deepspeed
* torchvision
* triton
* vllm
* vllm_flash_attn
* xformers

All of which depend on torch for their builds. They all include either just the package or the package with a version range (like `>=2.3.1`) in their pyproject.toml. We take the same approach with those of constructing the virtualenv using the version of torch we are using before building the package.

None of these packages are perfect. We're carrying some patches for all of them to fix up the builds, in some cases similar to the patch I've proposed here and in other cases to remove a pinned version. For example, [vllm](https://github.com/vllm-project/vllm/blob/main/requirements-build.txt) pins torch in their requirements list.

mgorny:
> Does this mean that under the pyproject.toml implementation, flash-attention would always pull the latest version of torch and ignore any currently installed?

I think there may be some confusion here.

1. If you call `setup.py` directly, it just uses the installed version, and the pull request doesn't change anything in that scenario.
2. If you install it from source via `pip install ...`, it will normally pull the newest version to build the wheel. However, without the pull request, `pip install ...` doesn't work at all — it fails because pip does build isolation anyway, and tries to build it in a virtual environment without pytorch at all.
3. If you install it from source via `pip install --no-build-isolation ...`, then the current version is used for the build — the pull request doesn't change anything in that scenario.
4. When installing the built wheel (either from PyPI or after `pip install`), the current version of PyTorch is also used (provided it satisfies `install_requires`) — so again, nothing changes.

So unless I'm missing something, this pull request fixes scenario 2. that currently doesn't work at all, and doesn't cause any regressions in the remaining scenarios.

rgommers:
> So unless I'm missing something, this pull request fixes scenario 2.

There are two flavors of that, with `pip install .` or `pip install flash-attn` being run in an environment that:
1. Doesn't have `torch` installed yet
2. Does have a version of `torch` installed already

(1) will indeed be fixed with this PR. 
(2) will work if the already-installed `torch` version is the latest one. If it's older, then due to build isolation, `flash-attn` will be built against latest `torch` and then find an older `torch` at runtime in the environment. Because this package builds PyTorch extensions with `torch.utils.cpp_extension` and PyTorch does not have a stable ABI, this will result in bad things (crashes probably, perhaps incorrect results). I suspect that @Skylion007's comment was hinting at the problem with the latter scenario, and why the `pyproject.toml` that was present before was moved to `flash_attn/pyproject.toml` 1.5 years ago in commit 73bd3f3bbb.

That ABI is indeed bad (crashes are worse than no build starting). The good news is that this is easily fixable: in `setup.py`, query the version of PyTorch used at build time, and then dynamically update `install_requires` to contain `torch==version_at_build_time`. That will fix the metadata in the wheel produced by `pip install`, and ensure a new `torch` version is pulled in.

If that were added to this PR, I'd say build isolation is no longer a problem, that's simply a feature of how Python packaging works. And all four of the scenarios @mgorny outlined will work as they should.

Given the number of related issues and PRs on this repo, as well as incoming links, it looks like a fair amount of people would benefit from having this PR merged. @tridao with that change added, would you be willing to merge this PR? Or is there anything else you'd like to see or are worried about?

pyproject.toml
```diff
@@ -0,0 +1,3 @@
+[build-system]
+requires = ["setuptools", "packaging", "psutil", "ninja", "torch", "wheel"]
```
weiji14, line 2: We've been able to get away with just `setuptools`, `torch` and `ninja` at https://github.com/conda-forge/flash-attn-feedstock/blob/9e6cd984d2d3018ff92b3ad4cbb867641638ba39/recipe/pyproject.toml#L2 when building the `flash-attn` package for conda-forge.

```suggestion
requires = ["setuptools", "ninja", "torch"]
```

Does `packaging`, `psutil` and `wheel` need to be declared here? They _are_ declared in `setup_requires` since https://github.com/Dao-AILab/flash-attention/pull/937 FWIW.

weiji14 (commented):
> The good news is that this is easily fixable: in setup.py, query the version of PyTorch used at build time, and then dynamically update install_requires to contain torch==version_at_build_time. That will fix the metadata in the wheel produced by pip install, and ensure a new torch version is pulled in.\n\nIf that were added to this PR, I'd say build isolation is no longer a problem

This exactly 👆🏻. We are effectively doing something similar for the conda-forge `flash-attn` package which has a constraint to install the same (minor) version of pytorch at runtime as that used at build. E.g. [`flash-attn=2.7.2`](https://anaconda.org/conda-forge/flash-attn/files?version=2.7.2) can only be installed with `pytorch>=2.5.1,<2.6.0a0`. I'm not sure about Pytorch's ABI guarantees, but constraining to allow any patch version within a minor version seems reasonable I suppose.

We've actually created a minimal [`pyproject.toml`](https://github.com/conda-forge/flash-attn-feedstock/blob/main/recipe/pyproject.toml) file for the conda-forge recipe (paired with a simplified `setup.py`) that has worked nicely since May 2024. Would like it if this `pyproject.toml` file is just available upstream here.

rgommers:
> I'm not sure about Pytorch's ABI guarantees, but constraining to allow any patch version within a minor version seems reasonable I suppose.

Technically PyTorch does not offer any ABI stability at all even for micro/bugfix releases, and exactly matching versions are needed. In the past, bugfix releases have tended to be mostly compatible, however not always.

pyproject.toml
mgorny: `packaging` is used directly in `setup.py`, so it must be declared here.

tiran: Same with `psutil` and `wheel`. The `wheel` package is imported in a top-level import. `psutil` is imported in a local import unless the user sets `MAX_JOBS` env var. `build-system requires` must list all non-stdlib packages that are needed by the build backend or `setup.py`.

hyliu1989: Could you pin the version of torch here (build-time environment) to match exactly the one installed in your target environment? I think that will solve the issue of different versions.

Issue #725, opened 2023-12-14
[Feature Request] Compact format for softmax_lse in mha_varlen_fwd and mha_varlen_bwd


GD06:
In the implementation of mha_varlen_fwd and mha_varlen_bwd, the softmax_lse has a shape of {batch_size, num_heads, max_seqlen_q}. Given that we have cu_seqlens_q, is it possible to change this to a compact format such as {num_heads, total_q}? With the help of cu_seqlens_q, we can still index the j-th token of the i-th sequence quickly in the kernel by accessing `cu_seqlens_q[i] + j`. But this could need changes to both APIs and CUDA kernels.

# Issue & Benefits:

Denote the total_q as `S`, if the batch of sequence lengths for Q is {1, 1, 1, ..., 1, S/2} which is composed of (S/2 + 1) sequences, the softmax_lse will consume `(S/2 + 1) x (S/2) x num_heads` memory space. This memory space complexity is O(S^2) while if we use a compact memory layout of {num_heads, total_q}, the memory space complexity is only O(S). This helps a lot on large model training on long variable sequences.

PR #970, opened 2024-05-28
Support unpadded LSE layout

sgrigory:
Closes https://github.com/Dao-AILab/flash-attention/issues/725

For variable seqlen case, keep LSE in `(H, total_q_seqlen)` format instead of current `(B, H, max_q_seqlen)`. This helps to save memory when variance of sequence length within the batch is large, e.g. during long sequence training. 

Based on https://github.com/GD06/flash-attention/pull/2

```
pytest tests/test_flash_attn.py 2>&1 | tee ~/log_on_branch.txt

...
============= 268004 passed, 152064 skipped in 4382.92s (1:13:02) ==============
```
Full output: https://gist.github.com/sgrigory/c95d8bdcd87721a60fb51f8c02be0f6b

cc @GD06 @jianyuh

### DIFF
--- a/csrc/flash_attn/src/flash_fwd_kernel.h
+++ b/csrc/flash_attn/src/flash_fwd_kernel.h
@@ -24,6 +24,20 @@ using namespace cute;
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+template<typename ElementAccum, typename Params, int kBlockM, bool Is_even_MN>
+//template<typename ElementAccum, int kBlockM, bool Is_even_MN>
+inline __device__ auto get_lse_tile(const Params &params, const int bidb, const int bidh, const int m_block, const BlockInfo</*Varlen=*/!Is_even_MN> &binfo) {
+
+        auto gmem_ptr_lse = make_gmem_ptr(reinterpret_cast<ElementAccum*>(params.softmax_lse_ptr) + (params.unpadded_lse ? binfo.q_offset(params.seqlen_q, 1, bidb): 0));
+        auto lse_shape = params.unpadded_lse? make_shape(1, params.h, params.total_q): make_shape(params.b, params.h, params.seqlen_q);
+        auto lse_layout = make_layout(lse_shape, LayoutRight{});
+
+        Tensor mLSE = make_tensor(gmem_ptr_lse, lse_layout);
+        auto mLSE_slice = params.unpadded_lse? mLSE(0, bidh, _): mLSE(bidb, bidh, _);
+        return local_tile(mLSE_slice, Shape<Int<kBlockM>>{}, make_coord(m_block));
+}
+
+
 template<typename Kernel_traits, bool Is_dropout, bool Is_causal, bool Is_local, bool Has_alibi, bool Is_even_MN, bool Is_even_K, bool Return_softmax, typename Params>
 inline __device__ void compute_attn_1rowblock(const Params &params, const int bidb, const int bidh, const int m_block) {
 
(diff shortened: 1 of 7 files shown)


csrc/flash_attn/src/flash_fwd_kernel.h
```diff
@@ -24,6 +24,19 @@ using namespace cute;
 
 ////////////////////////////////////////////////////////////////////////////////////////////////////
 
+template<typename ElementAccum, typename Params, int kBlockM, bool Is_even_MN>
+inline __device__ auto get_lse_tile(const Params &params, const int bidb, const int bidh, const int m_block, const BlockInfo</*Varlen=*/!Is_even_MN> &binfo) {
```
tridao: should be `__forceinline__` instead of `inline`

csrc/flash_attn/src/flash_fwd_kernel.h
```diff
@@ -131,6 +143,7 @@ inline __device__ void compute_attn_1rowblock(const Params &params, const int bi
                             make_stride(params.v_row_stride, params.v_head_stride, _1{}));
     Tensor gV = local_tile(mV(_, bidh / params.h_h_k_ratio, _), Shape<Int<kBlockN>, Int<kHeadDim>>{},
                            make_coord(_, 0));  // (kBlockN, kHeadDim, nblocksN)
+
```
jianyuh: nit: remove.

csrc/flash_attn/src/flash_fwd_kernel.h
```diff
@@ -1145,7 +1175,18 @@ inline __device__ void combine_attn_seqk_parallel(const Params &params) {
     // lse_logsum is log(0.0) = -INFINITY and we get NaN when we do lse_accum(l) - lse_logsum.
     ElementAccum lse_logsum = (lse_sum == 0.f || lse_sum != lse_sum) ? INFINITY : logf(lse_sum) + lse_max;
     // if (bidx == 0 && tidx < 32) { printf("tidx = %d, lse = %f, lse_max = %f, lse_logsum = %f\n", tidx, lse_accum(0), lse_max, lse_logsum); }
-    if (tidx % kRowsPerLoadTranspose == 0 && tidx / kRowsPerLoadTranspose < kBlockM) { gLSE(tidx / kRowsPerLoadTranspose) = lse_logsum; }
+    if (tidx % kRowsPerLoadTranspose == 0 && tidx / kRowsPerLoadTranspose < kBlockM) {
+        if (params.unpadded_lse) {
+        const index_t seq_q_idx = q_offset + (tidx / kRowsPerLoadTranspose);
```
jianyuh: nit: lint

tests/test_flash_attn.py
```diff
@@ -1563,7 +1564,6 @@ def test_flash_attn_varlen_causal(
     nheads = 9
     window_size = (-1, -1) if not local else torch.randint(0, seqlen_k, (2,))
     q = torch.randn(batch_size, seqlen_q, nheads, d, device=device, dtype=dtype, requires_grad=True)
-
```
jianyuh: nit: intentionally remove?

jianyuh (commented):
Looks great! BTW, do we worry about the BC issues: like some people store softmax_lse in previous FAv2 version, and then load it after the current version. Or using different softmax_lse format for FWD vs. BWD?

tridao:
> Looks great! BTW, do we worry about the BC issues: like some people store softmax_lse in previous FAv2 version, and then load it after the current version. Or using different softmax_lse format for FWD vs. BWD?

I think it's ok to break BC here since (1) previously for varlen the returned softmax_lse can have arbitrary values for indices that are not touched (2) most people only use the output, not the softmax_lse.
I view it as more of a bug fix.

sgrigory:
Sounds good, I've made unpadded format `(H, total_q_len)` default for variable seqlen path. That required some changes compared to the previous PR version, in particular in `combine_attn_seqk_parallel` - previously MHA + split-kv + unpadded LSE wasn't triggered by tests, since unpadded was off by default, and it started crashing once enabled. Now all the tests pass.
Note that this PR doesn't change the format on fixed seqlen path - it's `(B, H, Mq)` as before.

jianyuh (approved):

PR #978, opened 2024-06-03
Fix +/-inf in LSE returned by forward

sgrigory:
Forward op was returning `+inf` in LSE for queries which have no keys to attend to, e.g. when K/V length happens to be 0. This diverges from the definition of `LSE = log(exp(L1) + ... exp(L2))` which would give `log(0) = -inf`. 
This PR fixes it, which allows feeding the output LSE directly into ops like [merge_attentions](https://github.com/facebookresearch/xformers/blob/36464229859a177a165d142db788db45cbe6b272/xformers/ops/fmha/triton_splitk.py#L1629) without postprocessing.

```
pytest tests/test_flash_attn.py
...
======================================================================================== 268004 passed, 152064 skipped in 4404.00s (1:13:23) =========================================================================================
```

### DIFF
--- a/csrc/flash_attn/src/flash_fwd_kernel.h
+++ b/csrc/flash_attn/src/flash_fwd_kernel.h
@@ -100,7 +100,7 @@ inline __device__ void compute_attn_1rowblock(const Params &params, const int bi
         #pragma unroll
         for (int m = 0; m < size<1>(tOgO); ++m) {
             const int row = get<0>(tOcO(0, m, 0));
-            if (row < binfo.actual_seqlen_q - m_block * kBlockM && get<1>(tOcO(0, m, 0)) == 0) { gLSE(row) = INFINITY; }
+            if (row < binfo.actual_seqlen_q - m_block * kBlockM && get<1>(tOcO(0, m, 0)) == 0) { gLSE(row) = -INFINITY; }
         }
         return;
     }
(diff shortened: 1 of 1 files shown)


GD06:
QQ: do we plan to merge this PR as it has been pending for months.

sgrigory:
> QQ: do we plan to merge this PR as it has been pending for months.

Sorry, I didn't follow-up on @tridao's comment above. Basically I think there should be no NaNs after this change because the code actually checks for `-inf` before computing `exp(score - lse)` in the backward pass
https://github.com/Dao-AILab/flash-attention/blob/40fa35acd8269ebc4777e682f8bfb690f1f12bb5/csrc/flash_attn/src/softmax.h#L75

Also, in the Hopper kernel we write `-inf` for out-of-bounds positions

https://github.com/Dao-AILab/flash-attention/blob/40fa35acd8269ebc4777e682f8bfb690f1f12bb5/hopper/epilogue_fwd.hpp#L379-L385

https://github.com/Dao-AILab/flash-attention/blob/40fa35acd8269ebc4777e682f8bfb690f1f12bb5/hopper/flash_fwd_kernel_sm90.h#L406

If that makes sense and FA2 code is still relevant, I add a test which cover backward behaviour in this situation to make the PR mergeable

Issue #952, opened 2024-05-14
Value-dim differnet from query-dim/key-dim is not supported


hahnyuan:
Hello flash_attn Maintainers and Community,

When attempting to execute the provided code snippet utilizing the flash_attn library, a runtime error occurs with the following message:

```python
import torch
import flash_attn

bs=1
seqlen=128
qkdim=128
vdim=256
nheads=8
q=torch.ones(bs,seqlen,nheads,qkdim,dtype=torch.float16).cuda()
k=torch.ones(bs,seqlen,nheads,qkdim,dtype=torch.float16).cuda()
v=torch.ones(bs,seqlen,nheads,vdim,dtype=torch.float16).cuda()

attn=flash_attn.flash_attn_func(q,k,v)
```
Result:
`RuntimeError: v must have shape (batch_size, seqlen_k, num_heads_k, head_size_og)`

It seems that the error is due to a mismatch between the dimensions of the value tensor v and the query and key tensors q and k, respectively. 

Is it possible to extend support to scenarios where the dimensionality of the value tensor v differs from that of the query and key tensors q and k?

tridao:
No we don't plan to support that.

Issue #753, opened 2024-01-05
Is it possible to relax V shape requirements to have different head dim than q/k?


Maykeye:
Torch's SDPA doesn't require V to have the same dimensions as inputs, it even noted in docs with different dimensions [E and Ev](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html) as when V is multiplied by, head dimensions is gone and we have only L x L matrix. 

```python
In [23]: qk = torch.randn(4, 4, 4, 8).bfloat16().cuda()  

In [24]: v = torch.randn(4, 4, 4, 16).bfloat16().cuda()

In [25]: F.scaled_dot_product_attention(qk, qk, v).shape
Out[25]: torch.Size([4, 4, 4, 16])
```

same with [xfrormers](https://facebookresearch.github.io/xformers/components/ops.html), they use `K` and `Kv` in doc.

```python
In [26]: xops.memory_efficient_attention(qk, qk,v).shape
Out[26]: torch.Size([4, 4, 4, 16])
```

However flash attention 2 [2.4.2] requires head dimensions to match. 

```python
In [27]: flash_attn.flash_attn_func(qk,qk,v)....
RuntimeError: v must have shape (batch_size, seqlen_k, num_heads_k, head_size_og)
```
(as documented it requires all tensors to have `headdim` per head (error uses different name than documentation))

can it be relaxed to have different head_size for v or implementation depends on head dimensions match?

tridao:
While it's theoretically possible, we don't plan to do that. The reason is that we're already templating on the head dimension (32, 64, 96, 128, 160, 192, 224, 256). If V has a different head dimension we'd need to increase the number of templates by 8x, and compilation time will increase by 8x.

Maykeye:
I see

NiuMa-1234:
> While it's theoretically possible, we don't plan to do that. The reason is that we're already templating on the head dimension (32, 64, 96, 128, 160, 192, 224, 256). If V has a different head dimension we'd need to increase the number of templates by 8x, and compilation time will increase by 8x.

Hi,I found that the latest flash3 only supports head sizes of 64 or 128, are you planning to include more?

PR #980, opened 2024-06-06
[Draft] support qk head_dim different from vo head_dim

defei-coder:
Support query/key head_dim different from  value head_dim, fix [issue-753](https://github.com/Dao-AILab/flash-attention/issues/753 ) and [issue-952](https://github.com/Dao-AILab/flash-attention/issues/952 )
Recently, [DeepSeek-V2](https://github.com/deepseek-ai/DeepSeek-V2/tree/main )proposed a new attention called MLA (Multi-head Latent Attention), which utilizes low-rank key-value union compression to eliminate the bottleneck of inference-time key-value cache, thus supporting efficient inference. MLA will use query/key head_dim=192 and value head_dim=128, but flashAttention not support the combination. Although this can be achieved by padding value head_dim from 128 to 192, but this way will increase global memory and hurt the performence.
In order to expand the versatility of flashAttention, I modify the code to support this ability. For compilation time considerations, only one combination is added, other combinations can be implemented by the user as needed.
Compared with padding value head_dim from 192  to 128, use query/key head_dim=192 and value head_dim=128 will save global memory and improve performence(forward will speedup about 15%, backward will speedup 5%).

### DIFF
--- a/csrc/flash_attn/flash_api.cpp
+++ b/csrc/flash_attn/flash_api.cpp
@@ -43,7 +43,9 @@ void set_params_fprop(Flash_fwd_params &params,
                       float softmax_scale,
                       int window_size_left,
                       int window_size_right,
-                      bool seqlenq_ngroups_swapped=false) {
+                      bool seqlenq_ngroups_swapped=false,
+                      int d_v=0,
+                      int d_v_rounded=0) {
 
     // Reset the parameters
     params = {};
@@ -97,6 +99,8 @@ void set_params_fprop(Flash_fwd_params &params,
     params.seqlen_k_rounded = seqlen_k_rounded;
     params.d = d;
     params.d_rounded = d_rounded;
+    params.d_value = d_v;
+    params.d_value_rounded = d_v_rounded;
 
     // Set the different scale values.
     params.scale_softmax = softmax_scale;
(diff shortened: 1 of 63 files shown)


NiuMa-1234:
Hi, the latest flash-attn3 is released and it supports the head-dim of 64 and 128 only for now, do you plan to support more?

bzantium:
supporting combinations of (256, 256), (128, 256) would be great.

bg51717:
@NiuMa-1234 
Thanks for your work.😊My current work also needs flash_attn to support different qk head_dim from vo head_dim. 
I see this PR is still open. How's the progress? Is it due to failed tests or lack of review? Will you keep improving this PR?

D-Keqi:
@defei-coder Hi, I found your implementation very interesting and helpful. I noticed that you currently support the configuration with Q/K head dimension of 192 and V head dimension of 128. I tried to extend the implementation to support other combinations, such as Q/K head dim = 288 and V head dim = 256.

I found that although the code runs correctly in general, an issue arises under a specific condition: when the number of query heads is 1 and the key/value length exceeds 64, the following error occurs:
packages/flash_attn-2.5.9.post1-py3.12-linux-x86_64.egg/flash_attn/flash_attn_interface.py", line 1201, in flash_attn_with_kvcache
    out, softmax_lse = flash_attn_cuda.fwd_kvcache(
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
RuntimeError: CUDA error: invalid argument
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

If I switch back to the 192 and 128 configuration, this issue disappears. The problem only occurs with the 288 and 256 configuration and when the number of query heads is 1; for example, it runs fine when the number of query heads is 2.

Do you have any thoughts or comments on this based on your experience? Thanks a lot!

PR #989, opened 2024-06-13
Add return_softmax_lse in flash_attn_with_kvcache

ovowei:
This is helpful for combining flash attention with ring attention.


lucidrains:
@ovowei i'm interested in this as well and can help out

PR #1072, opened 2024-07-19
Add var-seq-len to FA3 fp16 / bf16 fwd

ipiszy:
This PR contains following changes:
* Add var-seq-len support to fp16 / bf16 fwd kernel.
* Add unittests for var-seq-len fwd.
* Add a simple benchmark to compare fp16 / bf16 fwd kernel perf across different backends.


goldhuang:
@ipiszy @tridao I'm trying to make varlen fwd working with fp8. Could you help to share the reason that the PR is only for fp16 and bf16, but not working for fp8?

PR #1112, opened 2024-07-31
Add how to import FA3 to documentation.

AdamLouly:
I've had this issue and also seen many people questioning how to use FA3.

usually people try to 
`import as flash_attn.hopper `
or as
`import flashattn_hopper
`
but it should be 
`import flash_attn_interface`


saurabh-kataria:
`from flash_attn_interface import *` would be better, since there you find `flash_attn_func`

PR #1129, opened 2024-08-06
bwd benchmark + small fixes

ipiszy:
Benchmarks:
+ bwd benchmarks: var-seq-len with cudnn and fav3

Small fixes:
* Set -infinity to LSE;

### DIFF
--- a/hopper/benchmark_attn.py
+++ b/hopper/benchmark_attn.py
@@ -48,14 +48,13 @@ def convert_to_cudnn_type(torch_type):
         raise ValueError("Unsupported tensor data type.")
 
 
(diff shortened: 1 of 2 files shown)


hopper/flash_api.cpp
```diff
@@ -283,10 +283,6 @@ mha_fwd(at::Tensor &q,         // batch_size x seqlen_q x num_heads x head_size
 
     CHECK_DEVICE(q); CHECK_DEVICE(k); CHECK_DEVICE(v);
 
-    TORCH_CHECK(q.stride(-1) == 1, "Input tensor must have contiguous last dimension");
```
tridao: We still need the last strides to be 1.

ipiszy: ah right, my mistake.

PR #1139, opened 2024-08-08
Add custom ops for compatibility with PT Compile

ani300:
This PR adds basic support for torch.compile() for the non-varlen variants of Flash Attention.

This essentially allows for models that use the `flash_attn_qkvpacked_func`, `flash_attn_kvpacked_func`, and `flash_attn_func` to compile without graph breaks.

I can add unit tests if it makes sense. I'll test it with our own training pipeline for performance measurements and I'll post them later.

This uses the new custom operators API in Pytorch 2.4. I can move to the older APIs if needed, or I can look up how to make both coexist

### DIFF
--- a/flash_attn/flash_attn_interface.py
+++ b/flash_attn/flash_attn_interface.py
@@ -881,7 +881,7 @@ class FlashAttnVarlenFunc(torch.autograd.Function):
     ):
         if softmax_scale is None:
             softmax_scale = q.shape[-1] ** (-0.5)
-        head_size_og = q.size(3)
+        head_size_og = q.size(2)
         if head_size_og % 8 != 0:
             q = torch.nn.functional.pad(q, [0, 8 - head_size_og % 8])
             k = torch.nn.functional.pad(k, [0, 8 - head_size_og % 8])
--- a/flash_attn/flash_attn_interface.py
+++ b/flash_attn/flash_attn_interface.py
@@ -47,6 +47,7 @@ def _get_block_size_n(device, head_dim, is_dropout, is_causal):
 def round_multiple(x, m):
     return (x + m - 1) / m * m
 
+
 @torch.library.custom_op("flashattn::_flash_attn_forward", mutates_args=(), device_types="cuda")
 def _flash_attn_forward(
     q: torch.Tensor,
@@ -57,7 +58,7 @@ def _flash_attn_forward(
     causal: bool,
     window_size: Tuple[int, int],
     softcap: float,
-    alibi_slopes: torch.Tensor,
+    alibi_slopes: Optional[torch.Tensor],
     return_softmax: bool
 ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
     q, k, v = [maybe_contiguous(x) for x in (q, k, v)]
@@ -105,28 +106,29 @@ def _flash_attn_forward_fake(
     if (return_softmax):
         p = torch.empty((batch_size, num_heads, round_multiple(seqlen_q, 128), round_multiple(seqlen_k, 128)), dtype=q.dtype, device=q.device, layout=q.layout)
     rng_state = torch.empty((2,), dtype=torch.int64, device=q.device)
-    return out, q, k, v, out, softmax_lse, p, rng_state
+    return out[..., :head_size], q, k, v, out, softmax_lse, p, rng_state
 
 
+@torch.library.custom_op("flashattn::_flash_attn_varlen_forward", mutates_args=(), device_types="cuda")
 def _flash_attn_varlen_forward(
-    q,
-    k,
-    v,
-    cu_seqlens_q,
-    cu_seqlens_k,
-    max_seqlen_q,
-    max_seqlen_k,
-    dropout_p,
-    softmax_scale,
-    causal,
-    window_size=(-1, -1),
-    softcap=0.0,
-    alibi_slopes=None,
-    return_softmax=False,
-    block_table=None,
-    leftpad_k=None,
-    seqused_k=None,
-):
+    q: torch.Tensor,
+    k: torch.Tensor,
+    v: torch.Tensor,
+    cu_seqlens_q: torch.Tensor,
+    cu_seqlens_k: torch.Tensor,
+    max_seqlen_q: int,
+    max_seqlen_k: int,
+    dropout_p: float,
+    softmax_scale: float,
+    causal: bool,
+    window_size: Tuple[int, int] = (-1, -1),
+    softcap: float = 0.0,
+    alibi_slopes: Optional[torch.Tensor] = None,
+    return_softmax: bool = False,
+    block_table: Optional[torch.Tensor] = None,
+    leftpad_k: Optional[torch.Tensor] = None,
+    seqused_k: Optional[torch.Tensor] = None,
+) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
     q, k, v = [maybe_contiguous(x) for x in (q, k, v)]
     out, q, k, v, out_padded, softmax_lse, S_dmask, rng_state = flash_attn_cuda.varlen_fwd(
         q,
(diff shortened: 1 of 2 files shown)


ani300:
Added varlen functions too

tridao:
Thank you! If this requires torch 2.4, is there some way to make it optional for older pytorch version? e.g. define `torch.library.custom_op` to be a no-op? Idk what people usually do.

ani300:
There's a different API that was used in previous pytorch versions. It's more cumbersome to use, but let me think about how to use both APIs at the same time with a version selection. I can also do a noop wrapper based on version

ani300:
Ok, I've updated the code to only wrap things if Pytorch is 2.4 or higher. Even though we're comparing strings, `torch.__version__` implements semantic comparisons, so it should always work correctly. As soon as I have the training performance comparison I'll update the PR

mayank31398:
aah, yikes
@ani300 I had started working on the same thing https://github.com/Dao-AILab/flash-attention/pull/1145 😓 

Ill let you handle this 😃

ani300:
@tridao I've updated the code after testing it with SFTTrainer to make sure everything was getting called correctly and compiling without graph breaks. I've also added a check for the pytorch version so this only activates if the APIs are available.

One question I have is if it's ok to change the C++ pytorch interface so the `out`/`out_padded` padding/unpadding can be done outside of the custom op. Right now, Pytorch custom ops don't allow outputs to be aliases of other outputs, which this pair is when `head_dim % 8 == 0`. This means I have to clone one of the outputs to unalias them, which is a waste of memory bandwidth.

tridao:
Yes let's change the C++ code to not pad the output, and have the Python part handle that.

mayank31398:
@GLivshits I dont think it can be handled in older versions of torch

mayank31398:
@tridao @ani300 is there any progress/updates on this?
Its a pretty neat feature to have Flash Attention fully end-to-end traceable natively.

ani300:
Hey, I'm almost done with the C++ portion, hopefully by Monday I can update the PR and rerun all the CI to make sure it's still good with all the changes. I'll also try the suggestions from @GLivshits to Mae it work with previous versions

ani300:
@GLivshits the way the code is currently written, the backward functions need to have `mutate_args`, or we need to do an extra memory copy on GPU, which might significantly impact performance

ani300:
@tridao this is ready for final review and/or merging. I tested with the whole unit test suite for both pytorch 2.3 and pytorch 2.4 on an A100

raghukiran1224:
@tridao Any update on this?

anijain2305:
Thanks @ani300 for taking this on. I started working on this here  (https://github.com/Dao-AILab/flash-attention/pull/1209). But this is way further along. So, I closed mine.

cc @zou3519 for custom ops API usage review @Chillee

flash_attn/flash_attn_interface.py
```diff
@@ -45,11 +46,57 @@ def _get_block_size_n(device, head_dim, is_dropout, is_causal):
...
+                return func(*args, **kwargs)
+            return inner
+        if fn is None:
+            return wrap
+        return wrap(fn)
+    _torch_custom_op_wrapper = custom_op_wrapper
+    _torch_register_fake_wrapper = register_fake_wrapper
+
+
+@_torch_custom_op_wrapper("flashattn::_flash_attn_forward", mutates_args=(), device_types="cuda")
```
anijain2305: Nit - `s/flashattn/flash_attn/`

flash_attn/flash_attn_interface.py
```diff
@@ -58,36 +105,71 @@ def _flash_attn_forward(
...
+    softmax_lse = torch.empty((batch_size, num_heads, seqlen_q), dtype=torch.float32, device=q.device, layout=q.layout)
+    p = torch.empty((0,), dtype=q.dtype, device=q.device, layout=q.layout)
+    if return_softmax:
+        p = torch.empty((batch_size, num_heads, round_multiple(seqlen_q, 128), round_multiple(seqlen_k, 128)), dtype=q.dtype, device=q.device, layout=q.layout)
+    rng_state = torch.empty((2,), dtype=torch.int64, device=q.device)
+
+    return out, softmax_lse, p, rng_state
+
+
+try:
```
anijain2305: Perhaps use the same if condition

```
if torch.__version__ >= "2.4.0":
```

flash_attn/flash_attn_interface.py
```diff
@@ -45,11 +46,57 @@ def _get_block_size_n(device, head_dim, is_dropout, is_causal):
...
+
+if torch.__version__ >= "2.4.0":
+    _torch_custom_op_wrapper = torch.library.custom_op
+    _torch_register_fake_wrapper = torch.library.register_fake
+else:
+    def custom_op_wrapper(name, fn=None, /, *, mutates_args, device_types=None, schema=None):
+        def wrap(func):
+            def inner(*args, **kwargs):
+                return func(*args, **kwargs)
+            return inner
```
anijain2305: Is this equivalent of `return func`?

flash_attn/flash_attn_interface.py
```diff
@@ -45,11 +46,57 @@ def _get_block_size_n(device, head_dim, is_dropout, is_causal):
...
+        if fn is None:
+            return wrap
+        return wrap(fn)
+    def register_fake_wrapper(op, fn=None, /, *, lib=None, _stacklevel=1):
+        def wrap(func):
+            def inner(*args, **kwargs):
+                return func(*args, **kwargs)
+            return inner
```
anijain2305: Ditto - `return func`

flash_attn/flash_attn_interface.py
```diff
@@ -45,11 +46,57 @@ def _get_block_size_n(device, head_dim, is_dropout, is_causal):
...
+        return torch_op
+    else:
+        return func
+
+
+if torch.__version__ >= "2.4.0":
+    _torch_custom_op_wrapper = torch.library.custom_op
+    _torch_register_fake_wrapper = torch.library.register_fake
+else:
+    def custom_op_wrapper(name, fn=None, /, *, mutates_args, device_types=None, schema=None):
```
anijain2305: Nit - a more desciptive name - noop_custom_op_wrapper

And maybe a comment that we dont support < 2.4 versions.

flash_attn/flash_attn_interface.py
```diff
@@ -45,11 +46,57 @@ def _get_block_size_n(device, head_dim, is_dropout, is_causal):
...
+        def wrap(func):
+            def inner(*args, **kwargs):
+                return func(*args, **kwargs)
+            return inner
+        if fn is None:
+            return wrap
+        return wrap(fn)
```
anijain2305: `return fn`

anijain2305:
Thanks @ani300 

Custom ops changes look good to me. Hope @tridao  can take a look at the other parts.

ani300:
Tests all ran successfully again

tridao:
Awesome, thanks!

umarbutler:
I just built the flash-attention main branch from source and sadly I'm getting the same error message during training with the Hugging Face Trainer with `torch_compile=True`:
```
.venv/lib/python3.12/site-packages/torch/_dynamo/variables/functions.py:663: UserWarning: Graph break due to unsupported builtin flash_attn_2_cuda.PyCapsule.fwd. This function is either a Python builtin (e.g. _warnings.warn) or a third-party C/C++ Python extension (perhaps created with pybind). If it is a Python builtin, please file an issue on GitHub so the PyTorch team can add support for it and see the next case for a workaround. If it is a third-party C/C++ Python extension, please either wrap it into a PyTorch-understood custom operator (see https://pytorch.org/tutorials/advanced/custom_ops_landing_page.html for more details) or, if it is traceable, use torch.compiler.allow_in_graph.
```

umarbutler:
It looks like I somehow ran `python setup.py bdist_wheel` and forgot I was creating a wheel (which I also ended up mistakenly deleting moments ago) 😅 hopefully second time's the charm

ani300:
I was worried for a second, as I just used it with the Huggingface Trainer yesterday successfully 😥

umarbutler:
@ani300 I can confirm that it works! And much faster than PyTorch SDPA, shaved off 50 training hours (400h -> 350h) 😆

vadimkantorov:
Does it also work for fullgraph?

Thanks!

PR #1147, opened 2024-08-13
Clearer install instructions for CUDA and ROCm backends

garrettbyrd:
The current README has two `Installation and Features` sections (one for "normal" flash attention with CUDA backend) and one for AMD GPUs. Both `Installation and Features` were at header level 2, and the `AMD GPU Support` section was arbitrarily far down away from the other installation instructions. The AMD installation instructions also failed to mention `packaging` and `ninja` as requirements.

This updated README consolidates CUDA and ROCm backends under the single `Installation and Features` section near the top.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,26 +63,21 @@ pytest -q -s test_flash_attn.py
 
 
 ## Installation and features
-
-Requirements:
-- CUDA 11.6 and above.
+**Requirements:**
+- A GPU backend (CUDA or ROCm)
 - PyTorch 1.12 and above.
+- `packaging` Python package (`pip install packaging`)
+- `ninja` Python package (`pip install ninja`) *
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,26 +63,21 @@ pytest -q -s test_flash_attn.py
 
 
 ## Installation and features
-
-Requirements:
-- CUDA 11.6 and above.
+**Requirements:**
+- A GPU backend (CUDA or ROCm)
```
tridao: Let's phrase it as "CUDA toolkit or ROCm toolkit" to avoid people thinking you need a GPU to install. E.g. you just need nvcc, not the GPU itself.

garrettbyrd: Done. Also added "using CUDA toolkit" in reference to the `ninja` compile times. AMD compile times are still pretty long even with `ninja`.

PR #1180, opened 2024-08-27
Add ArchTag to pre/postprocess bwd kernels

cameronshinn:
Certain versions of CUTLASS are having compiler errors related to the additional backwards kernels not having an `ArchTag` defined. When you launch the kernel it will check this type for feature compatibility. I'm compiling with a different CUTLASS version, but when FA updates its version errors will start appearing too. These changes just copy the semantics from [`CollectiveMainloopBwd`](https://github.com/Dao-AILab/flash-attention/blob/c92ca63268836a6ea519e9892eda4895fc31811d/hopper/mainloop_bwd_sm90_tma_gmma_ws.hpp#L37) into the pre and postprocess bwd kernels.

### DIFF
--- a/hopper/flash_bwd_postprocess_kernel.h
+++ b/hopper/flash_bwd_postprocess_kernel.h
@@ -18,7 +18,7 @@ namespace flash {
 
 using namespace cute;
 
-template <class TileShape_MK_, class Element, class ElementAccum, int kNThreads, class SmemLayoutdQaccumTMA,
+template <class TileShape_MK_, class Element, class ElementAccum, class ArchTag_, int kNThreads, class SmemLayoutdQaccumTMA,
           class TiledMma, bool dQ_swapAB>
 class FlashAttnBwdPostprocessConvertdQ {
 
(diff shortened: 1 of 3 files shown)


hopper/flash_bwd_preprocess_kernel.h
```diff
@@ -17,13 +17,16 @@ namespace flash {
...
-template <class TileShape_MK_, class Element, class ElementAccum, bool Clear_dQaccum, bool Varlen>
+template <class TileShape_MK_, class Element, class ElementAccum, class ArchTag_, bool Clear_dQaccum, bool Varlen>
 class FlashAttnBwdPreprocess {
 
 public:
 
     // Type Aliases
     using TileShape_MK = TileShape_MK_;
+    using ArchTag = ArchTag_;
+
```
tridao, line 28: This kernel doesn't need SM90. It needs SM75 (if fp16), SM80 (if bf16), and SM89 (if fp8). Is there an easy way to rewrite this check depending on type?

cameronshinn: How does this look? It seems to work for me. I added it to the other kernel as well, assuming that has the same compute capability requirements.

hopper/flash_bwd_postprocess_kernel.h
```diff
@@ -18,14 +18,19 @@ namespace flash {
...
           class TiledMma, bool dQ_swapAB>
 class FlashAttnBwdPostprocessConvertdQ {
 
 public:
 
     // Type Aliases
     using TileShape_MK = TileShape_MK_;
+    using ArchTag = ArchTag_;
+
+    static_assert(std::is_same_v<Element, cutlass::half_t> && ArchTag::kMinComputeCapability >= 75 ||
```
tridao: This one does need SM90 since it's using TMA

PR #1209, opened 2024-09-06
[rfc][torch.compile] Make custom kernels torch.compile compatible

anijain2305:
This PR is incomplete (and very possibly buggy). This is my first attempt to make the CUDA kernels torch.compile compatible. So, PR is not ready for full review yet.

But, I am looking for a very high level feedback on GO or NO-GO. I am making some changes on the C++ side to make the ops pt2 compatible. Are these acceptable?

cc @zou3519 and @Chillee


anijain2305:
Just saw https://github.com/Dao-AILab/flash-attention/pull/1139 is already way ahead. So, lets move the discussion there. I will close this PR.

PR #1233, opened 2024-09-16
Add local attention in Hopper FAv3

ipiszy:
## PR Description
ATT, add local_window support in Hopper FAv3 and unittests.
* Local masking is checked in each mainloop iteration. It's possible to further optimize it by only checking local masking in the first m and last n iterations, and leaving iterations in-between unchecked. This PR doesn't do this optimization to make code easier to read. 
* This support is only added to fp16 / bf16 for now. fp8 support will be in separate PRs.

## Tests
Added unittest for local attention.
`python3.9 -m pytest -q -s test_flash_attn.py` passed.

Benchmark: `python3.9 benchmark_attn.py`
No perf difference before v.s. after.


hopper/flash_bwd_kernel.h
```diff
@@ -247,7 +249,7 @@ class FlashAttnBwd {
                     auto block_coord = work_tile_info.get_block_coord(params.scheduler);
                     auto [n_block, bidh, bidb] = block_coord;
                     if constexpr (Varlen) {
-                        if (n_block * kBlockM >= collective_mainloop.get_seqlen_k(params.mainloop, bidb)) { continue; }
+                        if (n_block * kBlockN >= collective_mainloop.get_seqlen_k(params.mainloop, bidb)) { continue; }
                     }
                     if constexpr (Is_causal) {
                         int const m_block_min = collective_mainloop.get_m_block_min(params.mainloop, n_block, bidb);
```
tridao, line 255: We'd need the same change to m_block_max here?

hopper/flash_bwd_launch_template.h
```diff
@@ -170,9 +171,11 @@ template<typename T>
 void run_mha_bwd_hdim64(Flash_bwd_params &params, cudaStream_t stream) {
     constexpr static int Headdim = 64;
     BOOL_SWITCH(params.is_causal, Is_causal, [&] {
-        BOOL_SWITCH(params.cu_seqlens_q != nullptr || params.cu_seqlens_k != nullptr, Varlen, [&] {
-            BOOL_SWITCH(params.deterministic, Deterministic, [&] {
-                run_flash_bwd<Headdim, 128, 128, T, Is_causal, Varlen, Deterministic, false, false, 1, 2, 2>(params, stream);
+        BOOL_SWITCH(params.is_local, Is_local, [&] {
+            BOOL_SWITCH(params.cu_seqlens_q != nullptr || params.cu_seqlens_k != nullptr, Varlen, [&] {
+                BOOL_SWITCH(params.deterministic, Deterministic, [&] {
+                    run_flash_bwd<Headdim, 128, 128, T, Is_causal, Is_local, Varlen, Deterministic, false, false, 1, 2, 2>(params, stream);
```
tridao: Let's template with `Is_local && !Is_causal` to reduce the number of kernels.
And add an static_assert(!(Is_causal && Is_local)) inside `run_flash_bwd`.
i.e. it's either "causal" or "local", not both.

hopper/flash_bwd_launch_template.h
```diff
@@ -182,9 +185,11 @@ template<typename T>
 void run_mha_bwd_hdim96(Flash_bwd_params &params, cudaStream_t stream) {
     constexpr static int Headdim = 96;
     BOOL_SWITCH(params.is_causal, Is_causal, [&] {
-        BOOL_SWITCH(params.cu_seqlens_q != nullptr || params.cu_seqlens_k != nullptr, Varlen, [&] {
-            BOOL_SWITCH(params.deterministic, Deterministic, [&] {
-                run_flash_bwd<Headdim, 64, 128, T, Is_causal, Varlen, Deterministic, false, false, 1, 2, 1>(params, stream);
+        BOOL_SWITCH(params.is_local, Is_local, [&] {
+            BOOL_SWITCH(params.cu_seqlens_q != nullptr || params.cu_seqlens_k != nullptr, Varlen, [&] {
+                BOOL_SWITCH(params.deterministic, Deterministic, [&] {
+                    run_flash_bwd<Headdim, 64, 128, T, Is_causal, Is_local, Varlen, Deterministic, false, false, 1, 2, 1>(params, stream);
```
tridao: Same templating `Is_local && !Is_causal` here and below

hopper/mainloop_bwd_sm90_tma_gmma_ws.hpp
```diff
@@ -592,6 +613,15 @@ struct CollectiveMainloopBwd {
             }
             cutlass::arch::NamedBarrier::arrive(kNThreadsdQ + cutlass::NumThreadsPerWarp, static_cast<int>(BwdNamedBarriers::dQEmpty) /*id*/);  // sdQ empty, ready to be written to
         }
+        if constexpr (Deterministic) {
```
tridao: Let's condition on `Is_local` here

hopper/flash_fwd_launch_template.h
```diff
@@ -115,11 +117,13 @@ template<typename T>
...
-        SEQLEN_SWITCH(params.cu_seqlens_q, Seqlen_traits, [&] {
-            run_flash_fwd<
-                Flash_fwd_kernel_traits<Headdim, 192, 128, 16, 2, false, 1, T>, 
-                Is_causal, Seqlen_traits
-            >(params, stream);
+        BOOL_SWITCH(params.is_local, Is_local, [&] {
+            SEQLEN_SWITCH(params.cu_seqlens_q, Seqlen_traits, [&] {
+                run_flash_fwd<
+                    Flash_fwd_kernel_traits<Headdim, 192, 128, 16, 2, false, 1, T>, 
+                    Is_causal, Is_local, Seqlen_traits
```
tridao: same here, let's template with `Is_local && !Is_causal`

hopper/flash_bwd_kernel.h
ipiszy: Good catch! Missed this.

hopper/mainloop_bwd_sm90_tma_gmma_ws.hpp
ipiszy: Why is it necessary to condition on Is_local? I think Deterministic is also applied to local?

ipiszy (commented):
Thanks @tridao !

hopper/mainloop_bwd_sm90_tma_gmma_ws.hpp
tridao: Yes but without local we wound't need additional calls to `Barrier::arrive_inc` right?

ipiszy: Ah right, my brain is broken lol.

PR #1236, opened 2024-09-18
FA3 kvcache + split kv + gqa parallelization

jayhshah:
This PR adds split KV ("Flash decoding") and GQA parallelization improvements for FA3. Some essential parts of the KV cache API are added as well, including the `cache_seqlens` and `cache_batch_idx` arguments.

Up to 15x improvement over FA2 measured on my H100 PCIe in exceptional cases, e.g.

```
DTYPE: FLOAT16, CAUSAL, QHEADS:16, KVHEADS:1, HEADDIM:128
CONTEXT:16384, BSZ:4, QLEN:4, FA2:402.86, FA3:26.93, NUM SPLITS:22, RATIO:14.96, GB/s:1245.77
```

Times given in microseconds. GB/s is measured in terms of loading the KV cache. Note that theoretical max bandwidth is 2 TB/s for H100 PCIe. 

TODO on this PR before merge: add split kv heuristic, implement for FP8.
 
[fa3-decoding-times-091724.log](https://github.com/user-attachments/files/17037681/fa3-decoding-times-091724.log)


hopper/flash_attn_interface.py
```diff
@@ -174,7 +175,8 @@ def forward(
             causal,
             descale_q=descale_q,
             descale_k=descale_k,
-            descale_v=descale_v,
+            descale_v=descale_v,     
+            gqa_decoding=False,
```
ipiszy: I wonder does it make sense to give user an option to enable GQA optimization for general use cases outside of decoding?

e.g. It's generally useful for small seq_len prefill. In this case we don't really need split-kv, but we want to have each threadblock handle multiple Q heads with the same KV head.

ipiszy: Furthermore, does it make sense to just enable GQA optimization by default when input is GQA? I feel it won't cause perf regressions even for long sequence length.

tridao: I feel it might slow things down a bit, but I haven't tried

csrc/composable_kernel
tridao, line 1: we probably don't want to update this?

jayhshah: Sorry, don't know what happened here, maybe rebase problem? I'll get rid of this.

jayhshah: Issue should be resolved now, I think I forgot to update the submodule when rebasing. Commit was showing a reversion from main.

hopper/test_flash_attn.py
```diff
@@ -23,32 +23,158 @@ def print_diffs(out, out_ref):
...
+    # dS = torch.einsum('bthd,bshd->bhts', g.float(), v.float())
+    # P = torch.softmax(qk, -1)
+    # dP = P * (dS - do_o.unsqueeze(1))
+    # dQ = torch.einsum('bhts,bshd->bthd', dP, k.float())
+    # dV = torch.einsum('bhts,bthd->bshd', P, g.float())
+    # dK = torch.einsum('bhts,bthd->bshd', dP, q.float())
+    # breakpoint()
+    
+    # just test correctness of fp8 kernel w/o further quantization techniques
+    assert (out - out_ref).abs().max().item() <= 40 * (out_pt - out_ref).abs().max().item()
```
ipiszy: 40 seems to be very big, maybe add something like `torch.testing.assert_close()` which allows specifying a meaningful `atol` and `rtol`.

hopper/flash_api.cpp
```diff
@@ -236,36 +240,220 @@ void set_params_dgrad(Flash_bwd_params &params,
...
+            
+            int block_n = 128;
+            if (head_size == 128 && !is_causal) {
+                block_n = 176;
+            } else if (head_size == 256) {
+                block_n = use_one_mma_wg ? 96 : 80;
+            }
+            const int num_n_blocks = (max_seqlen_k + block_n - 1) / block_n;
+            
+	        int batch_nheads = use_gqa_decoding ? batch_size * num_heads_k : batch_size * num_heads;
```
ipiszy: indentation

hopper/flash_api.cpp
```diff
@@ -236,36 +240,220 @@ void set_params_dgrad(Flash_bwd_params &params,
...
+    // Goal of the starting threshold is to determine whether to split or not.
+    // Empirically, the efficiency threshold can be much lower than 80% depending on num_n_blocks.
+    int num_m_blocks = batch_nheads_mblocks/batch_nheads;
+    float start_threshold;
+    float num_n_blocksf = float(num_n_blocks);
+    if (head_size == 128) {
+        if (std::log2f(num_n_blocksf) <= 4) { // 2048 -- .25
+            start_threshold = .20f + (std::log2f(num_n_blocksf) - 3) * .05f;
+        } else if (std::log2f(num_n_blocksf) <= 5) { // 4096 -- .25
+            start_threshold = .25f;
```
ipiszy, line 259: This is different from the others.

hopper/flash_api.cpp
```diff
@@ -236,36 +240,220 @@ void set_params_dgrad(Flash_bwd_params &params,
...
+inline int num_splits_heuristic(int batch_nheads_mblocks, int batch_nheads, int num_SMs, int num_n_blocks,
+    int max_splits, int head_size, bool use_one_mma_wg) {
+    // Goal of the starting threshold is to determine whether to split or not.
+    // Empirically, the efficiency threshold can be much lower than 80% depending on num_n_blocks.
+    int num_m_blocks = batch_nheads_mblocks/batch_nheads;
+    float start_threshold;
+    float num_n_blocksf = float(num_n_blocks);
+    if (head_size == 128) {
+        if (std::log2f(num_n_blocksf) <= 4) { // 2048 -- .25
+            start_threshold = .20f + (std::log2f(num_n_blocksf) - 3) * .05f;
```
ipiszy, line 257: Add some comments to describe how are these numbers generated?

hopper/seq_len.h
```diff
@@ -4,15 +4,25 @@
...
 #include <cute/layout.hpp>
 
 namespace flash {
 
 static constexpr int kMaxTileSize = 128;
 
-template <bool UseVarSeqLen> class SeqLenTraits {
+static constexpr int FixedSeqLenType = 0;
+static constexpr int VarSeqLenType = 1;
+static constexpr int DecodingGQASeqLenType = 2;
```
ipiszy: Why cannot it be both var-seq-len and decoding-gqa? I think at least the API should be prepared for this use case. There are lots of production workloads which use var-seq-len.

hopper/flash_fwd_launch_template.h
```diff
@@ -106,148 +152,465 @@ void run_flash_fwd(Flash_fwd_params &params, cudaStream_t stream) {
...
+      static_assert(is_same_v<ElementAccum, float>, "ElementAccum must be float.");
+      // We want kBlockM to be as small as possible for more parallelism.
+      // With 128 threads we can load 512 elements at a time, so if headdim is divisible by 128, kBlockM = 4.
+      // If headdim is divisible by 64, then we set kBlockM = 8, etc.
+      constexpr static int kHeadDim = Kernel_traits::kHeadDim;
+      constexpr static int kBlockM = kHeadDim % 128 == 0 ? 4 : (kHeadDim % 64 == 0 ? 8 : 16);
+      constexpr static bool Is_even_K = true; // always true for our current setting
+      void *kernel_combine;
+      int smem_size_combine;
+      if (params.num_splits <= 2) {
```
ipiszy: define a macro inside static_switch?

hopper/flash_fwd_launch_template.h
```diff
@@ -106,148 +152,465 @@ void run_flash_fwd(Flash_fwd_params &params, cudaStream_t stream) {
...
-                run_flash_fwd<
-                    Flash_fwd_kernel_traits<Headdim, 192, 128, 16, 2, false, 1, T>, 
-                    Is_causal, Is_local && !Is_causal, Seqlen_traits
-                >(params, stream);
+      BOOL_SWITCH(params.is_local, Is_local, [&] {
+        MMA_3WG_SWITCH(params.seqlen_q, kNumMmaWGs, [&] {
+          SEQLEN_SWITCH(params.cu_seqlens_q, Seqlen_traits, [&] {
+            BOOL_SWITCH(params.num_splits > 1, Is_split, [&] {
+              run_flash_fwd<
+                  Flash_fwd_kernel_traits<Headdim, kNumMmaWGs * 64, 128, 4 + kNumMmaWGs * 4,
```
ipiszy, line 204: Adjusting Q tile size would make it more efficient for small seq len, thanks for the fix~

hopper/seq_len.h
```diff
@@ -4,15 +4,25 @@
...
+
+template <int SeqLenType, bool Is_dynamic_> class SeqLenTraits {
```
ipiszy: What does `Is_dynamic_` mean here? How is it different from var-seq-len?

hopper/mainloop_fwd_sm90_tma_gmma_ws.hpp
```diff
@@ -454,25 +540,41 @@ struct CollectiveMainloopFwd {
...
                 copy(mainloop_params.tma_load_V.with(*pipeline_v.producer_get_barrier(smem_pipe_write), mcast_mask_kv),
                     tVgV(_, n_block), tVsV(_, smem_pipe_write.index()));
             }
+        }
 
-            shared_storage.barrier_O.wait((work_idx + 1) % 2);            
-                        
+        // if constexpr(Is_causal || Is_local || Is_split) {
+        {
+            #if 0
```
ipiszy: Why not just removing deprecated code?

ipiszy (commented):
Thanks @jayhshah @ganeshcolfax .
It seems that gqa can be used without split-k, is it correct? I wonder what's the perf if it's just enabled by default, could you test it using `benchmark_attn.py`? If it's suitable for general use cases, it seems that code can be simplified quite a bit.

hopper/mainloop_fwd_sm90_tma_gmma_ws.hpp
jayhshah: Removed it -- I thought it was convenient to leave in when testing fp8 kernel perf, but it does clutter the code more.

hopper/seq_len.h
jayhshah: I meant to include a template parameter to guard against reading `actual_seq_len` when not needed as that degrades fp8 kernel perf. Actual seqlen can differ per batch with both fixed or variable seq len. Do you have a better name for this?

hopper/flash_api.cpp
jayhshah: Fixed and changed int to const int.

hopper/flash_api.cpp
jayhshah: Specific numbers were trial and error but there were some broad correlations I observed with respect to increasing batch size versus number of splits (in this case, determining when to not split at all!). I will include comments on this in the code before merging.

hopper/flash_api.cpp
jayhshah: It does look odd but was what I observed to work with some trial and error.

hopper/flash_fwd_launch_template.h
jayhshah: Done, thanks for the suggestion!

hopper/seq_len.h
ipiszy: Why could actual seqlen differ per batch for fixed seq len?

jayhshah: I'm using `FixedSeqLenTraitsDynamic` in the kv cache setup where actual seqlen is <= cache seqlen. I only need actual seqlen to be updated every batch and don't need the other layout changes. However, I suppose `VarSeqLenTraits` can be used for this instead -- let me look into it.

jayhshah: In the kv cache setup I don't have access to `total_k`, so won't be able to use `VarSeqLenTraits` layouts.

hopper/seq_len.h
jayhshah: Can be both, I changed the SeqLenTraits class to make it easier to enable this in a future PR.

hopper/test_flash_attn.py
jayhshah: Changed to 

```
atol = 4 * (out_pt - out_ref).abs().max().item()
torch.testing.assert_close(out, out_ref, rtol=1e-1, atol=atol, check_dtype=False)
```

hopper/seq_len.h
jayhshah: In the latest update I removed `Is_dynamic_` entirely. I found another way to handle perf regression with the fp8 kernel, through introducing another init method `init_no_guard` that involves no conditioning on the pointer not being nullptr internally, but instead moving that logic outside and using a boolean in smem.

PR #1268, opened 2024-10-10
Paged Attention support for FA3

kadeng:
Adding support for Paged Attention / block table to the Flash-Attention 3 Kernel.

 Limits:
  - ~~Not yet integrated with #1236 which modifes the same code (and more).~~
  - No fp8 support yet - as that support also seems missing in the non-paged varlen version so far.

Test Plan:

```
cd hopper
pytest test_flash_attn.py -k "test_flash_attn_varlen_paged"  -s
```

**Update**
 - Rebased on top of main, which includes #1236 now
 - Original version of this PR ( not rebased ) still available on [this branch](https://github.com/kadeng/flash-attention/tree/fa3-paged-original)

### DIFF
--- a/hopper/copy_paged_sm90_tma.hpp
+++ b/hopper/copy_paged_sm90_tma.hpp
@@ -1,0 +1,391 @@
+
+#pragma once
+
+#include <cute/arch/copy_sm90_tma.hpp>
+#include <cute/atom/copy_traits_sm90_tma.hpp>
+
+struct PagedCopyArgs {
+
+  CUTE_HOST_DEVICE
+  PagedCopyArgs() : block_table_batch_stride{0}, page_block_size(0), block_table(nullptr)  {
+  };
+
+  CUTE_HOST_DEVICE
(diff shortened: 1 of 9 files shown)


alexngng:
Hi, do you have plans to support block table of flash_attn_with_kvcache()?

kadeng:
Hi Alex,

I think that makes sense and should be straightforward to implement. The
main difference would be that the version in here is based on the varlen
variant and flash_attn_with_kvcache would be based off the fixed length
variant + sequence length info for KV. So the difference would be mostly in
the memory format of the Q Tensor (varlen packed or not). But I would not
put it into this PR, it is large enough already. I want to get this merged
first.

best,

Kai

Alex Ng ***@***.***> schrieb am Di., 22. Okt. 2024, 09:11:

alexngng:
Thank you for your response!

kadeng:
I was investigating a flaky test failure that I was seeing on this PR. I could narrow it down to a preexisting issue: The flash attention varlen implementation does not seem to work properly yet for d=256. I have disabled testing for d=256 for now, similar to how it's handled in hopper/test_flash_attn.py::test_flash_attn_varlen_output

PR #1284, opened 2024-10-18
Feat: Add support for PyTorch 2.5 in workflows

NanoCode012:
PyTorch 2.5 has just been released: https://pytorch.org/blog/pytorch2-5/

It supports cuda 11.8->12.4

image

I also changed PyTorch 2.4 to support cuda 12.4 too

image

PR #1326, opened 2024-11-10
[Fix] Use CUDA 12.4 in the build system so that the current build system stops failing

EthanSteinberg:
Very recently, a pull request was merged to add PyTorch 2.5 support (https://github.com/Dao-AILab/flash-attention/pull/1284)

However, this pull request seems to have accidentally broken the build system, which I uncovered when trying to build wheels for PyTorch 2.5.

The problem is that the build system builds with CUDA 12.3, but PyTorch doesn't have any 12.3 
releases. See https://github.com/EthanSteinberg/flash-attention/actions/runs/11762473443/job/32765408364#step:8:56 for the failed run.

This PR simply changes the build system to build against 12.4.1 instead.

As an aside, if this PR gets accepted, I would really appreciate if you can do a flash-attention 2.6.4 release that has PyTorch 2.5 support.

I have prepared a cherry pick branch for a 2.6.4 that is the minimal change over 2.6.3 that only has the two commits needed for a PyTorch 2.5 release: https://github.com/EthanSteinberg/flash-attention/tree/cuda_124_rebase

https://github.com/EthanSteinberg/flash-attention/actions/runs/11762648654 is my test run that this appears to build. (Note that it still failed because I don't have permission to upload artifacts. But the wheel did build successfully).

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -45,7 +45,7 @@ jobs:
           os: [ubuntu-20.04]
           python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
           torch-version: ['2.1.2', '2.2.2', '2.3.1', '2.4.0', '2.5.1']
-          cuda-version: ['11.8.0', '12.3.2']
+          cuda-version: ['11.8.0', '12.4.1']
           # We need separate wheels that either uses C++11 ABI (-D_GLIBCXX_USE_CXX11_ABI) or not.
           # Pytorch wheels currently don't use it, but nvcr images have Pytorch compiled with C++11 ABI.
           # Without this we get import error (undefined symbol: _ZN3c105ErrorC2ENS_14SourceLocationESs)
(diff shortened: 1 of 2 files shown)


tridao:
We've built fine with CUDA 12.3 before (e.g. torch 2.4) even though pytorch has never had a release with 12.3.
I'm not sure I understand the problem here.

EthanSteinberg:
@tridao 

The problem is that https://github.com/Dao-AILab/flash-attention/pull/1284/files#diff-551d1fcf87f78cc3bc18a7b332a4dc5d8773a512062df881c5aba28a6f5c48d7R122 updated the version limits so that the code tries to pull 12.3 even when it doesn't exist. 

Previously there was a maximum limit of 12.1, so it would pull 12.1 even when building with 12.3

But that maximum got increased to 12.4 so the code started pulling 12.3.

tridao:
Ah I see, I understand now. Thanks!

tridao:
Yes i'm running tests right now to cut a release

PR #1333, opened 2024-11-13
wrap func into torch ops to avoid torch.compile graphbreaks

kumarkrishna:
This version fixes the graph breaks issue with torch.compile. It is similar in spirit to #1139 and passes all tests and benchmarks.

### DIFF
--- a/hopper/flash_api.cpp
+++ b/hopper/flash_api.cpp
@@ -461,10 +461,10 @@ void run_mha_fwd(Flash_fwd_params &params, cudaStream_t stream, bool force_split
 }
 
 std::vector<at::Tensor>
-mha_fwd(at::Tensor &q,         // batch_size x seqlen_q x num_heads x head_size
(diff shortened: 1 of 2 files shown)


vadimkantorov:
@tridao Are there any CI/tests for testing torch.compile-ability (including fullgraph)?

To know at least which scenarios are now working and which are not. E.g. from HF:
https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_flash_attention_utils.py#L390-L392

```python
    # This is a limitation of flash attention API, as the function `flash_attn_varlen_func`
    # requires `max_length_q`, `max_length_k` to be passed as `int` and not `torch.Tensor`.
    max_length_q = max_length_q.item()
```

Basically, this determines which HF models can be torch.compile-d/fullgraph-d (using `attn_implementation='flash_attention_2'`/`attn_implementation='flash_attention_3'`) - and for benchmarking different attention implementations in this realistic setting

PR #1353, opened 2024-11-22
Added a Benchmark for Rotary and Improved Rotary Performance

alexkranias-amd:
- changed rotary block tiling config resulting in 15% speed up
- added a benchmarking script to test rotary performance


alexkranias-amd:
Apologies. Meant to pull this into a forked repo.

PR #1383, opened 2024-12-12
don't save inputs/outputs buffer of FlashAttenFunc to reduce memory usage for inference mode

XiaobingSuper:
In inference mode, we don't need to save the inputs/outputs used by the training mode. This PR can reduce memory usage during LLM serving(such as vllm, it uses those flash attn APIs for a better performance).

### DIFF
--- a/flash_attn/flash_attn_interface.py
+++ b/flash_attn/flash_attn_interface.py
@@ -452,6 +452,7 @@ class FlashAttnQKVPackedFunc(torch.autograd.Function):
         deterministic,
         return_softmax,
     ):
+        is_grad = torch.is_grad_enabled() and qkv.requires_grad
         if softmax_scale is None:
             softmax_scale = qkv.shape[-1] ** (-0.5)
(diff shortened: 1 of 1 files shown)


XiaobingSuper:
> @XiaobingSuper I found this PR make pytest (https://github.com/Dao-AILab/flash-attention/tree/main/tests) fail.
> 
> https://github.com/Dao-AILab/flash-attention/blob/0dfb28174333d9eefb7c1dd4292690a8458d1e89/flash_attn/flash_attn_interface.py#L455
> 
> torch.is_grad_enabled() always be False in the test script, which make backward cannot access ctx.saved_tensors

Sorry, it is my mistake, gradient computation is already disabled in custom autograd.Functions by default(https://discuss.pytorch.org/t/is-torch-no-grad-making-a-difference-in-custom-autograd-functions/186627), it need to check it before calling the customer op.

XiaobingSuper:
@rocking5566 I created a PR https://github.com/Dao-AILab/flash-attention/pull/1397 to fix this issue. Thanks.

PR #1384, opened 2024-12-12
Support dedicated compile[For Research]

AllenDou:
Support dedicated compile, HEADDIM=64 DTYPE=fp16 ENABLE_SM90=FALSE only build for headdim(64) fp16 and non-sm_90, this method is only for research, could reduce compile time from 50mins -> 5mins.

compile command
```
MAX_JOBS=8 NVCC_THREADS=2 HEADDIM=64 DTYPE=fp16 ENABLE_SM90=FALSE time pip install -e . -v
```
run tests
```
pytest tests/test_flash_attn.py -k test_flash_attn_kvcache -s
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,6 +94,11 @@ variable `MAX_JOBS`:
 MAX_JOBS=4 pip install flash-attn --no-build-isolation
 ```
 
+**Dedicated build(only for research):**
+```sh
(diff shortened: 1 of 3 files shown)


KevinZeng08:
Hi, what is the status of this PR? It seems to reduce compilation time which is very helpful.

Issue #1390, opened 2024-12-16
a small typo and fix


liguohao96:
I compiled from source and run a simple test but failed at this line https://github.com/Dao-AILab/flash-attention/blob/0dfb28174333d9eefb7c1dd4292690a8458d1e89/flash_attn/flash_attn_interface.py#L853
with the following error: `unboundLocalError: cannot access local variable 'out' where it is not associated with a value` 

I believe it is caused by the wrong ident of this line
https://github.com/Dao-AILab/flash-attention/blob/0dfb28174333d9eefb7c1dd4292690a8458d1e89/flash_attn/flash_attn_interface.py#L852

we should move  this line out of  `if` block.

tridao:
I see, thanks for catching it. Can you send a PR?

jiagaoxiang:
Could you please merge this PR ASAP to unblock? Thank you very much!

PR #1392, opened 2024-12-17
fix a bug (issue #1390) caused by typo

liguohao96:
move `flash_attn/flash_attn_interface.py` line 852 out of `if block`.


tridao:
This has been fixed. Thank you.

PR #1394, opened 2024-12-18
Create PEP 517 build metadata

frostming:
Add pyproject.toml with the build system table.

The default doesn't work because the project has additional dependency on `torch` for build


rgommers:
This looks like the right thing to do indeed - it's also a duplicate of gh-958.

PR #1397, opened 2024-12-19
check torch.is_grad_enabled before calling customer flash atten ops

XiaobingSuper:
https://github.com/Dao-AILab/flash-attention/pull/1383 do a ```torch.is_grad_enabled``` check to avoid temporary tensor save when doing inference, but for torch customer ops using ```torch.autograd.Function```, the autograd is always disabled, and it will cause failed for training path.

### DIFF
--- a/flash_attn/flash_attn_interface.py
+++ b/flash_attn/flash_attn_interface.py
@@ -452,7 +452,7 @@ class FlashAttnQKVPackedFunc(torch.autograd.Function):
         deterministic,
         return_softmax,
     ):
-        is_grad = torch.is_grad_enabled() and qkv.requires_grad
+        is_grad = qkv.requires_grad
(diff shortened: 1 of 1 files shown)


rocking5566:
LGTM  @tridao  please merge

tridao:
@XiaobingSuper can you just remove the `training_mode` argument and do `is_grad = qkv.requires_grad`?

XiaobingSuper:
> @XiaobingSuper can you just remove the `training_mode` argument and do `is_grad = qkv.requires_grad`?

Yes，it is updated.

rocking5566:
looks like this has been fix in this PR
https://github.com/Dao-AILab/flash-attention/pull/1430

tridao:
Right. Thanks for your help. I'll close this one for now

PR #1419, opened 2025-01-02
Add a macro for namespace

drisspg:
# Summary

Working changes to see what is needed to change flash's packaging in PyTorch


.eggs/README.txt
```diff
@@ -0,0 +1,6 @@
+This directory contains eggs that were downloaded by setuptools to build, test, and run plug-ins.
```
drisspg: Add to .gitignore

csrc/flash_attn/src/flash_fwd_split_hdim224_fp16_sm80.cu
```diff
@@ -1,7 +0,0 @@
-// Copyright (c) 2023, Tri Dao.
-// Splitting the different head dimensions to different files to speed up compilation.
```
drisspg: There wasn't a config for 224 in the gen configs, so deleted but need to double check if needed

drisspg: > There wasn't a config for 224 in the gen configs, so deleted but need to double check if needed

csrc/flash_attn/src/namespace_config.h
```diff
@@ -0,0 +1,67 @@
...
+ * - Include this header in all files that need access to the Flash namespace
+ *
+ */
+#pragma once
+
+#ifndef FLASH_NAMESPACE_CONFIG_H
+#define FLASH_NAMESPACE_CONFIG_H
+
+// Set default namespace to flash
+#ifndef FLASH_NAMESPACE
```
drisspg, line 56: This file might be overkill..

PR #1433, opened 2025-01-09
Expose `zero_tensors` arg in varlen functions

ksivaman:
There are certain workloads where we use these internal functions directly and it is useful for us to have this argument exposed as a kwarg. The default behavior remains the same.


tridao:
I'm not sure what the torch compile registration needs but can you change the "fake" functions as well to take in zero_tensors?

PR #1446, opened 2025-01-15
Support ROCM builds from source distribution, and improve error handling

mgorny:
Currently, the source distribution contains only sources needed for CUDA builds (it's missing `csrc/composable_kernel`). Fix that, and while at it clean up some logic in `setup.py`. More specifically:

- always update both submodules when inside the git repo, so that they are both included in sdist
- add missing `.py` files to sdist
- run `git submodule ...` commands only when inside the git repo; otherwise (e.g. when installing from sdist), just verify that the submodules were checked out, and issue a more explanatory error if they weren't
- use `subprocess.run()` throughout to run commands, and check their return status

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -3,6 +3,7 @@ recursive-include csrc *.h
 recursive-include csrc *.cuh
 recursive-include csrc *.cpp
 recursive-include csrc *.hpp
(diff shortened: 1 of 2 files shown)


mgorny:
Thank you!

PR #1455, opened 2025-01-21
Fix a wrong reference to seqlen_k variable in the fwd_splitkv kernel

muoshuosha:
Use params.seqlen_k is wrong with varlen cases，this bug had been fixed in [vllm-project/flash-attntion](https://github.com/vllm-project/flash-attention/pull/18)


muoshuosha:
params.seqlen_k also is wrong in page attn when use `cu_seqlen_k` store the real seqlen_k https://github.com/Dao-AILab/flash-attention/blob/6b1d059eda21c1bd421f3d352786fca2cab61954/csrc/flash_attn/flash_api.cpp#L1386-L1393

PR #1468, opened 2025-01-28
fa3: include bert_padding utilities

### LINKED ISSUE #1467: FA3 package is missing padding utilities
I'm using FA3 as so:

>uv pip install git+https://github.com/Dao-AILab/flash-attention#subdirectory=hopper --no-build-isolation

When I install FA3, the only files installed are:

```
❯ ls .venv/lib/python3.11/site-packages/flash_attn*
flash_attn-3.0.0b1.dist-info/                      
flash_attn_3_cuda.cpython-311-x86_64-linux-gnu.so  
flash_attn_interface.py
```

The `hopper/padding.py` is not present. 

It seems to get the full set of padding utils, I need `flash_attn.bert_padding` which is only available if I install FA2 as well. But installing FA2 and FA3 in the same pip env seems tricky given they have the same package name.

tmm1:
one way to fix #1467

### DIFF
--- a/hopper/flash_attn/bert_padding.py
+++ b/hopper/flash_attn/bert_padding.py
@@ -1,0 +1,218 @@
+# Adapted from https://github.com/mlcommons/training_results_v1.1/blob/main/NVIDIA/benchmarks/bert/implementations/pytorch/padding.py
+
+import torch
+import torch.nn.functional as F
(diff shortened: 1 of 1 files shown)


tmm1:
This didn't work out, because exposing `flash_attn` makes transformers thinks fa2 is available.

Issue #1470, opened 2025-02-03
Getting Error While Extracting


emirardagn:
Extracting flash_attn-3.0.0b1-py3.12-linux-x86_64.egg to /data/conda/envs/vllm/lib/python3.12/site-packages

 Adding flash-attn 3.0.0b1 to easy-install.pth file
 
Installed /data/conda/envs/vllm/lib/python3.12/site-packages/flash_attn-3.0.0b1-py3.12-linux-x86_64.egg

 Processing dependencies for flash-attn==3.0.0b1

 Searching for ninja==1.11.1.3

 Best match: ninja 1.11.1.3

 Processing ninja-1.11.1.3-py3.12-linux-x86_64.egg

 Adding ninja 1.11.1.3 to easy-install.pth file

 Traceback (most recent call last):

   File "/home/A06506/test/flash-attention/hopper/setup.py", line 600, in <module>

     setup(

   File "/data/conda/envs/vllm/lib/python3.12/site-packages/setuptools/__init__.py", line 117, in setup

     return distutils.core.setup(**attrs)

            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^

   File "/data/conda/envs/vllm/lib/python3.12/site-packages/setuptools/_distutils/core.py", line 183, in setup

     return run_commands(dist)

            ^^^^^^^^^^^^^^^^^^

   File "/data/conda/envs/vllm/lib/python3.12/site-packages/setuptools/_distutils/core.py", line 199, in run_commands

     dist.run_commands()

   File "/data/conda/envs/vllm/lib/python3.12/site-packages/setuptools/_distutils/dist.py", line 954, in run_commands

     self.run_command(cmd)

   File "/data/conda/envs/vllm/lib/python3.12/site-packages/setuptools/dist.py", line 950, in run_command

     super().run_command(command)

   File "/data/conda/envs/vllm/lib/python3.12/site-packages/setuptools/_distutils/dist.py", line 973, in run_command

     cmd_obj.run()

   File "/data/conda/envs/vllm/lib/python3.12/site-packages/setuptools/command/install.py", line 97, in run

     self.do_egg_install()

   File "/data/conda/envs/vllm/lib/python3.12/site-packages/setuptools/command/install.py", line 158, in do_egg_install

     cmd.run(show_deprecation=False)

   File "/data/conda/envs/vllm/lib/python3.12/site-packages/setuptools/command/easy_install.py", line 439, in run

[...]

PR #1484, opened 2025-02-09
Remove ninja runtime dependency

kevmo314:
Fixes https://github.com/Dao-AILab/flash-attention/issues/1470

This section is for runtime dependencies and `ninja` is more necessary at build time, not at run time. I'm not sure why adding it is causing the error in the linked issue but removing it resolves the error.

### DIFF
--- a/hopper/setup.py
+++ b/hopper/setup.py
@@ -631,6 +631,5 @@ setup(
         "torch",
         "einops",
         "packaging",
-        "ninja",
     ],
 )


l1cacheDell:
I've met the same problem, and this works out fine for me. Thank you!

RuiWang1998:
Works for me as well!

hopper/setup.py
```diff
@@ -631,6 +631,5 @@ def run(self):
         "torch",
         "einops",
         "packaging",
-        "ninja",
     ],
```
weiji14, line 637: Still need to put `ninja` under `setup_requires` (i.e. as a build time dependency) I think. See previous PR at #937.

```suggestion
    ],
    setup_requires=[
        "ninja",
```

kevmo314: Thanks, I've moved packaging as well to align with your PR

PR #1488, opened 2025-02-11
[docs][FA3] Update build guide of FlashAttention-3

Ethkuil:
I encounter some problems when building and installing FlashAttention-3. So I'm here to make the docs better and may help others.

This PR involves 2 things:

1. CUDA version.
2. deprecated `setup.py install`. https://github.com/pypa/pip/issues/12330

Hope these help.


zdxff:
how can I use pip to install FA3? Could you tell me the instruction?

PR #1497, opened 2025-02-16
[`FA3`] Make seqused optional in varlen interface

### LINKED ISSUE #1495: [`FA3`] Forced usage of `seqused_(q/k)` in varlen
Just a quick question on the hopper interface, it seems that the recent refactor and subsequent updates changed the varlen interface, i.e. seqused is now an arg:
https://github.com/Dao-AILab/flash-attention/blob/fa445ff6c215026438cca496a97242b8269aa428/hopper/flash_attn_interface.py#L566-L567

Whereas before the refactor it was an optional kwarg:
https://github.com/Dao-AILab/flash-attention/blob/284e2c6e5beff017996d72de6e028b2dc605acf8/hopper/flash_attn_interface.py#L441-L442

I'm honestly not sure if that was intentional and if it was, I'd love to know about the reasons :D

vasqu:
Only changed the top-level interface and adjusted the calls made in other files (benchmark, test). cc @tridao 

Closes #1495


tridao:
Thanks!

PR #1500, opened 2025-02-21
fix: promote index to type longlong to avoid numerical overflow

xin-w8023:
When using Flash Attention, if the batch size is relatively large, say 10k or 15k, an "illegal memory access" error will occur. Eventually, it turns out that the issue stems from `row_offset_dq_accum`  numerical  overflow during the backward of Flash Attention.


xin-w8023:
cc @tridao

xin-w8023:
@tridao Hi, could you help to review this ? We want to use the open-sourced version, but this issue limits us.

tridao:
Thanks!

PR #1504, opened 2025-02-22
Enable MLA flag in FA3 (rope=64, latent=512)

tzadouri:
Adding an MLA feature allows ropeDim=64 and latentDim=512, and updates conditional checks in flash_api.cpp to handle MLA-enabled paths.

### DIFF
--- a/hopper/flash_api.cpp
+++ b/hopper/flash_api.cpp
@@ -271,7 +271,14 @@ void run_mha_fwd(Flash_fwd_params &params, cudaStream_t stream) {
                         if (!params.is_e4m3) {
                             if (params.is_bf16) {
                                 #ifndef FLASHATTENTION_DISABLE_HDIM64
-                                if (params.d <= 64) { return run_mha_fwd_<Arch, cutlass::bfloat16_t, 64, 64, Split, PagedKV, Has_softcap, PackGQA>(params, stream); }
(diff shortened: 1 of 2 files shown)


YLGH:
Hi! For MLA I believe we need QK hidden dim = 576 and v dim 512. Please correct me if I'm wrong but this like it will support 64 QK head dim and 512z is there a way we can reduce across QK head dim afterward? ( and do like 256 256 64?) Thanks!

tridao:
Here 64 refers to the rope dimension and 512 refers to the non-rope dimension. This corresponds to the MLA configuration in DeepSeek V3 and R1.
You can pass in:
```
flash_attn_with_kvcache(
    q_rope, # hdim 64
    k_rope, # hdim 64
    latent,  # hdim 512
    qv=q_nope  # hdim 512
)
```

YLGH:
> Here 64 refers to the rope dimension and 512 refers to the non-rope dimension. This corresponds to the MLA configuration in DeepSeek V3 and R1. You can pass in:
> 
> ```
> flash_attn_with_kvcache(
>     q_rope, # hdim 64
>     k_rope, # hdim 64
>     latent,  # hdim 512
>     qv=q_nope  # hdim 512
> )
> ```

Ah, I see thanks! Didn't realize qv was a parameter!

PR #1511, opened 2025-02-26
Fix CUDA 12.1 build

LucasWilkinson:
Building using CUDA 12.1 was resulting in:
```
.../hopper/mainloop_fwd_sm90_tma_gmma_ws.hpp(1068): error: no operator "=" matches these operands
  operand types are: cutlass::bfloat16_t = float
    sScale(get<0>(taccOcO_row(mi)), stage) = scales(mi);
```
which appears to be from slight differences in template type deduction


tridao:
I see, this fix seems fine to me

PR #1543, opened 2025-03-19
Fix invalid cross-device link

yuanwei-uipath:


sam-twlabs:
Is there a reason you closed this? I think this affects quite a bit of people and don't see why this shouldn't be the default.

PR #1552, opened 2025-03-25
[Easy] fix num_head reference in tile_scheduler.hpp

yeqcharlotte:
It throws build error. num_head -> args.num_head to fix that.


yeqcharlotte:
cc: @ipiszy

PR #1561, opened 2025-03-28
don't use set_varlen_params in varlen_fwd

jfactory07:
n set_varlen_params function, calculating max_seqlens_q and max_seqlens_k requires copying the values from the GPU to compute the maximum. However, since max_seqlen_q and max_seqlen_k are inputs to varlen_fwd, they do not need to be recalculated and can be copied directly.

### DIFF
--- a/flash_attn/flash_attn_triton_amd/interface_fa.py
+++ b/flash_attn/flash_attn_triton_amd/interface_fa.py
@@ -280,8 +280,12 @@ def varlen_fwd(
     metadata = MetaData(sm_scale=softmax_scale)
     if return_softmax:
         metadata.return_scores = True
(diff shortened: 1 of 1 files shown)


tridao:
@micmelesse could you review this?

micmelesse:
@jfactory07 this should be fine but do you mind first adding this to our fork at main_perf [here](https://github.com/ROCm/flash-attention/tree/main_perf). We are working on a second pr which will be adding a lot of features in the next few weeks.

jfactory07:
> @jfactory07 this should be fine but do you mind first adding this to our fork at main_perf [here](https://github.com/ROCm/flash-attention/tree/main_perf). We are working on a second pr which will be adding a lot of features in the next few weeks. We also have ci there.

okay. Please help to add it. I will discard this PR.

micmelesse:
This is fixed in https://github.com/Dao-AILab/flash-attention/pull/1610

Issue #1487, opened 2025-02-10
Support hdimQK=192, hdimV=128 backward


jacob-cursor:
Very happy to see https://github.com/Dao-AILab/flash-attention/commit/2a204125ae71d2010bd3c9634d72a81c63967f3b which added support for the Deepseek attention configuration (qk_dim=192, v_dim=128) to FA3.

Currently, backward is not supported:
```
[rank1]:     dq, dk, dv, softmax_d, *rest = flash_attn_3_cuda.bwd(
[rank1]:                                    ^^^^^^^^^^^^^^^^^^^^^^
[rank1]: RuntimeError: out must have shape (batch_size, seqlen_q, num_heads, head_size)
```
The backward pass for this op would be really useful!

tridao:
I'm personally not working on that soon, but I can review PRs if there's contribution

PR #1604, opened 2025-04-21
Support hdimQK != hdimV backward

shcho1118:
Recently, there are cases where ~~hdimQK >= hdimV~~ hdimQK != hdimV, such as DeepSeek (#1487).
So we worked on a simple trick (rounding hdimQK and hdimV to the same value) to support that configuration in the backward kernel as well.
Of course, it would be best to tile K and V with different rounding values (which we're working on internally), but this implementation allows us to skip explicit padding & unpadding hdimV, resulting in a performance improvement of about 15%.

This feature was developed for internal use at @kakao.

### DIFF
--- a/hopper/epilogue_bwd.hpp
+++ b/hopper/epilogue_bwd.hpp
@@ -107,6 +107,7 @@ struct CollectiveEpilogueBwd {
         ShapedKV const shape_dK;
         StridedKV const stride_dK;
         Element* ptr_dV;
+        ShapedKV const shape_dV;
         StridedKV const stride_dV;
         int const num_heads_q;
         int* dk_semaphore;
@@ -121,6 +122,7 @@ struct CollectiveEpilogueBwd {
         ShapedKV const shape_dK;
         StridedKV const stride_dK;
         Element* ptr_dV;
+        ShapedKV const shape_dV;
         StridedKV const stride_dV;
         TMA_dKV tma_store_dK, tma_store_dV;
         int const* cu_seqlens = nullptr;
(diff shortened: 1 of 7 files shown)


tridao:
Thank you! Could you rebase on the latest in `main` and then I'll merge?

shcho1118:
We've rebased to the latest as you requested. It's probably ready to merge.

shcho1118:
@tridao 
We've done the rebase as you requested, and I think we're ready to merge.

tridao:
The hdim dispatching should be based on `max(params.d, params.d_v)` instead of based on `params.d`?
E.g. if d = 128, d_v = 192, then it should dispatch to `kHeadDim=192` instead of `kHeadDim=128`?

shcho1118:
@tridao
I took your advice and extended the support a bit more, now if `max(params.d, params.dv) <= 256` it should work fine.

ehuaa:
Hi Sanghun@shcho1118 , I noticed that you have implemented hdimQK != hdimV for backward for both sm80 and sm90 arch,

ehuaa:
> @ehuaa After I finished implementing sm80 backward, I discovered that if-statement, so sm80 backward kernel is not fully validated in this PR. If you want to enable the same functionality in sm80 forward kernel, the fastest way to do it is to touch rounding hdim like this PR.
Thanks for your quick reply, i'll try it later.

Graham1025:
> Recently, there are cases where ~hdimQK >= hdimV~ hdimQK != hdimV, such as DeepSeek (#1487). So we worked on a simple trick (rounding hdimQK and hdimV to the same value) to support that configuration in the backward kernel as well. Of course, it would be best to tile K and V with different rounding values (which we're working on internally), but this implementation allows us to skip explicit padding & unpadding hdimV, resulting in a performance improvement of about 15%.
> 
> This feature was developed for internal use at @kakao.
Thanks for your great work! So does it use TMA oob feature for (rounding hdimQK and hdimV to the same value)?

shcho1118:
@Graham1025
Yes, for sm90, it was already using that feature to do rounding, so I just modified it slightly.

Issue #1620, opened 2025-04-26
RuntimeError: CUDA error: invalid configuration argument. When I use RMSNorm with zero dimension input


Luciennnnnnn:
Hi, I'm trying to use `flash_attn.ops.triton.layer_norm import RMSNorm` for my transformer, since my data have dynamic length, it even become 0 in some batch. I found RMSNorm of FlashAttention will introduce odd CUDA error, while it is fine for pytorch's RMSNorm.

Example use case:
```python
import torch

from flash_attn.ops.triton.layer_norm import RMSNorm as FusedRMSNorm

norm = FusedRMSNorm(2520, eps=1e-5).to(device='cuda', dtype=torch.bfloat16)
x = torch.randn([3, 0, 2520], dtype=torch.bfloat16, device='cuda')
out = norm(x)
```

Above code do not produce error immediately, but results in CUDA error in following execution.

tridao:
Do you want to add some check and return early in RMSNorm in a PR?

Luciennnnnnn:
@tridao Sure! Could you please provide some guidance on how to add those checks? I'm not very familiar with CUDA ops.

tridao:
It's just in this file (python / triton only, no cuda):
https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/ops/triton/layer_norm.py

Luciennnnnnn:
Ok, I'll take a look. However, I have a concern: if we directly skip the computation, could it affect gradient synchronization across multiple GPUs? If one GPU skips the Norm computation, it might not produce gradients for the Norm parameters on that GPU.

tridao:
I see, you'd want the gradient to be zero in that case.

Luciennnnnnn:
so, can we directly return `x + 0 * weights`? this will create a zero gradient for parameters.

tridao:
No you'd want to modify both the forward and backward pass.

Luciennnnnnn:
That is to say, should I add checks in both forward and backward pass? 1. Skip computation in forward. 2. Return zero gradient in backward

tridao:
Right

Luciennnnnnn:
Thanks, I'll try it

Luciennnnnnn:
Hi, I have added checks for LayerNorm. Also, I'm wondering which implementation is faster, CUDA or Triton?

tridao:
About the same speed. Triton is easier to use (no need to compile).

[...]

PR #1621, opened 2025-04-27
add checks for zero elements input of triton LayerNorm impl

Luciennnnnnn:
#1620 I have made some checks for zero elements input, please see whether or not it is right. Or how can I add proper tests？

### DIFF
--- a/flash_attn/ops/triton/layer_norm.py
+++ b/flash_attn/ops/triton/layer_norm.py
@@ -754,6 +754,31 @@ class LayerNormFn(torch.autograd.Function):
         residual_out=None
     ):
         x_shape_og = x.shape
+        # Check for zero sequence length
+        if x.numel() == 0:
+            ctx.zero_seq_length = True
+            # Return empty tensors of correct shape
+            y = torch.empty_like(x)
+            y1 = torch.empty_like(x) if x1 is not None else None
+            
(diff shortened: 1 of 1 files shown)


Luciennnnnnn:
Hi, My code can pass following simple test, it see it results in consistent output as `torch.nn.RMSNorm`. But I'm not sure whether or not my code is correct in determining `dtype` of forward/backward output 

```python
import torch

from flash_attn.ops.triton.layer_norm import RMSNorm as FusedRMSNorm

norm = FusedRMSNorm(2520, eps=1e-5).to(device='cuda', dtype=torch.bfloat16)
ref_norm = torch.nn.RMSNorm(2520, eps=1e-5).to(device='cuda', dtype=torch.bfloat16)

x = torch.randn([3, 0, 2520], dtype=torch.bfloat16, device='cuda', requires_grad=True)
ref_x = torch.randn([3, 0, 2520], dtype=torch.bfloat16, device='cuda', requires_grad=True)

out = norm(x)
ref_out = ref_norm(ref_x)

out.sum().backward()
ref_out.sum().backward()

print(out)
print(ref_out)

print(x.grad)
print(ref_x.grad)

print(norm.weight.grad)
print(ref_norm.weight.grad)
```

PR #1662, opened 2025-05-12
[fa3] Use Python stable ABI

danthe3rd:
This should enable us to build a single wheel for all python versions
(still requires a different build per OS/PyTorch version)

### DIFF
--- a/hopper/flash_api.cpp
+++ b/hopper/flash_api.cpp
@@ -16,7 +16,7 @@
 #include "heuristics.h"
 #include "cuda_check.h"
 
-#define STRINGIZE(x) #x
+
 extern "C" {
 /* Creates a dummy empty _C module that can be imported from Python.
     The import from Python will load the .so consisting of this file
--- a/hopper/flash_api.cpp
+++ b/hopper/flash_api.cpp
@@ -4,7 +4,6 @@
 
 #include <Python.h>
 #include <torch/nn/functional.h>
-#include <torch/version.h>  // For TORCH_VERSION* macros
 #include <ATen/cuda/CUDAContext.h>
 #include <c10/cuda/CUDAGuard.h>
 
--- a/hopper/flash_api.cpp
+++ b/hopper/flash_api.cpp
@@ -2,8 +2,7 @@
  * Copyright (c) 2024, Jay Shah, Ganesh Bikshandi, Ying Zhang, Vijay Thakkar, Pradeep Ramani, Tri Dao.
  ******************************************************************************/
 
-// Include these 2 headers instead of torch/extension.h since we don't need all of the torch headers.
-#include <torch/python.h>
+#include <Python.h>
 #include <torch/nn/functional.h>
 #include <torch/version.h>  // For TORCH_VERSION* macros
 #include <ATen/cuda/CUDAContext.h>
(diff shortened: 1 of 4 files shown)


hopper/flash_api.cpp
```diff
@@ -17,44 +16,25 @@
 #include "heuristics.h"
 #include "cuda_check.h"
 
-// Copied from https://github.com/pytorch/pytorch/commit/7931eee5c5ebcdf468bff4d308510b03355cd909
-// This is so that we can pass in torch.dtype as a parameter to the function.
-#if TORCH_VERSION_MAJOR < 2 || (TORCH_VERSION_MAJOR == 2 && TORCH_VERSION_MINOR < 4)
```
danthe3rd, line 22: Note: I had to remove this because we're no longer using PyBind. This will probably break with PyTorch < 2.4, is this fine, or do we want a backup solution for older PyTorch versions?

hopper/setup.py
```diff
@@ -539,10 +539,10 @@ def nvcc_threads_args():
...
         CUDAExtension(
-            name="flash_attn_3_cuda",
+            name=f"{PACKAGE_NAME}._C",
             sources=sources,
             extra_compile_args={
-                "cxx": ["-O3", "-std=c++17"] + feature_args,
+                "cxx": ["-O3", "-std=c++17", "-DPy_LIMITED_API=0x03090000"] + feature_args,
                 "nvcc": nvcc_threads_args() + nvcc_flags + cc_flag + feature_args,
             },
             include_dirs=include_dirs,
```
janeyx99, line 548:
```suggestion
            include_dirs=include_dirs,
            py_limited_api=True,
```

janeyx99 (commented):
Looks good from my perspective! Thanks Daniel for taking this on! Did you locally verify this change?

danthe3rd:
I created a wheel with cp39, and installed in on a cp39 and cp312 env. The import seemed to work:
```
python -c "import torch; import flash_attn_3._C"
```

janeyx99:
> I created a wheel with cp39, and installed in on a cp39 and cp312 env. The import seemed to work:
> 
> ```
> python -c "import torch; import flash_attn_3._C"
> ```

Would be best to also run the existing test cases for both envs!

hopper/flash_api.cpp
tridao: that's fine, we won't support pytorhc < 2.4

hopper/flash_api.cpp
```diff
@@ -2,8 +2,7 @@
  * Copyright (c) 2024, Jay Shah, Ganesh Bikshandi, Ying Zhang, Vijay Thakkar, Pradeep Ramani, Tri Dao.
  ******************************************************************************/
 
-// Include these 2 headers instead of torch/extension.h since we don't need all of the torch headers.
-#include <torch/python.h>
+#include <Python.h>
 #include <torch/nn/functional.h>
 #include <torch/version.h>  // For TORCH_VERSION* macros
```
janeyx99:
```suggestion
```
nit: can be deleted since line 22 is gone, but is no big deal at all

ryderhwang:
we are using pytorch 2.7.0 + cuda 12.8 hardware with cuda 12.6 lib + python 3.10 and we are seeing build issues after this commit, have you verified that python3.10 works?

winglian:
@danthe3rd Could this be backported to FA2 in the meantime too since it seems there are issues with the current wheels https://github.com/Dao-AILab/flash-attention/issues/1717 ?

danthe3rd:
> we are using pytorch 2.7.0 + cuda 12.8 hardware with cuda 12.6 lib + python 3.10 and we are seeing build issues after this commit, have you verified that python3.10 works?

Hi,
thanks for the report @ryderhwang 
Can you open a new issue with the corresponding build issue you get? Make sure to include your system configuration + the build log (and tag me please :) )

PR #1698, opened 2025-06-06
[fa3] Some fixes for windows build

danthe3rd:
Those changes are necessary to build on windows, as MSVC does not like `#ifdefs` inside preprocessor macro arguments.

### DIFF
--- a/hopper/flash_api.cpp
+++ b/hopper/flash_api.cpp
@@ -1143,7 +1143,7 @@ void run_mha_bwd(Flash_bwd_params &params, cudaStream_t stream) {
 }
 #else
 template <int Arch, bool Has_softcap>
-void run_mha_bwd(Flash_bwd_params &params, cudaStream_t stream) {
+void run_mha_bwd_constexpr(Flash_bwd_params &params, cudaStream_t stream) {
     if (!params.is_bf16) {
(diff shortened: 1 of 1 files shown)


hopper/flash_api.cpp
```diff
@@ -244,6 +244,108 @@ void set_params_dgrad(Flash_bwd_params &params,
...
+        if (params.is_bf16) {
+            #ifndef FLASHATTENTION_DISABLE_HDIM64
+            if (params.d <= 64) {
+                if constexpr (Arch == 90) {
+                    if (params.dv > 256) {
+                        return run_mha_fwd_<Arch, cutlass::bfloat16_t, 64, 512, Split, PagedKVNonTMA, Has_softcap, PackGQA>(params, stream);
+                    } else if (params.dv > 64) {
+                        return run_mha_fwd_<Arch, cutlass::bfloat16_t, 64, 256, Split, PagedKVNonTMA, Has_softcap, PackGQA>(params, stream);
+                    }
+                }
```
danthe3rd, line 259: I also had to use `if constexpr` to avoid linking errors, because some variants are not built for eg sm80.
Apparently gcc is smart enough to optimize out the calls, but MSVC is not

tridao:
Thank you Daniel!

Issue #1715, opened 2025-06-14
Windows Compile is broken... i fixed it


loscrossos:
The windows compile with MSVC is broken since somewhere after the last official release (2.7.4post1).

## Problem:
When compiling on Windows with MSVC, the following error occurs:
```
error: namespace "cutlass::platform" has no member "is_unsigned_v"
static_assert(cutlass::platform::is_unsigned_v<Storage>, "...");
```
This happens because is_unsigned_v is a C++17 feature, that Cutlass is encapsulating in its platform namespace but only exposes when the C++ flag has been set. Currently the Flashatt setup doesn’t pass the flags on Windows in a way that the MSVC understands it.

## Root Cause
The compilation flags are not correctly set for MSVC compiler (and only gcc/clang style flags are passed), so that the namespace is not exposing the feature over the Cutlass include.

## Solution:
Cutlass needs to receive the c++17 flag AND `/Zc:__cplusplus` flag to enable the presentation of the internal types.

## fix:

i created a PR that refactors the flag settings and adds the correct flags. All it does is set the MSVC-style g++17 flag and add the __cplusplus flag to the compiler and to nvcc (to pass to the host compiler) under windows so Cutlass can see it. The linux code path is untouched. The flags are only set if the OS is windows and the `DISTUTILS_USE_SDK` environ var is set, which is meant to be used for MSVC, this way mingw or other compilers are not affected  on windows.

drew1two:
I can confirm the issue, and I can confirm that this change fixed the issue for me  👍 Thank you @loscrossos

Wallawalla47:
I spent so many hours troubleshooting and adding the few lines from this PR made everything finally work!  Thank you!

david-littlefield:
# Flash Attention Installation Guide for Windows 10

This is how I successfully installed Flash Attention on Windows 10. I have an NVIDIA GeForce RTX 3090. I'm using Python 3.10.6. It took about two hours to compile.

Thanks to **loscrossos** for the PR fix: https://github.com/Dao-AILab/flash-attention/issues/1715

## Prerequisites

1. Install **CUDA Toolkit 12.8**
2. Install **cuDNN 9.1.0**

## Option 1: Pre-compiled Wheel
**Quick install for Python 3.10.6, Windows AMD64, CUDA 12.8:**

[...]

PR #1716, opened 2025-06-14
Windows compile fix for MSVC

loscrossos:
this PR fixes the windows compile for MSVC by adding MSVC compatible flags for C++17, so that Cutlass exposes the correct types. For details see #1715


tridao:
thank you!

Issue #1718, opened 2025-06-16
Flash-Attention 3 no longer builds with CUTLASS 4.0


danthe3rd:
Hi,
Since updating to CUTLASS 4.0, the build seems broken:
```
$ /PATH/third_party/nvidia/backend/bin/nvcc \
-I/PATH/hopper -I/PATH/csrc/cutlass/include -I/checkpoint/YYYY/ZZZZ/XXXX/envs/YYYY_250508/targets/x86_64-linux/include/ -I/checkpoint/YYYY/ZZZZ/XXXX/envs/YYYY_250508/lib/python3.11/site-packages/torch/include -I/checkpoint/YYYY/ZZZZ/XXXX/envs/YYYY_250508/lib/python3.11/site-packages/torch/include/torch/csrc/api/include -I/checkpoint/YYYY/ZZZZ/XXXX/envs/YYYY_250508/include -I/checkpoint/YYYY/ZZZZ/XXXX/envs/YYYY_250508/include/python3.11 -c -c /PATH/hopper/instantiations/flash_fwd_hdim128_bf16_paged_split_softcapall_sm80.cu -o /tmp/tmp8od_gs1q.build-temp/instantiations/flash_fwd_hdim128_bf16_paged_split_softcapall_sm80.o -D__CUDA_NO_HALF_OPERATORS__ -D__CUDA_NO_HALF_CONVERSIONS__ -D__CUDA_NO_BFLOAT16_CONVERSIONS__ -D__CUDA_NO_HALF2_OPERATORS__ --expt-relaxed-constexpr --compiler-options ''"'"'-fPIC'"'"'' --threads 2 -O3 -std=c++17 --ftemplate-backtrace-limit=0 --use_fast_math --resource-usage -lineinfo -DCUTE_SM90_EXTENDED_MMA_SHAPES_ENABLED -DCUTLASS_ENABLE_GDC_FOR_SM90 -DCUTLASS_DEBUG_TRACE_LEVEL=0 -DNDEBUG -gencode arch=compute_80,code=sm_80 -DTORCH_API_INCLUDE_EXTENSION_H -DPy_LIMITED_API=0x03090000 -DTORCH_EXTENSION_NAME=_C -ccbin /checkpoint/YYYY/ZZZZ/XXXX/envs/YYYY_250508/bin/x86_64-conda-linux-gnu-cc
nvcc warning : incompatible redefinition for option 'compiler-bindir', the last value of this option was used
[...]

PR #1719, opened 2025-06-16
fix: add tile shape to copy op template args

rafacelente:
With the introduction of CUTLASS 4.0, `sm90_get_smem_store_op_for_accumulator` [added](https://github.com/NVIDIA/cutlass/blob/dc4817921edda44a549197ff3a9dcf5df0636e7b/include/cutlass/epilogue/collective/builders/sm90_common.inl#L40-L42) `EpilogueTile_MN` as a new template arg, which is missing from `CopyOpR2S` [here](https://github.com/Dao-AILab/flash-attention/blob/main/hopper/epilogue_fwd.hpp#L95).  This is mentioned on issue #1718 . 

Here's the error thrown during instantiation:

```
error: no instance of function template "cutlass::epilogue::collective::detail::sm90_get_smem_store_op_for_accumulator" matches the argument list
          decltype(cutlass::epilogue::collective::detail::sm90_get_smem_store_op_for_accumulator<StrideO, Element>()),
```

This fix adds the correct epilogue tile shape to the function so it correctly.

### DIFF
--- a/hopper/epilogue_fwd.hpp
+++ b/hopper/epilogue_fwd.hpp
@@ -89,10 +89,11 @@ struct CollectiveEpilogueFwd {
     using ShapeLSEPacked = std::conditional_t<!PackGQA, cute::Shape<int32_t, int32_t, int32_t, int32_t>, cute::Shape<cute::Shape<int32_t, int32_t>, int32_t, int32_t, int32_t>>;
     using StrideLSEPacked = std::conditional_t<!PackGQA, StrideLSE, cute::Stride<cute::Stride<int64_t, _1>, int64_t, int64_t, int64_t>>;
 
+    using EpilogueTile_MN = decltype(select<0, 1>(TileShape_MNK_PV{})());
     using CopyOpR2S = std::conditional_t<
         ArchTag::kMinComputeCapability >= 90,
         // cute::SM90_U32x4_STSM_N if Element size is 2 bytes (fp16, bf16)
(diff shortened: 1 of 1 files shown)


tridao:
Thank you!

hopper/epilogue_fwd.hpp
```diff
@@ -89,10 +89,11 @@ struct CollectiveEpilogueFwd {
     using ShapeLSEPacked = std::conditional_t<!PackGQA, cute::Shape<int32_t, int32_t, int32_t, int32_t>, cute::Shape<cute::Shape<int32_t, int32_t>, int32_t, int32_t, int32_t>>;
     using StrideLSEPacked = std::conditional_t<!PackGQA, StrideLSE, cute::Stride<cute::Stride<int64_t, _1>, int64_t, int64_t, int64_t>>;
 
+    using EpilogueTile_MN = decltype(select<0, 1>(TileShape_MNK_PV{})());
```
goiri, line 92: When compiling in A100, this now triggers:
```
epilogue_fwd.hpp(92): error: call of an object of a class type without appropriate operator() or conversion functions
```

danthe3rd: Can confirm

zhujian19891203: Same error in H100

Sireer: It appears that replacing the original line with `using EpilogueTile_MN = decltype(select<0, 1>(TileShape_MNK_PV{}));` resolves the compilation issues. Not sure if there is any other problem.

PR #1744, opened 2025-07-09
[BE] Better compress flash attention binaries

Skylion007:
Better compress wheels similar to https://github.com/pytorch/pytorch/pull/157791


tridao:
THANK YOU!

Skylion007:
@tridao FYI, because of Nvidia driver shenanigans, this doesn't actually work on older pre CUDA 12.4 Nvidia drivers because it's missing functionality to decompress. We may want to revert unfortunately to keep the wheels compatible. :(

rohit-gupta:
@Skylion007 this totally breaks cuda 11.8 compilation

PR #1748, opened 2025-07-10
Suppress warnings in windows compilation

XXXXRT666:
Disable some warnings generated in windows compilations

```text
cl : Command line warning D9025 : overriding '/D__CUDA_NO_HALF_OPERATORS__' with '/U__CUDA_NO_HALF_OPERATORS__'
cl : Command line warning D9025 : overriding '/D__CUDA_NO_HALF_CONVERSIONS__' with '/U__CUDA_NO_HALF_CONVERSIONS__'
cl : Command line warning D9025 : overriding '/D__CUDA_NO_HALF2_OPERATORS__' with '/U__CUDA_NO_HALF2_OPERATORS__'
cl : Command line warning D9025 : overriding '/D__CUDA_NO_BFLOAT16_CONVERSIONS__' with '/U__CUDA_NO_BFLOAT16_CONVERSIONS__'

D:\a\flash-attention\flash-attention\csrc\flash_attn\src\flash_fwd_launch_template.h(66): warning #177-D: variable "is_even_K" was declared but never referenced
      const bool is_even_K = params.d == Kernel_traits::kHeadDim;
      
D:\a\flash-attention\flash-attention\csrc\flash_attn\src\softmax.h(76): warning #221-D: floating-point value does not fit in required floating-point type
          const float max_scaled = max(mi) == -((float)(1e+300)) ? 0.f : max(mi) * (Scale_max ? scale : float(1.44269504088896340736));
          
D:/a/flash-attention/flash-attention/csrc/cutlass/include\cute/layout.hpp(1443): warning #550-D: variable "iseq" was set but never used
    auto iseq = detail::nullspace_seq<0>(flat_layout.stride(), seq<>{});
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -26,6 +26,9 @@ var/
 
 # IDE-related
 .idea/
+.vscode/
 
 # Dev
 venv
+.mypy_cache/
+.ruff_cache/


XXXXRT666:
These warnings may be triggered hundreds of times during a `pip install`, resulting in logs that can span over 20,000 lines. I have observed this behavior in other users’ [GitHub Actions workflows](https://github.com/kingbri1/flash-attention/actions/runs/14623561659/job/41042697925).

loscrossos:
i think the warning is fair: a macro that was defined (/D) is now being undefined (/U). 
Its ok to remove the warning but:
Not knowing the exact need for it, i am not sure if its a good idea to remove the macro definition "/D".. specially at it wasnt included in the setup.py.

Also you are removing "--expt-relaxed-constexpr", which didnt have a warning by your original post?

Also this PR adds LOTS of new entries to .gitignore that dont really have a relevance to the project...  specially to lots of AI coding tool created temp files while at the same time removing useful entries like "bin/" "venv" and ".eggs/"

did an AI create the PR?

XXXXRT666:
It's clearly that this repo does not need the defined macro preset in PyTorch, actually I found that the some nvcc flags are defined twice, it absolutely not a problem, but the D_ preset in PyTorch nvcc flags is annoying, gcc could ignore it, but msvc won't

And for gitignore, I just use the GitHub Template

XXXXRT666:
> Also you are removing "--expt-relaxed-constexpr", which didnt have a warning by your original post?

expt-relaxed-constexpr has been add in setup.py
It's duplicated, not a big problem

loscrossos:
thy thx for explaining.

i tried it out and can confirm, that the warnings only appear in Windows. Linux gcc do not display the warnings.

also on windows the warnings produce several megabyte(!) of useless textoutput, which makes it very hard to debug for actual useful warnings and errors; with the patch the output equals the output of linux, which is very clean and tidy.


also i just realized you can not simply disable the first warning and thats why you need to remove the /D flag to remove it.

the other two warnings are very annoying.


just disabling a warning has no effect on the outcome. i do not know if removing the /D has an effect on the code (it should not).
i didnt have the time to test the resulting wheel, can you confirm that test pass?


i think it would be better to remove the gitignore as it clutters the project.

otherwise, this PR would be a great improvement for the windows compile.

XXXXRT666:
> just disabling a warning has no effect on the outcome. i do not know if removing the /D has an effect on the code (it should not). I didn‘t have the time to test the resulting wheel, can you confirm that test pass?

I don’t have any Windows machines, so I use [GitHub Actions](https://github.com/XXXXRT666/flash-attention/actions/runs/16200802098) to build the package. You can find the prebuilt wheels [here](https://github.com/XXXXRT666/flash-attention/releases).

I didn’t include the backward ops and other optional components that can be skipped via compile flags, since building everything takes over 6 hours.

If the warning says it’s being overridden, then I think the /D flag is simply ignored. In that case, it should be safe to remove it. But it's better to have some test.

> i think it would be better to remove the gitignore as it clutters the project.

Done, just adding 3 lines

Issue #1749, opened 2025-07-11
device kernel image is invalid, flash_attn==2.8.1


GitMonkey0:
flash_attn/flash_attn_interface.py", line 165, in _flash_attn_varlen_forward
[rank0]:     out, softmax_lse, S_dmask, rng_state = flash_attn_gpu.varlen_fwd(
[rank0]: RuntimeError: CUDA error: device kernel image is invalid
[rank0]: CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
[rank0]: For debugging consider passing CUDA_LAUNCH_BLOCKING=1
[rank0]: Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

GitMonkey0:
2.8.0 works for me, maybe a problem of new release?

trmusson:
Just a note to confirm that I was getting the same error as GitMonkey0, and reverting to an older version solved it. (Specifically, I replaced 2.8.1 with 2.8.0.post2 via pip.)

GitMonkey0:
> Just a note to confirm that I was getting the same error as GitMonkey0, and reverting to an older version solved it. (Specifically, I replaced 2.8.1 with 2.8.0.post2 via pip.)

yep, I fixed this issue in version 2.8.0.post2 too. But now, most mainstream projects default to installing the latest version 2.8.1. That means every time I start the container, I have to go through the tedious process of reinstalling with  pip install flash_attn==2.8.0.post2 . I'm curious—does this problem happen across all GPU types and CUDA versions?

trmusson:
> I'm curious—does this problem happen across all GPU types and CUDA versions?

Here I'm using an RTX 3060 (according to nvidia-smi: Driver Version: 535.247.01, CUDA Version: 12.2) on Ubuntu 22.04.5 LTS.

JamesSand:
>  I'm curious—does this problem happen across all GPU types and CUDA versions?

Same problem when using A100 (Driver Version: 545.23.08    CUDA Version: 12.3 ), with Ubuntu 22.04.3 LTS

JamesSand:
Another question. Does 2.8.0.post2 support torch 2.8.0? I don't find any release for torch 2.8.0 on flash-attention 2.8.0.post2.

PR #1751, opened 2025-07-11
Revert "[BE] Better compress flash attention binaries (#1744)" due to compatibility issues

### LINKED ISSUE #1750: Broken backward compat. w/ CUDA 11.x by command-line flag `-compress-mode=size` in setup.py
### The Cause:

The build script for `flash-attn` is instructing the NVIDIA CUDA Compiler (`nvcc`) to use the command-line flag `-compress-mode=size`. This flag is a feature found only in newer versions of the CUDA Toolkit (specifically, CUDA 12.x).

If your CUDA Toolkit is an older version (CUDA 11.x or earlier) that does not recognize this option, the compilation of `flash-attn` will fail with the `nvcc fatal: Unknown option '-compress-mode=size'` error message.

### The Bug Location:
https://github.com/Dao-AILab/flash-attention/blob/adf27d1db38223288981c4dc3509efafbddd3422/setup.py#L211

### The Solution:

Perhaps add a switch that will omit the `-compress-mode=size` flag when older CUDA versions have been detected.

imoneoi:
This reverts commit 8ba246f6cc8813d41f9289e2781b7d8fa22a97cb.

Binary compression introduced several compatibility issues in 2.8.1.
Resolves #1750, #1749.


lucasjinreal:
Need this!

PR #1756, opened 2025-07-15
[AMD] Torch Compile Issues

micmelesse:
A few users encountered issues when using flash attention with torch compile on AMD. The code similar to the one below would run into asserts because our output didnot match what the op registerations where expecting.

```
q = torch.randn((BATCH, N_CTX_Q, HQ, D_HEAD))
k = torch.randn((BATCH, N_CTX_K, HK, D_HEAD))
v = torch.randn((BATCH, N_CTX_K, HK, D_HEAD))

flash_attn_func_compiled = torch.compile(flash_attn_func)
o = flash_attn_func_compiled(q, k, v, causal=True)
o.sum().backward()
```

The first issue was that op was expecting certain dimensions to be a factor of 128 which seems to be a detail of the Nvidia backend. Both the AMD CK and Triton don't have that limitation. We conditionaly change the op registration so that for AMD we expect the dimension to be match the input dimesinosns. Another issue was that the triton backend, when using varlen, was returning an lse value which was transposed shape. We have fixed that as well.

The Triton AMD Backend Tests pass.

image

### DIFF
--- a/flash_attn/flash_attn_interface.py
+++ b/flash_attn/flash_attn_interface.py
@@ -127,7 +127,10 @@ def _flash_attn_forward_fake(
     softmax_lse = torch.empty((batch_size, num_heads, seqlen_q), dtype=torch.float32, device=q.device, layout=q.layout)
     p = torch.empty((0,), dtype=q.dtype, device=q.device, layout=q.layout)
     if return_softmax:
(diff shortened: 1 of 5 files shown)


micmelesse:
@rocking5566 Can you take a look?

rocking5566:
LGTM!

micmelesse:
Hi @tridao, Have you had a chance to take a look at this pr and the interface change suggested? Is this something that is acceptable to you if not we will adapt to the interface.

rocking5566:
@tridao please merge

PR #1766, opened 2025-07-21
Enable the deterministic mode option in the backward kernel

GD06:
The deterministic flag is dropped in the BWD launch template. This PR adds back the deterministic flag, and turn off "Slice_dQKV_Mma" when the deterministic mode is ON for head_dim=256. 

NOTE: This could lead to register pressures, and slow down kernel performance for head_dim=256 when running with the deterministic mode.

### DIFF
--- a/hopper/mainloop_bwd_sm90_tma_gmma_ws.hpp
+++ b/hopper/mainloop_bwd_sm90_tma_gmma_ws.hpp
@@ -264,7 +264,9 @@ struct CollectiveMainloopBwdSm90 {
     static constexpr bool dQacc_use_TMA = kHeadDim < 256;
     // For hdim256, we want to slice the dQ MMA (64 x 256 on 2 WGs) into two (64 x 128 on 2 WGs) so that we can
     // do atomic add on one half before doing the other half of the MMA, to reduce register pressure.
-    static constexpr bool Slice_dQKV_Mma = kHeadDim == 256 && !dQacc_use_TMA && dQ_swapAB && AtomLayoutMdQ == 1 && NumMmaWarpGroups == 2;
+    static constexpr bool Slice_dQKV_Mma =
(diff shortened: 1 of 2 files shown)


GD06:
Hi @tridao @sgrigory , may I know if anyone can help with reviewing this PR?

We found that flash-attn v3 built from commits this year cannot have deterministic results even when the deterministic mode is ON. Then we deep dived into the issue a bit, and found that the flag was dropped in the BWD template. This PR aims to fix the issue. But please let me know if there are issues other than slowing down the cases with head_dim=256.

hopper/mainloop_bwd_sm90_tma_gmma_ws.hpp
```diff
@@ -607,7 +609,7 @@ struct CollectiveMainloopBwdSm90 {
             seqlen_info, n_block, bidb, params.window_size_left,
             params.window_size_right, 0 /*sink_token_length*/);
         // It's possible to have m_block_max <= m_block_min. Exit early
-        if constexpr (Is_causal || Is_local || Varlen) {
+        if constexpr ((Is_causal || Is_local || Varlen) && !Deterministic) {
```
GD06: Hi @tridao @sgrigory , this line of change can fix the issue of backward kernels hang with the deterministic mode. 

I think we are still having some issues on the head_dim=256 with the deterministic mode, but most of other cases I tested look fine.

sgrigory:
To check that the output is deterministic, maybe add this test to `test_flash_attn.py`?
https://github.com/Dao-AILab/flash-attention/blob/d2e3fc30f02426e0c2a06ad45791b19491c92760/tests/test_flash_attn_triton_amd.py#L2493

lisheng-spaghetti:
there is another PR https://github.com/Dao-AILab/flash-attention/pull/1678 which has the same fix but with more unit test coverage.

PR #1768, opened 2025-07-21
Add `torch.compile` support to flash attention 3

guilhermeleobas:
This is a follow-up PR of https://github.com/Dao-AILab/flash-attention/pull/1139 that adds torch.compile support to FA3.


guilhermeleobas:
I just realized that there was already a PR opened for this in https://github.com/Dao-AILab/flash-attention/pull/1333

Issue #1730, opened 2025-06-25
libtorch ABI stable flash-attention3 (FA3)


janeyx99:
## Context

PyTorch is working on a libtorch ABI stable subset of APIs to support custom ops to be ABI stable with libtorch, meaning you can build with one libtorch version and run with another if you only use this stable subset. I'd like to help make FA3 libtorch stable on top of its current CPython stable so that we can build 1 FA3 wheel to support multiple CPythons and multiple libtorchs!

## What is this

This issue is a parallel tracker tied to https://github.com/pytorch/pytorch/issues/154908 to track progress on both repos. I don't anticipate there to be many changes to FA3 needed, but am opening this issue to facilitate any discourse.

janeyx99:
This is done!

PR #1791, opened 2025-07-31
ABI stable fa3

mikaylagawarecki:
Related to https://github.com/pytorch/pytorch/issues/154908 and https://github.com/Dao-AILab/flash-attention/issues/1730

This is an ABI stable version of hopper/flash_api.cpp that should be ABI stable from the 08/30 nightly onwards.

```
pip install --pre torch==2.9.0.dev20250830+cu129 --index-url https://download.pytorch.org/whl/nightly/cu129
```

## Test results
[`pytest -q -s test_flash_attn.py` results on `main`](https://gist.github.com/mikaylagawarecki/150ad97fcb4affc84fc0d5631fc7c526)

[`pytest -q -s test_flash_attn.py` results on this branch](https://gist.github.com/mikaylagawarecki/2ee7264996c395cf226769f0e4a6e416)

## Results of python benchmark_attn.py 

`main`
```
### headdim = 128, causal = False, seqlen = 8192 ###
Fav2 fwd: 2.972ms, 369.9 TFLOPS
Fav2 bwd: 9.460ms, 290.6 TFLOPS
Fav3 fwd: 1.668ms, 659.1 TFLOPS
Fav3 bwd: 4.962ms, 553.9 TFLOPS

### headdim = 128, causal = True, seqlen = 8192 ###
Fav2 fwd: 1.935ms, 284.1 TFLOPS
Fav2 bwd: 4.788ms, 287.1 TFLOPS
Fav3 fwd: 0.873ms, 630.0 TFLOPS
Fav3 bwd: 2.585ms, 531.6 TFLOPS
```

this branch

```
### headdim = 128, causal = False, seqlen = 8192 ###
Fav2 fwd: 2.963ms, 371.1 TFLOPS
Fav2 bwd: 9.515ms, 288.9 TFLOPS
Fav3 fwd: 1.667ms, 659.6 TFLOPS
Fav3 bwd: 4.878ms, 563.5 TFLOPS

### headdim = 128, causal = True, seqlen = 8192 ###
Fav2 fwd: 1.954ms, 281.4 TFLOPS
Fav2 bwd: 4.755ms, 289.1 TFLOPS
Fav3 fwd: 0.872ms, 630.5 TFLOPS
Fav3 bwd: 2.586ms, 531.5 TFLOPS
```

## Note

The `CHECK_SHAPE` macro is rewritten rather weirdly using `.size()` rather than `.sizes()` as we do not have a stable version of IntArrayRef yet, but we can fix this in a later PR.

cc @janeyx99 @tridao

### DIFF
--- a/hopper/setup.py
+++ b/hopper/setup.py
@@ -532,7 +532,7 @@ if not SKIP_CUDA_BUILD:
     target_version = parse("2.9.0.dev20250830")
     stable_args = []
       
-    if torch_version > target_version:
+    if torch_version >= target_version:
         flash_api_source = "flash_api_stable.cpp"
         stable_args = ["-DTORCH_STABLE_ONLY"]  # Checks against including unstable Tensor APIs
     else:


janeyx99:
header only ScalarType is landing! https://github.com/pytorch/pytorch/pull/159416

hopper/flash_api.cpp
```diff
@@ -1034,22 +1074,22 @@ mha_fwd(at::Tensor q,   // (b, s_q, h, d) or (total_q, h, d) if there is cu_seql
...
+    Tensor out_accum, softmax_lse_accum;
     auto outaccum_type = at::ScalarType::Float;
     if (params.num_splits > 1) {
-        TORCH_CHECK(params.num_splits <= 256, "num_splits > 256 not supported");
+        STD_TORCH_CHECK(params.num_splits <= 256, "num_splits > 256 not supported");
         if (!is_varlen_q) {
-            out_accum = torch::empty({params.num_splits, batch_size, num_heads, seqlen_q, head_size_v}, opts.dtype(outaccum_type));
-            softmax_lse_accum = torch::empty({params.num_splits, batch_size, num_heads, seqlen_q}, opts.dtype(at::kFloat));
+            out_accum = torch::stable::new_empty(q, {params.num_splits, batch_size, num_heads, seqlen_q, head_size_v}, std::make_optional(outaccum_type));
+            softmax_lse_accum = torch::stable::new_empty(q, {params.num_splits, batch_size, num_heads, seqlen_q}, std::make_optional(c10::ScalarType::Float));
```
janeyx99:
```suggestion
            softmax_lse_accum = torch::stable::new_empty(q, {params.num_splits, batch_size, num_heads, seqlen_q}, std::make_optional(torch::headeronly::ScalarType::Float));
```
let's use torch::headeronly for these

janeyx99: This specific line can stay `outaccum_type`, right?

hopper/flash_api.cpp
```diff
@@ -1134,12 +1175,14 @@ mha_fwd(at::Tensor q,   // (b, s_q, h, d) or (total_q, h, d) if there is cu_seql
             run_mha_fwd_combine(params, stream, true /*enable_pdl*/);
         } else if (scheduler_needs_semaphore && params.skip_scheduler_metadata_computation) {
             // need to zero out the semaphore in this case
-            tile_count_semaphore.index({torch::indexing::Slice(0, 1)}).zero_();
+            auto slice = torch::stable::narrow(tile_count_semaphore, 0, 0, 1);
+            torch::stable::zero_(slice);
+            // torch::stable::zero_(tile_count_semaphore.index({torch::indexing::Slice(0, 1)}));
```
janeyx99: to be deleted?

janeyx99 (commented):
Left some minor nits --> I'm guessing the commented out stuff will be removed?

hopper/flash_api.cpp
```diff
@@ -1646,23 +1719,164 @@ mha_combine(at::Tensor out_partial,         // num_splits x batch_size x seqlen
...
+    auto attention_chunk = to<int64_t>(stack[19]);
+    auto has_softcap = to<bool>(stack[20]);
+    auto num_splits = to<int64_t>(stack[21]);
+    auto pack_gqa = to<std::optional<bool>>(stack[22]);
+    auto sm_margin = to<int64_t>(stack[23]);
+
+    auto scheduler_metadata = mha_fwd_get_scheduler_metadata(batch_size, max_seqlen_q, max_seqlen_k, num_heads, num_heads_k, headdim, headdim_v, qkv_dtype, seqused_k, cu_seqlens_q, cu_seqlens_k, cu_seqlens_k_new, seqused_q, leftpad_k, page_size, max_seqlen_k_new, is_causal, window_size_left, window_size_right, attention_chunk, has_softcap, num_splits, pack_gqa, sm_margin);
+
+    stack[0] = from(scheduler_metadata);
+}
```
danthe3rd: Do you think there is a way to automate this? It could be quite error prone.
Probably we can do something with C++ template magic?

hopper/flash_api.cpp
```diff
@@ -1140,29 +1203,30 @@ mha_fwd(at::Tensor q,   // (b, s_q, h, d) or (total_q, h, d) if there is cu_seql
     }
 
     #ifdef FLASHATTENTION_DISABLE_LOCAL
-    TORCH_CHECK(!params.is_local, "This flash attention build does not support local attention.");
+    STD_TORCH_CHECK(!params.is_local, "This flash attention build does not support local attention.");
```
danthe3rd: Could we avoid having to do this kind of changes? Like could we just replace `TORCH_CHECK`'s definition with `STD_TORCH_CHECK`'s definition - maybe when a given compile flag is set?

hopper/flash_api.cpp
```diff
@@ -15,11 +12,69 @@
...
+  int device_index;
+  cudaError_t err = cudaGetDevice(&device_index);
+  if (err != cudaSuccess) {
+    STD_TORCH_CHECK(false, "cudaGetDevice failed: " +
+                               std::string(cudaGetErrorString(err)));
+  }
+
+  std::call_once(device_flags[device_index], initDeviceProperty, device_index);
+  return &device_properties[device_index];
+}
```
danthe3rd: I can see multiple libraries could need this kind of functions. Would it be possible to have `getCurrentDeviceProperties`/`getCurrentCUDAStream`/... also part of the stable ABI? This should be stable anyway because CUDA ABI should be stable between minor versions

hopper/flash_api.cpp
janeyx99: We do want to eventually automate this and hide it from the user completely! We ultimately want to support this in the dispatcher which requires a bit of a lift as we support more IValues. Maybe it could be worth adding an intermediary template solution in the meantime.

hopper/flash_api.cpp
janeyx99: We haven't yet explored having a compile time flag control what gets exposed, and that will be a part of our consideration when we figure out how to set up ABI target versions and such :D

hopper/flash_api.cpp
janeyx99: We explicitly chose to support the accelerator agnostic variations of these APIs in stable for 2.9, as we want people to eventually migrate to generic accelerator APIs, but maybe we should revisit this if there's interest! cc @albanD

janeyx99:
Thanks @danthe3rd for your comments! They very much align with where we're headed, especially as @mikaylagawarecki and I start enabling other repos with many kernels soon. That said, the goal of this PR is to update FA3 to be libtorch-stable as soon as possible and still be readable, where the level of readability was based on the initial proposal PR: https://github.com/Dao-AILab/flash-attention/pull/1685/files.

To that end, I think this PR does mostly adhere to the initial plan, and so we'd like to verify with @tridao that we can make FA3 libtorch stable 2.9 onwards with the current PR. From there, we do plan on gradually improving the UX with utilities and additional APIs (e.g., for libraries that have more than one kernel :)). 

> A lot of these changes are essentially just renaming stuff. Would it be possible to have a namespace that would mirror as much as possible the existing class hierarchy?
Like I would like (as a library maintainer) to just set a compile flag -DTORCH_LIMITED_API=0x02090000, and it would hide all non-stable API.
Then maybe I can just set using namespace torch::stable;, and this namespace contains torch::stable::at::ScalarType, torch::stable::at::Tensor, torch::stable::at::cuda::DeviceGuard etc...

We are exploring ABI version build targets and ABI versioning (cc @mikaylagawarecki), and I think to some extent we want to offer ways for users to opt into only a subset of stable APIs! So yes! And then regarding the namespace change, I think it would be a bit confusing (even if convenient) to expose namespaces like at in torch::stable. We are aiming for a northstar where one day libtorch namespace separations would make sense, and people would use the right APIs from the getgo. We understand that this may make enabling more libraries feel daunting, so we intend to get in the weeds to help library authors in the migration process!

hopper/flash_api.cpp
```diff
@@ -1646,23 +1719,164 @@ mha_combine(at::Tensor out_partial,         // num_splits x batch_size x seqlen
...
-        out = out.index({"...", torch::indexing::Slice(torch::indexing::None, head_size_og)});
+        out = torch::stable::narrow(out, -1, 0, head_size_og);
         // if (out_.has_value()) { out_.value().copy_(out); }
     }
 
     return {out, softmax_lse};
 }
 
-TORCH_LIBRARY(flash_attn_3, m) {
+void boxed_mha_fwd(
```
tridao: I don't quite understand why we need the "boxed" version

tridao:
The API looks great to me!

hopper/flash_api.cpp
albanD: Yes it is an interesting balance between having generic enough APIs vs not.
tbh in this case, if the cuda API was good enough we wouldn't need this... 

But to get the current stream, you can use the generic API (and convert the object to cudaStream_t if ever needed).
To get properties it's trickier to get a generic API as each vendor provides different struct there (even though they contain mostly the same thing).

hopper/flash_api.cpp
mikaylagawarecki: This is a limitation of `STABLE_TORCH_LIBRARY` for now where it only accepts the boxed kernel. But echoing [Jane's comment](https://github.com/Dao-AILab/flash-attention/pull/1791#discussion_r2301938634) 

> We do want to eventually automate this and hide it from the user completely! We ultimately want to support this in the dispatcher which requires a bit of a lift as we support more IValues.

hopper/setup.py
```diff
@@ -526,8 +526,22 @@ def nvcc_threads_args():
     if DISABLE_BACKWARD:
         sources_bwd_sm90 = []
         sources_bwd_sm80 = []
+    
+    # Choose between flash_api.cpp and flash_api_stable.cpp based on torch version
+    torch_version_parsed = parse(torch.__version__)
```
janeyx99:
```suggestion
    torch_version = parse(torch.__version__)
```
for consistency with target_version

hopper/setup.py
```diff
@@ -526,8 +526,22 @@ def nvcc_threads_args():
...
+    target_version = parse("2.9.0.dev20250830")
+    stable_args = []
+      
+    if torch_version_parsed > target_version:
+        flash_api_source = "flash_api_stable.cpp"
+        stable_args = ["-DTORCH_STABLE_ONLY"]
```
janeyx99:
```suggestion
        stable_args = ["-DTORCH_STABLE_ONLY"]  # Checks against including unstable Tensor APIs
```

hopper/setup.py
```diff
@@ -526,8 +526,22 @@ def nvcc_threads_args():
...
+    else:
+        flash_api_source = "flash_api.cpp"
+
+    print(f"\n\nflash_api_source file = {flash_api_source}\n\n")
```
janeyx99:
```suggestion
```

hopper/setup.py
janeyx99: could just append to feature_args? no need to create new args?

mikaylagawarecki: I don't think we need to pass this to nvcc extra compile args, hence I separated it out

Issue #1782, opened 2025-07-29
cute.interface _flash_attn_fwd fucntion doesn't provide LSE tensor in the return value


beiw-nv:
For inference, it seems LSE tenor is returned as None [here](https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/cute/interface.py#L118). Is there anyway cute interface can return LSE as in other FA flash_attn_fwd function, for example: [here](https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/flash_attn_interface.py#L91])

tridao:
Right, for inference we don't return LSE since that's typically not needed. Not hard to change, it's just 1 line in pytorch

beiw-nv:
Can you please point me the line need to change in pytorch?

tridao:
That's the line you linked to:
https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/cute/interface.py#L118

Just change it to always construct the LSE tensor

beiw-nv:
I am wondering how multi-gpu run will affect the correctness of [_flash_attn_fwd function ](https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/cute/interface.py#L51) on B200. I am trying to integrate the latest FA3 cutlass implementation to LongContextAttention [here](https://github.com/feifeibear/long-context-attention). During 2-gpu testing, I found that gpu 0 always returns the correct out and lse. However, gpu 1 sometime returns mostly 0 values for out and lse depending on the synchronization between two GPUs. The run is launched with torchrun --nproc_per_node=2 so we have 1 proc per gpu.

tridao:
Maybe we need `with torch.cuda.device`, sth like this?
https://github.com/Dao-AILab/flash-attention/blob/413d07e9deef1e3c793c7de59d7146b43ae4d558/flash_attn/ops/triton/rotary.py#L158

beiw-nv:
thanks! let me try that.

beiw-nv:
Yes, adding 

> with torch.cuda.device(q.device.index):

 to[here](https://github.com/Dao-AILab/flash-attention/blob/413d07e9deef1e3c793c7de59d7146b43ae4d558/flash_attn/cute/interface.py#L154) fixes the correctness issue in multi-gpu runs completely. We may also need to add it for backward function as well.

NJX-njx:
[...]

PR #1793, opened 2025-08-01
fix race condition bug in cute _flash_attn_fwd in multiple gpu env

beiw-nv:
In multi-gpu runs, the cute implementation of _flash_attn_fwd returns incorrect values for gpu != 0. This can be fixed with torch.cuda.device context manager as sugguested [issue1782](https://github.com/Dao-AILab/flash-attention/issues/1782):


tridao:
I'm not sure it's "race condition", I suspect it just launches the kernel on cuda device 0 even when the data is on cuda device 1

beiw-nv:
I see. When do you expect we will have Blackwell support for _flash_attn_bwd?

tridao:
3-4 weeks

PR #1795, opened 2025-08-05
[BugFix] Fix flash_attn_with_kvcache with scalar cache_seqlen

stepinto:
When the parameter `cache_seqlen` is scalar, it should be expand to vector of shape (batch_size).  In the original code, whenever `block_table` is used, the shape of `k_cache` is (num_blocks, page_size, ...), and thus `cache_seqlen` is expanded to shape (num_blocks) instead of (batch_size), which is wrong.  This fix uses the shape of `q`, which is always `batch_size`.


tridao:
Thank you!

PR #1823, opened 2025-08-19
Add sorting and head swizzle to varlen scheduler

jayhshah:
This PR adds inter-batch sort and intra-batch LPT scheduling for varlen FA3 workloads.

On the backend, sorting is enabled when not local, and head swizzle is enabled when causal or local (same as for dynamic tile scheduler). The option to sort/swizzle or not is not accessible from the frontend to preserve the same interface for the user.

Changes to prepare kernel:
1. Sort by `num_n_blocks` or `num_n_blocks * kBlockN - num_m_blocks * kBlockM` for causal. Uses cub BlockMergeSort. Overhead is ~4 us for 8 warps (e.g. 8 x 31 = 248 batches), or ~10 us for 32 warps (e.g. 32 x 31 = 992 batches).
2. Call the sorted position of the batch index the "virtual batch index". We write out 4 arrays, each of size `num_batches`, that represent the following mappings: `vbidx -> bidx`, `vbidx -> num_m_blocks[bidx]`, `vbidx -> num_heads_in_l2[bidx]`, `vbidx -> num_splits[bidx]`.

Changes to tile scheduler:
1. Takes in metadata from prepare kernel to traverse batches in sorted order and do head swizzle for intra-batch LPT scheduling.
2. Changes localized to `tile_idx_to_work_tile` method, but `tile_idx` in WorkTileInfo now has semantic meaning of starting group tile index for next invocation of `tile_idx_to_work_tile` -- we were never using `tile_idx` except for this purpose in any case.

We also have the combine kernel read in `vbidx -> bidx` mapping since we now only know the `vbidx -> num_splits[bidx]` mapping.

### DIFF
--- a/hopper/flash_api.cpp
+++ b/hopper/flash_api.cpp
@@ -643,7 +643,7 @@ mha_fwd_get_scheduler_metadata(
         }
     }
 
-    if (params.num_splits_dynamic_ptr) {
+    if (use_prepare_varlen) {
         auto kBlockMN_kernel_args_sm90 = tile_size_fwd_sm90(params.d_rounded, params.dv_rounded, params.is_causal, params.is_local, params.is_e4m3 ? 1 : 2 /*element_size*/, false /*v_colmajor*/, params.page_table && !params.pagedkv_tma, params.softcap > 0.f);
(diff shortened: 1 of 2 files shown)


jayhshah:
We don't tune the split heuristic yet, which could be improved. We should also use a persistent tile scheduler for the combine kernel in the case where some batches have much larger `seqlen_q` than the rest so it avoids launching a huge number of zero-work CTAs, which negatively impacts perf for mixed prefill-decode workloads. We'll address those problems in a separate PR.

jayhshah:
Numbers looking good for causal perf boost:
[speeds_mqa8.txt](https://github.com/user-attachments/files/21928854/speeds_mqa8.txt)
[speeds_gqa8.txt](https://github.com/user-attachments/files/21928856/speeds_gqa8.txt)
[speeds_mha.txt](https://github.com/user-attachments/files/21928860/speeds_mha.txt)

PR #1827, opened 2025-08-19
ci: Allow build/deploy of arbitrary configurations

ko3n1g:


.github/workflows/build.yml
```diff
@@ -0,0 +1,48 @@
...
+      upload-to-release:
+        description: "Upload wheel to this release"
+        required: false
+        type: boolean
+        default: false
+      release-version:
+        description: "Upload wheel to this release"
+        required: false
+        type: string
+  push:
```
chtruong814: This was just for testing?

.github/workflows/build.yml
```diff
@@ -0,0 +1,48 @@
...
+
+jobs:
+  build-wheels:
+    uses: ./.github/workflows/_build.yml
+    with:
+      runs-on: ubuntu-22.04 # ${{ inputs.runs-on }}
+      python-version: 3.12 # ${{ inputs.python-version }}
+      cuda-version: 12.8.1 # ${{ inputs.cuda-version }}
+      torch-version: 2.8.0 # ${{ inputs.torch-version }}
+      cxx11-abi: TRUE # ${{ inputs.cxx11-abi }}
```
chtruong814: You'll replace these in a bit?

chtruong814 (commented):
I think we'll want to replace the placeholder for the push event and use the input values, right?

PR #1829, opened 2025-08-20
Add support Orin, Thor, Spark and GB300

johnnynunez:
With cuda 13, will enable this devices

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -179,6 +179,8 @@ if not SKIP_CUDA_BUILD and not IS_ROCM:
     if "80" in cuda_archs():
         cc_flag.append("-gencode")
         cc_flag.append("arch=compute_80,code=sm_80")
+        cc_flag.append("-gencode")
+        cc_flag.append("arch=compute_87,code=sm_87")
     if CUDA_HOME is not None:
         if bare_metal_version >= Version("11.8") and "90" in cuda_archs():
             cc_flag.append("-gencode")
@@ -189,6 +191,15 @@ if not SKIP_CUDA_BUILD and not IS_ROCM:
         if bare_metal_version >= Version("12.8") and "120" in cuda_archs():
             cc_flag.append("-gencode")
             cc_flag.append("arch=compute_120,code=sm_120")
+        if bare_metal_version >= Version("13.0") and "103" in cuda_archs():
+            cc_flag.append("-gencode")
+            cc_flag.append("arch=compute_103,code=sm_103")
+        if bare_metal_version >= Version("13.0") and "110" in cuda_archs():
+            cc_flag.append("-gencode")
+            cc_flag.append("arch=compute_110,code=sm_110")
+        if bare_metal_version >= Version("13.0") and "121" in cuda_archs():
+            cc_flag.append("-gencode")
+            cc_flag.append("arch=compute_121,code=sm_121")
 
     # HACK: The compiler flag -D_GLIBCXX_USE_CXX11_ABI is set to be the same as
     # torch._C._GLIBCXX_USE_CXX11_ABI


johnnynunez:
@tridao @ko3n1g also here haha

tridao:
I'm worried this would make compilation much slower. And we don't have any specialized code for those arch.
Can e.g. sm87 run the code that was compiled for sm86?
Or sm103 run the code that was compiled for sm100?

johnnynunez:
> I'm worried this would make compilation much slower. And we don't have any specialized code for those arch. Can e.g. sm87 run the code that was compiled for sm86? Or sm103 run the code that was compiled for sm100?

yes, it can be possible

Building for sm_100 natively supports all sm_10x architectures.
The sm_121 architecture is SASS binary-compatible with sm_120.
For TensorCore kernels (e.g., when using CUTLASS), architecture- and family-specific conditional compilation is required.

tridao:
So what's the minimum set of arch we need to compile for?
[FA compilation is already quite long (~6h on github runner)]

johnnynunez:
> So what's the minimum set of arch we need to compile for? [FA compilation is already quite long (~6h on github runner)]

80
90
100
120

johnnynunez:
@tridao @ko3n1g Why should we implement [CCACHE](https://ccache.dev/?utm_source=chatgpt.com) in CI pipeline?

Github could catch binaries that are compiled and not modified, and not rerun again all

Also cibuildwheel can generate for windows and linux and N python versions

https://github.com/pypa/cibuildwheel/discussions/779

this should speedup a lot compilations

tridao:
The current setup.py (before this PR) already compiles for sm80, 90, 100, 120 and so we should be good?

johnnynunez:
> The current setup.py (before this PR) already compiles for sm80, 90, 100, 120 and so we should be good?

yes, i think so. Maybe another nvidian can comment here

ko3n1g:
> @tridao @ko3n1g Why should we implement [CCACHE](https://ccache.dev/?utm_source=chatgpt.com) in CI pipeline?
> 
> Github could catch binaries that are compiled and not modified, and not rerun again all
> 
> Also cibuildwheel can generate for windows and linux and N python versions
> 
> [pypa/cibuildwheel#779](https://github.com/pypa/cibuildwheel/discussions/779)
> 
> this should speedup a lot compilations

We need to save state since some builds exceed 6hrs of runtime. In my initial experiments, I wasn’t able to set up ccache properly but this hacky version worked. I had to move fast to help some downstream projects. I think with the next free cycles we can reiterate to ccache and see if it delivers extra value. 

Most important is arm/sbsa support though, we by the conversation we have?

johnnynunez:
> > @tridao @ko3n1g Why should we implement [CCACHE](https://ccache.dev/?utm_source=chatgpt.com) in CI pipeline?
> > Github could catch binaries that are compiled and not modified, and not rerun again all
> > Also cibuildwheel can generate for windows and linux and N python versions
> > [pypa/cibuildwheel#779](https://github.com/pypa/cibuildwheel/discussions/779)
> > this should speedup a lot compilations
> 
> We need to save state since some builds exceed 6hrs of runtime. In my initial experiments, I wasn’t able to set up ccache properly but this hacky version worked. I had to move fast to help some downstream projects. I think with the next free cycles we can reiterate to ccache and see if it delivers extra value.
> 
> Most important is arm/sbsa support though, we by the conversation we have?

true, for me is so important sbsa support...
i build for thor and orin here:
https://pypi.jetson-ai-lab.io/

johnnynunez:
@tridao i just understand this article...
I will do a refactor in this PR:

https://developer.nvidia.com/blog/nvidia-blackwell-and-nvidia-cuda-12-9-introduce-family-specific-architecture-features/
Family-specific features are similar to architecture-specific features, except that they are supported by devices of more than one minor compute capability. All devices within a family share the same major compute capability version. Family-specific features are guaranteed to be available in the same family of GPUs, which includes later GPUs of the same major compute capability and higher minor compute capability. 


this means that in the past we had to do(GB200,GB300,Thor,RTX50, Spark):
```bash
cuda_archs_loose_intersection(SCALED_MM_ARCHS "10.0a;10.3a,11.0a;12.0a;12.1a" "${CUDA_ARCHS}")
```
now it is more simple and beautiful:
```bash
cuda_archs_loose_intersection(SCALED_MM_ARCHS "10.0f;11.0f;12.0f" "${CUDA_ARCHS}")
```

PR #1835, opened 2025-08-22
ci: Move build job to workflow template

ko3n1g:
This PR tries to only move the build-wheel logic into a workflow template. 

A few noticeable changes:

Instead of re-checking for the version number in the build job (by parsing `GITHUB_REF`), I'm suggesting to extract the version once in `setup_release` and pass it downstream to follow up tasks (`needs.setup_release.outputs.release-version`). 

This allows to passing the release-version as input into the workflow template and to change:

```
- name: Get Release with tag
        id: get_current_release
        uses: joutvhu/get-release@v1
        with:
          tag_name: ${{ inputs.release-version }}
```

### DIFF
--- a/.github/workflows/_build.yml
+++ b/.github/workflows/_build.yml
@@ -42,10 +42,10 @@ jobs:
     runs-on: ${{ inputs.runs-on }}
     name: Build wheel (${{ inputs.release-version }}-${{ inputs.python-version }}-${{ inputs.cuda-version }}-${{ inputs.torch-version }}-${{ inputs.cxx11-abi }})
     steps:
-      - name: Set up Python
-        uses: actions/setup-python@v5
+      - name: Checkout
+        uses: actions/checkout@v4
         with:
-          python-version: ${{ inputs.python-version }}
+          ref: ${{ inputs.release-version }}
 
       - name: Set up Python
         uses: actions/setup-python@v5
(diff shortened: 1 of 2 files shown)


johnnynunez:
@tridao @ko3n1g 
about cuda manylinux, normally people create an script bash and they code cuda installation there. Then you can added script in pyproject.toml

script.sh
```bash
yum install -y yum-utils \
 && yum-config-manager --add-repo \
        https://developer.download.nvidia.com/compute/cuda/repos/rhel9/sbsa/cuda-rhel9.repo \
 && yum -y install \
        cuda-compiler-${VER}.${ARCH} \
        cuda-libraries-${VER}.${ARCH} \
        cuda-libraries-devel-${VER}.${ARCH} \
 && yum -y erase yum-utils \
 && yum clean all \
 && rm -rf /var/cache/yum/* \
 && echo "/usr/local/cuda/lib64" > /etc/ld.so.conf.d/999_nvidia_cuda.conf


export PATH="/usr/local/cuda/bin:${PATH}" \
    LD_LIBRARY_PATH="/usr/local/cuda/lib64:${LD_LIBRARY_PATH}" \
    CUDA_HOME=/usr/local/cuda \
    CUDA_ROOT=/usr/local/cuda \
    CUDA_PATH=/usr/local/cuda \
    CUDADIR=/usr/local/cuda
```

pyproject.toml
```
[tool.cibuildwheel]
archs = "auto64"
build-verbosity = 1
skip = ""

before-all = ".github/workflows/before_all.sh"
before-build = ".github/workflows/before_build.sh"

manylinux-x86_64-image = "quay.io/pypa/manylinux_2_28_x86_64"
manylinux-aarch64-image = "quay.io/pypa/manylinux_2_28_aarch64"
``` 

PD: the code can be generic using uname -r or target aarch or create 2 scripts for each arch

PR #1840, opened 2025-08-26
Refactors to enable FlexAttention

drisspg:
# Summary

This currently only enables score_mods with inline global loads. For a majority of score_mods I have found that unless you are indexing along the kv_index dim the slow down due to global loads is too high compared to the additional math inlined. KV_indexing however is a major performance hit ~700 Tflops and we will likely need to design something smarter here that goes through smem if there is room.

PT stack: https://github.com/pytorch/pytorch/pull/162031

## Changes for SM100/SM80/90:
* TensorSSA based score_mods
* Optional list of buffers that can used in score_mod functions
* Allow for caller to pass in lse and out instead of creating (needed for inductor manage allocations)
* Abstracted out softcap to score_mod -> No perf hit for h100 and pretty bad perf for blackwell but now supported

### Tests
One thing that I really thought was going to be necessary was to move the loads out of the softmax hotpath. And potentially have them pre-loaded to in SMEM (although I am not sure how much room we have left). However at least from early testing, for something like 

Example usage:
```Py
def dual_buffer_bias(b, n_heads, q, kv, dtype):
    """Dual buffer loading (tests loading from 2 separate tensors)."""
    head_bias = torch.randn(n_heads, device="cuda", dtype=dtype) * 0.2
    pos_scale = torch.randn(q + kv, device="cuda", dtype=dtype) * 0.1
    def dual_buffer_mod(score, b, h, q_idx, kv_idx):
        head_component = head_bias[h]
        pos_component = pos_scale[kv_idx]
        return score + head_component + pos_component
    return dual_buffer_mod
 ```

which currently produces:
```Py
    @cute.jit
    def score_mod(tSrS_ssa, b_idx, h_idx, q_idx, kv_idx, buffers):
        in_ptr4 = buffers[0]
        in_ptr5 = buffers[1]
        tmp0 = tSrS_ssa
        tmp1 = h_idx
        tmp2 = cute.make_fragment(1, cutlass.Int32)
        tmp3 = tmp2.store(tmp1)
        tmp4 = cute.make_fragment(1, cutlass.BFloat16)
        tmp5 = tmp2[0]
        tmp6 = tmp4[0] = (in_ptr4[tmp5])
        tmp7 = (tmp4.load()).to(cutlass.Float32)
        tmp8 = (tmp0 + tmp7)
        tmp9 = kv_idx
        tmp10 = tmp2.store(tmp9)
        tmp11 = tmp4[0] = (in_ptr5[tmp5])
        tmp12 = (tmp8 + tmp7)
        tSrS_ssa = tmp12

        return tSrS_ssa
```

We go from 
`FAv4:    3108.38 μs | 1420.43 TFLOPs`
to:
`FAv4:    3485.34 μs | 1266.80 TFLOPs`

### 1. General Perf Hit from Score Modifications

**Causal, HeadDim=128 Performance Comparison**

| SeqLen | No score_mod | | Identity score_mod (`lambda x, *_: return x`) |
|--------|-------------|---|-------------|
| 1024   | 0.168ms, 819.3 TFLOPS | | 0.179ms, 769.0 TFLOPS |
| 2048   | 0.242ms, 1136.1 TFLOPS | | 0.261ms, 1052.8 TFLOPS |
| 4096   | 0.398ms, 1379.7 TFLOPS | | 0.433ms, 1269.1 TFLOPS |
| 8192   | 0.726ms, 1513.7 TFLOPS | | 0.789ms, 1393.8 TFLOPS |
| 16384  | 1.385ms, 1587.9 TFLOPS | | 1.509ms, 1457.1 TFLOPS |
| 32768  | 2.697ms, 1630.6 TFLOPS | | 2.948ms, 1491.6 TFLOPS |

### 2. Dual Buffer Loading Performance

For dual buffer bias loading from 2 separate tensors:
- **Before:** 3108.38 μs | 1420.43 TFLOPs
- **After:** 3485.34 μs | 1266.80 TFLOPs

Which honestly is much less of a drop than I expected.

Motivation for all these changes:
```Shell
[Results]
Flex+FA:    3463.58 μs | 1274.76 TFLOPs
Triton: 11915.89 μs | 370.53 TFLOPs
Speedup: 3.44x
```
This is a little unfair because the PT implementation is block-sparse(TO BE DONE AS A FOLLOW UP HERE), but that being said a very significant performance increase.

### 3. Softcap Performance

**NOTE:** I just realized that on SM100 it doesn't appear like softcapping was even working..

#### SM100 Softcap Results (pretty miserable, need to grab H100 and compare)

| SeqLen | Latency | TFLOPS |
|--------|---------|--------|
| 1024   | 0.555ms | 247.7  |
| 2048   | 1.011ms | 271.9  |
| 4096   | 1.936ms | 284.0  |
| 8192   | 3.778ms | 291.0  |
| 16384  | 7.479ms | 294.0  |
| 32768  | 14.884ms| 295.5  |

#### H100 Softcap Comparison (w/ power limit)

| SeqLen | Before | | After |
|--------|--------|---|-------|
| 1024   | 0.531ms, 258.7 TFLOPS | | 0.540ms, 239.1 TFLOPS |
| 2048   | 0.910ms, 301.9 TFLOPS | | 0.915ms, 282.9 TFLOPS |
| 4096   | 1.669ms, 329.3 TFLOPS | | 1.663ms, 311.3 TFLOPS |
| 8192   | 3.237ms, 339.7 TFLOPS | | 3.184ms, 327.8 TFLOPS |
| 16384  | 6.796ms, 323.6 TFLOPS | | 6.740ms, 297.4 TFLOPS |
| 32768  | 13.823ms, 318.2 TFLOPS | | 13.527ms, 297.6 TFLOPS |

### Tests
I have lots of tests in PT and brought some of them over w/ score_mods that were produced via codegen -> not very readable but they work :)

On Sm100 machine also could attach a picture of all green :)
Screenshot 2025-10-07 at 6 08 14 PM

On sm90 Machine
Screenshot 2025-09-15 at 3 39 47 PM

### DIFF
--- a/flash_attn/cute/flash_fwd.py
+++ b/flash_attn/cute/flash_fwd.py
@@ -23,7 +23,7 @@ from flash_attn.cute import ampere_helpers as sm80_utils
 from flash_attn.cute import hopper_helpers as sm90_utils
 from flash_attn.cute import utils
 from flash_attn.cute.mask import AttentionMask
-from flash_attn.cute.softmax import Softmax
+from flash_attn.cute.softmax import Softmax, apply_score_mod_inner
 from flash_attn.cute.seqlen_info import SeqlenInfoQK
 from flash_attn.cute.block_info import BlockInfo
 from flash_attn.cute import pipeline
@@ -51,6 +51,7 @@ class FlashAttentionForwardBase:
         num_threads: int = 128,
         Q_in_regs: bool = False,
         score_mod: cutlass.Constexpr | None = None,
+        has_buffers: bool = False,
     ):
         """Initializes the configuration for a flash attention kernel.
 
@@ -67,7 +68,7 @@ class FlashAttentionForwardBase:
         :type num_threads: int
         :param is_causal: is causal
         :param score_mod: A callable that takes the attention scores and applies a modification.
-            Callable signature: 
+            Callable signature: ``score_mod(scores, batch_idx, head_idx, q_idx, kv_idx, buffers) -> Any``
         """
         self.dtype = dtype
         # padding head_dim to a multiple of 16 as k_block_size
@@ -89,6 +90,11 @@ class FlashAttentionForwardBase:
         self.num_stages = num_stages
         self.Q_in_regs = Q_in_regs
         self.score_mod = score_mod
+        self.qk_acc_dtype = Float32
+        if cutlass.const_expr(has_buffers):
+            self.vec_size: cutlass.Constexpr = 1
+        else:
+            self.vec_size: cutlass.Constexpr = 2
 
     @staticmethod
     def can_implement(
@@ -934,11 +940,15 @@ class FlashAttentionForwardSm80(FlashAttentionForwardBase):
         )
         if cutlass.const_expr(score_mod is not None):
             self.apply_score_mod(
-                acc_S, mma_params.thr_mma_qk, score_mod,
-                batch_idx, head_idx, m_block, n_block,
-                softmax_scale=softmax.softmax_scale,
+                acc_S,
+                mma_params.thr_mma_qk,
+                batch_idx,
+                head_idx,
+                m_block,
+                n_block,
+                softmax=softmax,
                 buffers=buffers,
-                fastdiv_mods=fastdiv_mods
+                fastdiv_mods=fastdiv_mods,
             )
             
         smem_pipe_write = self.advance_pipeline(smem_pipe_write)
@@ -1692,11 +1702,15 @@ class FlashAttentionForwardSm90(FlashAttentionForwardBase):
                 # Use vectorized score modification
                 if cutlass.const_expr(score_mod is not None):
                     self.apply_score_mod(
-                        acc_S, thr_mma_qk, score_mod,
-                        batch_idx, head_idx, m_block, n_block_max - 1,
-                        softmax_scale=softmax.softmax_scale,
+                        acc_S,
+                        thr_mma_qk,
+                        batch_idx,
+                        head_idx,
+                        m_block,
+                        n_block_max - 1,
+                        softmax=softmax,
                         buffers=buffers,
-                        fastdiv_mods=fastdiv_mods
+                        fastdiv_mods=fastdiv_mods,
                     )
                 # if cute.arch.thread_idx()[0] == 128: cute.print_tensor(utils.make_acc_tensor_mn_view(acc_S))
                 mask_fn(acc_S, n_block=n_block_max - 1, mask_seqlen=True)
@@ -1843,11 +1857,15 @@ class FlashAttentionForwardSm90(FlashAttentionForwardBase):
         pipeline_k.consumer_release(smem_pipe_read)
         if cutlass.const_expr(score_mod is not None):
             self.apply_score_mod(
-                acc_S, thr_mma_qk, score_mod,
-                batch_idx, head_idx, m_block, n_block,
-                softmax_scale=softmax.softmax_scale,
+                acc_S,
+                thr_mma_qk,
+                batch_idx,
+                head_idx,
+                m_block,
+                n_block,
+                softmax=softmax,
                 buffers=buffers,
-                fastdiv_mods=fastdiv_mods
+                fastdiv_mods=fastdiv_mods,
             )
         if const_expr(mask_fn is not None):
             mask_fn(acc_S, n_block=n_block)
@@ -1923,11 +1941,15 @@ class FlashAttentionForwardSm90(FlashAttentionForwardBase):
         pipeline_k.consumer_release(smem_pipe_read)
         if cutlass.const_expr(score_mod is not None):
             self.apply_score_mod(
-                acc_S, thr_mma_qk, score_mod,
-                batch_idx, head_idx, m_block, n_block,
-                softmax_scale=softmax.softmax_scale,
+                acc_S,
+                thr_mma_qk,
+                batch_idx,
+                head_idx,
+                m_block,
+                n_block,
+                softmax=softmax,
                 buffers=buffers,
-                fastdiv_mods=fastdiv_mods
+                fastdiv_mods=fastdiv_mods,
             )
         # if cute.arch.thread_idx()[0] == 128: cute.print_tensor(utils.make_acc_tensor_mn_view(acc_S))
         if const_expr(mask_fn is not None):
@@ -1965,73 +1987,32 @@ class FlashAttentionForwardSm90(FlashAttentionForwardBase):
         self,
         acc_S,
         thr_mma_qk,
-        score_mod,
         batch_idx,
         head_idx,
         m_block,
         n_block,
-        softmax_scale=None,
+        softmax,
         buffers=None,
         fastdiv_mods=None,
-        VEC_SIZE: cutlass.Constexpr[int] = 1,
     ):
-        # Get index tensors with proper domain offset
+        # Prepare index tensor
         cS = cute.make_identity_tensor((self.m_block_size, self.n_block_size))
         cS = cute.domain_offset((m_block * self.m_block_size, n_block * self.n_block_size), cS)
         tScS = thr_mma_qk.partition_C(cS)
-        
-        # Vectorized processing similar to SM100
-        n_vals = cutlass.const_expr(cute.size(acc_S.shape))
-        score_vec = cute.make_fragment(VEC_SIZE, Float32)
-        kv_idx_vec = cute.make_fragment(VEC_SIZE, cutlass.Int32)
-        
-        # Create broadcasted fragments for constant values
-        batch_idx_vec = utils.broadcast_scalar_to_vec(batch_idx, kv_idx_vec, cutlass.Int32)
-        head_idx_vec = utils.broadcast_scalar_to_vec(head_idx, kv_idx_vec, cutlass.Int32)
-        q_idx_vec = cute.make_fragment(VEC_SIZE, cutlass.Int32)
-        
-        # Load SSA values once  
-        batch_idx_ssa = batch_idx_vec.load()
-        head_idx_ssa = head_idx_vec.load()
 
-        # Build SSA slices and call into scoremod / writeback
-        for i in cutlass.range(0, n_vals, VEC_SIZE, unroll_full=True):
-            for j in cutlass.range(VEC_SIZE, unroll_full=True):
-                score_vec[j] = acc_S[i + j]
-                if softmax_scale is not None:
-                    score_vec[j] = score_vec[j] * softmax_scale
-
-                # Use FastDivmod for bounds checking if buffers and divmods are present
-                if cutlass.const_expr(buffers is not None and fastdiv_mods is not None):
-                    seqlen_q_divmod, seqlen_k_divmod = fastdiv_mods
-                    _, q_idx_wrapped = seqlen_q_divmod.divmod(tScS[i + j][0])
-                    _, kv_idx_wrapped = seqlen_k_divmod.divmod(tScS[i + j][1])
-                    q_idx_vec[j] = q_idx_wrapped
-                    kv_idx_vec[j] = kv_idx_wrapped
-                else:
-                    # No bounds checking needed - direct indexing
-                    q_idx_vec[j] = tScS[i + j][0]
-                    kv_idx_vec[j] = tScS[i + j][1]
-            score_ssa = score_vec.load()
-            q_idx_ssa = q_idx_vec.load()
-            kv_idx_ssa = kv_idx_vec.load()
-
-            buffer_args = []
-            if cutlass.const_expr(buffers is not None):
-                buffer_args = buffers
-
-            post_mod_scores = score_mod(
-                score_ssa,
-                batch_idx_ssa,
-                head_idx_ssa,
-                q_idx=q_idx_ssa,
-                kv_idx=kv_idx_ssa,
-                buffers=buffer_args
-            )
-
-            score_vec.store(post_mod_scores)
-            for j in cutlass.range(VEC_SIZE, unroll_full=True):
-                acc_S[i + j] = score_vec[j]
+        apply_score_mod_inner(
+            acc_S,
+            tScS,
+            self.score_mod,
+            batch_idx,
+            head_idx,
+            softmax.softmax_scale,
+            self.vec_size,
+            self.qk_acc_dtype,
+            buffers,
+            fastdiv_mods,
+            constant_q_idx=None
+        )
 
     def warp_scheduler_barrier_sync(self):
         if const_expr(self.use_scheduler_barrier):
--- a/flash_attn/cute/flash_fwd.py
+++ b/flash_attn/cute/flash_fwd.py
@@ -7,7 +7,7 @@
 
 import math
 from types import SimpleNamespace
-from typing import Type, Callable, Optional, Tuple
+from typing import Type, Callable, Optional
 from functools import partial
 
 import cuda.bindings.driver as cuda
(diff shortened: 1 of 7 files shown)


flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1907,3 +1938,97 @@ def make_and_init_load_kv_pipeline(self, load_kv_mbar_ptr):
...
+        # Load SSA values once
+        batch_idx_ssa = batch_idx_vec.load()
+        head_idx_ssa = head_idx_vec.load()
+        q_idx_ssa = q_idx_vec.load()
+
+        # Build SSA slices and call into scoremod / writeback
+        for i in cutlass.range(0, n_vals, VEC_SIZE, unroll_full=True):
+            for j in cutlass.range(VEC_SIZE, unroll_full=True):
+                score_vec[j] = tSrS_t2r[i + j] * softmax.softmax_scale
+                kv_idx_vec[j] = tScS_t2r[i + j][1]
```
drisspg: I need to do modulus wrapping so as to not load from out of bound vals

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -20,6 +20,7 @@
 import cuda.bindings.driver as cuda
 
 import cutlass
+from cutlass._mlir.dialects.llvm import intr_prefetch
```
drisspg: editor mistake

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -465,22 +468,31 @@ class SharedStorage:
...
         self.shared_storage = SharedStorage
 
-        # If there's tanh softcapping, we do tanh(scores * softmax_scale / softcap_val) * softcap_val.
-        # Right after this, we multiply by log2(e) before applying exp2.
-        # To reduce the number of instructions, we instead pre-multiply softmax_scale / softcap_val
-        # (assigning it to softcap_val) and pre-multiply softcap_val * log2(e)
-        # (assigning it to softmax_scale_log2).
         LOG2_E = math.log2(math.e)
-        if const_expr(softcap is None):
+        if const_expr(self.score_mod is None):
```
drisspg, line 476: I think that this is alittle counter intuitive, but was least intrusive in terms of changes, I can refactor if we dont like it

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -465,22 +468,31 @@ class SharedStorage:
...
+            softmax_scale_log2 = LOG2_E
+            softmax_scale = softmax_scale
+
         if const_expr(window_size_left is not None):
             window_size_left = Int32(window_size_left)
         if const_expr(window_size_right is not None):
             window_size_right = Int32(window_size_right)
+
+        fastdiv_mods = None
+        # Create FastDivmod objects only if buffers are present
```
drisspg: not needed comment

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -539,7 +553,7 @@ def kernel(
         tma_atom_V: cute.CopyAtom,
         tma_atom_O: Optional[cute.CopyAtom],
         softmax_scale_log2: Float32,
-        softcap_val: Optional[Float32],
+        softmax_scale: Float32 | None,
```
drisspg, line 559: not sure yet what version of Python is minimal should probs bump to Optional *

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -788,13 +806,16 @@ def kernel(
...
+                buffers=buffers,
+                fastdiv_mods=fastdiv_mods,
             )
 
             if const_expr(not self.s0_s1_barrier):
                 stage = Int32(0 if warp_idx < self.softmax1_warp_ids[0] else 1)
                 softmax_loop(
                     stage=stage,
-                    tStSi=cute.make_tensor(tStS.iterator + (self.tmem_s_offset[0] if stage == 0 else self.tmem_s_offset[1]), tStS.layout))
+                    tStSi=cute.make_tensor(tStS.iterator + (self.tmem_s_offset[0] if stage == 0 else self.tmem_s_offset[1]), tStS.layout),
```
drisspg: formatting

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1243,6 +1267,12 @@ def softmax_loop(
                 tStP_r2t=tStP_r2t,
                 sScale=sScale,
                 stage=stage,
+                batch_idx=batch_idx,
+                head_idx=head_idx,
+                m_block=m_block * 2 + stage,
```
drisspg: want to figure out statically if this should always be *2 not sure if its possible to have another work mapping that breaks this

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1907,3 +1958,113 @@ def make_and_init_load_kv_pipeline(self, load_kv_mbar_ptr):
...
+        thr_tmem_load,
+        thr_mma_qk,
+        batch_idx,
+        head_idx,
+        m_block,
+        n_block,
+        softmax,
+        buffers = None,
+        fastdiv_mods = (None, None),
+        VEC_SIZE: cutlass.Constexpr[int] = 1,
```
drisspg: VEC_SIZE must = 1 for for current global load patterns, at least in inductor). I tried a number of different vec_sizes w/ no buffers and never found a perf benefit

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1907,3 +1958,113 @@ def make_and_init_load_kv_pipeline(self, load_kv_mbar_ptr):
...
+        # Load SSA values once
+        batch_idx_ssa = batch_idx_vec.load()
+        head_idx_ssa = head_idx_vec.load()
+        q_idx_ssa = q_idx_vec.load()
+
+        # Build SSA slices and call into scoremod / writeback
+        for i in cutlass.range(0, n_vals, VEC_SIZE, unroll_full=True):
+            for j in cutlass.range(VEC_SIZE, unroll_full=True):
+                score_vec[j] = tSrS_t2r[i + j] * softmax.softmax_scale
+                # Use FastDivmod for bounds checking if buffers and divmods are present
```
drisspg: remove non needed comment

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1907,3 +1958,113 @@ def make_and_init_load_kv_pipeline(self, load_kv_mbar_ptr):
...
+
+            score_vec.store(post_mod_scores)
+            for j in cutlass.range(VEC_SIZE, unroll_full=True):
+                tSrS_t2r[i + j] = score_vec[j]
+
+
+
+### Grave yard
+
+# PRETTY WAY but uses to much rmem
```
drisspg: this is pretty but indeed shud remove

flash_attn/cute/interface.py
```diff
@@ -137,10 +151,25 @@ def _flash_attn_fwd(
...
     q_batch_seqlen_shape = (batch_size, seqlen_q) if cu_seqlens_q is None else (total_q,)
-    out = torch.empty(*q_batch_seqlen_shape, num_head, head_dim_v, dtype=out_torch_dtype, device=device)
     lse_shape = (batch_size, num_head, seqlen_q) if cu_seqlens_q is None else (num_head, total_q)
     requires_grad = q.requires_grad or k.requires_grad or v.requires_grad
-    lse = torch.empty(lse_shape, dtype=torch.float32, device=device) if requires_grad else None
+    
+    if out is None:
+        out = torch.empty(*q_batch_seqlen_shape, num_head, head_dim_v, dtype=out_torch_dtype, device=device)
+    else:
+        expected_out_shape = (*q_batch_seqlen_shape, num_head, head_dim_v)
```
drisspg, line 160: could make a helper so this isn't  inline

flash_attn/cute/utils.py
```diff
@@ -12,6 +13,29 @@
...
 
 
+def hash_callable(func: Callable) -> str:
+    """Hash a callable by combining bytecode and closure variables."""
+    parts = []
+
+    if hasattr(func, '__code__'):
+        parts.append(func.__code__.co_code)
+
+    if hasattr(func, '__closure__') and func.__closure__:
```
drisspg: the closure inclusion is a lil sketch

flash_attn/cute/flash_fwd.py
```diff
@@ -65,6 +66,8 @@ def __init__(
         :param num_threads: number of threads
         :type num_threads: int
         :param is_causal: is causal
+        :param score_mod: A callable that takes the attention scores and applies a modification.
+            Callable signature:
```
drisspg: fix this

flash_attn/cute/utils.py
```diff
@@ -12,6 +13,29 @@
...
+            except:
+                pass
+
+    combined = b''.join(str(p).encode() for p in parts)
+    return hashlib.sha256(combined).hexdigest()
+
+
+def create_softcap_scoremod(softcap_val):
+    def scoremod_premask_fn(acc_S_SSA, batch_idx, head_idx, q_idx, kv_idx, buffers):
+        return cute.math.tanh(acc_S_SSA * softcap_val, fastmath=True)
```
drisspg: this is not right should be softcap_val * (tanh(score * val))

where

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1907,3 +1959,94 @@ def make_and_init_load_kv_pipeline(self, load_kv_mbar_ptr):
     #     cute.arch.barrier_arrive(
     #         barrier_id=int(NamedBarrierFwd.WarpSchedulerWG1) + next_wg, number_of_threads=2 * 128,
     #     )
+
+    @cute.jit
```
v0i0, line 1968: how is this different from the apply_score mod in `flash_fwd.py`? maybe move to `softmax.py` and share?

flash_attn/cute/flash_fwd_sm100.py
v0i0: huh i would have expected vectorized fp32 sass for vec_size == 2, do you happen to have the sass?

flash_attn/cute/utils.py
v0i0: maybe give users a way to override this? this might be a lot less hacky if done from the inductor end

flash_attn/cute/utils.py
```diff
@@ -589,3 +617,31 @@ def coord_offset_i64(
...
+    a_broadcasted = cute.make_tensor(
+        a.iterator,
+        cute.make_layout(b.layout.shape, stride=stride_zero)
+    )
+    return a_broadcasted
+
+
+@cute.jit
+def broadcast_scalar_to_vec(scalar_value: cutlass.Numeric, b: cute.Tensor, dtype) -> cute.Tensor:
+    """Create a fragment of vec_size elements, all set to scalar_value.
```
v0i0: that's a straight-up lie, this create a single-element fragment, and then creates a broadcast tensor

flash_attn/cute/utils.py
v0i0: also might want this to be recursive somehow? what if we're closing over a function?

v0i0 (approved):
this is awesome!

flash_attn/cute/utils.py
drisspg: So upon some further reflection, since we are threading through the buffers from top to bottom. I think that simply hashing on the function as a code string might suffice. e.g. this might be a little low on recall but I think should have good precision

flash_attn/cute/flash_fwd_sm100.py
drisspg: this was my expectation as well and I found very counter intuitive, ill get some sass to compare

drisspg: just tested again after upgrading to cutedsl and for simple causal vecsize 2 does have an impact of around 100 tflops but for something like times 2 deson't have any meaningful impact. This feels very much like a good thing to autotune over

flash_attn/cute/utils.py
v0i0: would this correctly deal with e.g. the softcap with multiple different softcap scales? those are only present in the closure, right?

drisspg: The updated version that basically hash's the closures code and should `bake in` the different softcap vals

flash_attn/cute/flash_fwd_sm100.py
drisspg: tried 2, 4, 8, perf caps at 2 for the 5 different examples I looked at. updates so that if we have buffers we drop back to 1 else we set to 2

flash_attn/cute/flash_fwd_sm100.py
drisspg: So I thought about this and IMO they are just different enough to not warrant being merged I have been playing w different ways to do it and it just isn't that clean

drisspg: okay got it to work

flash_attn/cute/utils.py
drisspg: ohhh nvm I see what you are saying hmmm

flash_attn/cute/interface.py
```diff
@@ -73,7 +72,22 @@ def _flash_attn_fwd(
     num_threads: int = 384,
     pack_gqa: Optional[bool] = None,
     _compute_capability: Optional[int] = None,
+    score_mod: Callable | None = None,
+    return_lse: bool = False,
```
drisspg, line 76: we should probably just have the no_grad setting set this to true in the autograd func

flash_attn/cute/flash_fwd.py
```diff
@@ -580,27 +589,33 @@ def __call__(
...
-            
+            # NB: If a user passes in a score mod, we want to apply the score-mod in the sm_scaled qk
+            # But in the original base 10. We hijack softmax_scale_log2 to just be the change of base
+            # and correctly apply the softmax_scale prior to score_mod in the softmax step
+            softmax_scale_log2 = Float32(LOG2_E)
+            softmax_scale = Float32(softmax_scale)
+
+        fastdiv_mods = None
+        if cutlass.const_expr(buffers is not None):
+            seqlen_q = cute.size(mQ.shape[0])
```
tridao, line 605: This wouldn't work for varlen case right?
Do we focus on the non-varlen case for now?

tridao: Also there's the PackGQA case where mQ.shape[0] is actually seqlen_q * qheads_per_kv_head

flash_attn/cute/flash_fwd_sm100.py
tridao: it should be `self.q_stage`, but so far we've kept that at 2. There are other places where we hardcode m_block * 2 so it's fine to do that here.

flash_attn/cute/flash_fwd.py
tridao: Maybe for the varlen case we set the divmod to be None, and inside the kernel do `if const_expr`. But if we don't care about the varlen case for now we can just leave this as is.

tridao:
Looks good overall. 
Ideally softcap perf wouldn't be affected, but i'm not worried about that now given we haven't optimized for softcap at all.

tridao (approved):

flash_attn/cute/flash_fwd.py
drisspg: Okay, for both varlen and pack I added xfails for now. I will fix the packgqa indexing in a follow up so that its easier to review. Will also do varlen at least for score mods w/ no buffers for global loads we might need to fallback to non fastdiv mod

PR #1851, opened 2025-08-28
[`FA3`] Allow returning LSE via kwarg

vasqu:
Allow outputting the lse via a kwarg `return_attn_probs` just as in FA2 currently

cc @tridao

### DIFF
--- a/hopper/flash_attn_interface.py
+++ b/hopper/flash_attn_interface.py
@@ -168,6 +168,7 @@ class FlashAttnQKVPackedFunc(torch.autograd.Function):
         deterministic=False,
         num_heads_q=None,
         sm_margin=0,
+        return_softmax=False,
     ):
         if softmax_scale is None:
             softmax_scale = qkv.shape[-1] ** (-0.5)
@@ -210,8 +211,7 @@ class FlashAttnQKVPackedFunc(torch.autograd.Function):
         ctx.deterministic = deterministic
         ctx.ndim = qkv.dim()
         ctx.sm_margin = sm_margin
-        # return out, softmax_lse
-        return out
+        return (out, softmax_lse) if return_softmax else out
 
     @staticmethod
     def backward(ctx, dout, *args):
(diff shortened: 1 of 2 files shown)


tridao:
There should be an option `return_lse`. The standard interface is to not return lse since that's how FA2 / sdpa does it.

hopper/flash_attn_interface.py
```diff
@@ -168,6 +168,7 @@ def forward(
         deterministic=False,
         num_heads_q=None,
         sm_margin=0,
+        return_softmax=False,
```
vasqu, line 171: Checked with the fa2 interface, internally it's called `return_softmax`, e.g. https://github.com/Dao-AILab/flash-attention/blob/b2476552432fd6ac991003db4564eb289dd77332/flash_attn/flash_attn_interface.py#L911

hopper/flash_attn_interface.py
```diff
@@ -515,6 +519,7 @@ def flash_attn_func(
     pack_gqa=None,
     deterministic=False,
     sm_margin=0,
+    return_attn_probs=False,
```
vasqu, line 522: Also aligned with fa2 (user facing interface) https://github.com/Dao-AILab/flash-attention/blob/b2476552432fd6ac991003db4564eb289dd77332/flash_attn/flash_attn_interface.py#L1146

hopper/flash_attn_interface.py
```diff
@@ -210,8 +211,7 @@ def forward(
         ctx.deterministic = deterministic
         ctx.ndim = qkv.dim()
         ctx.sm_margin = sm_margin
-        # return out, softmax_lse
-        return out
+        return (out, softmax_lse) if return_softmax else out
```
vasqu, line 214:
```suggestion
        return (out, softmax_lse, *rest) if return_softmax else out
```
Could also be done, but kept it simple - lmk if that should be changed

tridao:
Thanks!

PR #1856, opened 2025-09-02
[BugFix] fix softcap condition

mingyangHao:
softcap should only be referenced when its not none, currently the logic is reversed and will result in an error


mingyangHao:
Update Patch 2 to align with sm_100

mingyangHao:
I just noticed there are more bugs, trying to fix them all.

mingyangHao:
I believe at this point flash_fwd.py could run successfully, but with one more changes i am not sure whether to add: SeqlenInfoQK is requiring batch_idx as an input where sm80 kernel do not pass. In my local version I just make it optional, but that might affect the logic of your sm100 kernel.

tridao:
Thank you! The code for Sm80 eventually should have the same interface w Sm90 and Sm100 but that's for the future.

PR #1860, opened 2025-09-03
[BUG] CUDA 13: make FA3 compatible with CUDA 13 Builds

johnnynunez:
@tridao Fix CUDA barrier init crash when num_consumers < NumThreadsPerWarpGroup

Previously, integer division caused num_consumer_warpgroups_per_cluster to be 0 when params.num_consumers (e.g., 32) was less than NumThreadsPerWarpGroup (128), leading to a compiler failure during barrier initialization. Changed to round-up division to ensure a minimum value of 1.


johnnynunez:
@tridao @ko3n1g could you merge it? it unlocks cuda 13

shahizat:
Great job, @johnnynunez. Thanks!

PR #1867, opened 2025-09-05
Modernize Build System with a minimal pyproject.toml file 

### LINKED ISSUE #1843: Use `pyproject.toml` to simplify packaging and metadata without breaking functionality
[pyproject.toml](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/) simplifies declaring configs for moving static metadata, while leaving complex build logic in setup.py. Since the release of [PEP 621](https://peps.python.org/pep-0621/), the pyproject.toml file has been increasingly adopted for simpler usage and its enhanced functionality.

This can make the packaging and usage of FA simpler to use and install.  WDYT?

ParagEkbote:
## Motivation

Installing FlashAttention can be quite painful when certain files aren’t packaged or when different tools (`pip`, `python setup.py install`, `build`) behave inconsistently. This PR moves the project to a modern, standard packaging setup that reduces those headaches. It also aligns with Python packaging standards (PEP [518](https://peps.python.org/pep-0518/), [621](https://peps.python.org/pep-0621/)), making the project compatible with modern tooling. Could you please review?

cc: @tridao

Fixes #1843

## What this PR does not do

- Does not change the CUDA extension code.

- Does not remove setup.py.

- Does not increase runtime dependencies.


ParagEkbote:
Could you please take a look when you have some time? Also, let me know if I should adjust anything

cc: @tridao

PR #1870, opened 2025-09-06
Improve FA3 deterministic backward

jayhshah:
This PR improves performance of deterministic FA3 backward and also fixes a few bugs for it:

1. Traverse mblocks in reverse by default, which is much better for deterministic causal backward. The attached benchmark shows ~25% perf improvement.
2. Extend bwd scheduler with head swizzle (`SingleTileBwdLPTScheduler`) to varlen case.
3. Fixes errors with dq_semaphore not being incremented for local and dk/dv_semaphore for varlen. For the latter, we pass the number of batches as the dimension of last mode of `shape_dKaccum`, since we write in the epilogue
```int const num_batch = get<2>(params.shape_dKaccum);```

[bwd_determinism_bench.txt](https://github.com/user-attachments/files/22184871/bwd_determinism_bench.txt)


jayhshah:
Closing this PR since we're now using SPT schedule for backward determinism.

PR #1881, opened 2025-09-12
benchmark: qualify all attention backends by methods list

rajesh-s:
Only run the user-selected methods for measurement. This will avoid confusion if someone is capturing NCU traces for instances by only isolating one of the methods.

Related discussion: https://github.com/Dao-AILab/flash-attention/discussions/1879


tridao:
Thanks!

PR #1888, opened 2025-09-15
[NVIDIA] cutlass v4.2.0

johnnynunez:
making possible from python >=3.10  
remove 3.9... EOL October
cc @tridao @ko3n1g


tridao:
Thanks! I'll need to check if cutlass-dsl 4.2 requires changes to the current cute code.

johnnynunez:
> Thanks! I'll need to check if cutlass-dsl 4.2 requires changes to the current cute code.

https://github.com/NVIDIA/cutlass/releases/tag/v4.2.0 here is the doc

PR #1891, opened 2025-09-16
[Cute] Bump pin for CuTeDSL

drisspg:
# Summary
Without these changes I was getting
```Shell
  DSLRuntimeError: 💥💥💥 Error during runtime code generation for function `__call__` 💥💥💥
    Caused exception: cannot access local variable 'acc_S_mn' where it is not associated with a value
```

and the dynamic protocol warnings for softmax: https://docs.nvidia.com/cutlass/media/docs/pythonDSL/cute_dsl_general/dsl_jit_arg_generation.html#direct-protocol-implementation-in-custom-types

Screenshot 2025-09-16 at 8 03 21 PM

### DIFF
--- a/flash_attn/cute/mask.py
+++ b/flash_attn/cute/mask.py
@@ -76,6 +76,12 @@ class AttentionMask:
             causal_row_offset = (
                 1 + self.seqlen_k - n_block * self.n_block_size - self.seqlen_q - thr_col_offset
             )
+            c = 0
+            col_limit_transformed = 0
+            ncol: cute.Constexpr = 0
+            col_limit_right_s = 0
+            mask = 0
+            in_bound = False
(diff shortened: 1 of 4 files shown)


flash_attn/cute/mask.py
```diff
@@ -76,6 +76,12 @@ def apply_mask(
             causal_row_offset = (
                 1 + self.seqlen_k - n_block * self.n_block_size - self.seqlen_q - thr_col_offset
             )
+            c = 0
+            col_limit_transformed = 0
+            ncol: cute.Constexpr = 0
```
drisspg, line 81: this is the only one that feels kinda sketch..

anakinxc: Hi @drisspg 

This should already been fixed with 4.2.1 release

PR #1893, opened 2025-09-17
Improve causal backward determinism perf with SPT schedule

jayhshah:
We improve the performance of deterministic mode for the FA3 backward kernel with causal masking by using a "shortest-processing time first" (SPT) schedule, traversing nblocks from right to left and visiting an mblock in nblock decreasing order for reduce add in the dQ computation. Typical performance gain is as follows:

```
### headdim = 128, causal = True, seqlen = 8192, nheads = 16, nheads_kv = 16, deterministic = True, varlen = False ###
Fav2 fwd: 1.796ms, 306.2 TFLOPS
Fav2 bwd: 8.279ms, 166.0 TFLOPS
CuDNN fwd: 0.972ms, 565.7 TFLOPS
CuDNN bwd: 2.925ms, 469.8 TFLOPS
Fav3 fwd: 0.833ms, 659.8 TFLOPS
Fav3 bwd, non-det: 2.391ms, 574.7 TFLOPS
Fav3 bwd, SPT: 2.900ms, 474.0 TFLOPS
Fav3 bwd, mblock reverse det: 2.962ms, 464.0 TFLOPS
Fav3 bwd, naive det: 3.804ms, 361.3 TFLOPS
```

We also make these changes and fixes similar to the old PR #1870 

1. Extend bwd scheduler with head swizzle (SingleTileBwdLPTScheduler) to varlen case.
2. Fixes errors with dq_semaphore not being incremented for local and dk/dv_semaphore for varlen.

A new test script `hopper/test_flash_attn_bwd_determinism.py` validates determinism with a default setting of 1000 trials for checking exact equality.

### DIFF
--- a/hopper/flash_api_stable.cpp
+++ b/hopper/flash_api_stable.cpp
@@ -1426,6 +1426,7 @@ std::tuple<Tensor, Tensor, Tensor, Tensor, Tensor, Tensor, Tensor, Tensor> mha_b
     int const arch = dprops->major * 10 + dprops->minor;
     int const head_size_rounded = round_up_headdim(std::max(head_size, head_size_v));
     int const head_size_v_rounded = head_size_rounded;
+    STD_TORCH_CHECK(!deterministic || head_size_rounded < 256, "Deterministic backward not supported for hdim 256.");
     // Very important that these match the kernel configs
     bool const is_local = (window_size_left >= 0 || window_size_right >= 0) && !is_causal;
     int const kBlockM_sm90 = head_size_rounded <= 64 ? (is_causal && softcap > 0.0 ? 96 : 128)


hopper/flash_api.cpp
```diff
@@ -1361,6 +1361,7 @@ std::tuple<at::Tensor, at::Tensor, at::Tensor, at::Tensor, at::Tensor, at::Tenso
     int const arch = at::cuda::getCurrentDeviceProperties()->major * 10 + at::cuda::getCurrentDeviceProperties()->minor;
     int const head_size_rounded = round_up_headdim(std::max(head_size, head_size_v));
     int const head_size_v_rounded = head_size_rounded;
+    TORCH_CHECK(!deterministic || head_size_rounded < 256, "Deterministic backward not supported for hdim 256.");
```
janeyx99, line 1364: Can you add a similar check in the stable version of this file now that #1791 has landed?

janeyx99: Should just be STD_TORCH_CHECK

jayhshah: Added

Issue #1918, opened 2025-09-30
Cute-DSL SM90 fwd produces error after bumping to nvidia-cutlass-dsl-4.2.1


imbr92:
The upgrade from nvidia-cutlass-dsl 4.2.0 to 4.2.1 done in 1ceaa984b2f348caea18b39a98458d33b4ea7a09 breaks the cute-dsl SM90 fwd compilation.

Getting this error
```
cutlass.base_dsl.common.DSLRuntimeError: DSLRuntimeError: 🧊🧊🧊 ICE IR Verification Failed 🧊🧊🧊
  Caused exception: Verification failed:
error: "online_softmax"("/mnt/flash-attention/flash_attn/cute/softmax.py":80:0): operand #0 does not dominate this use
 note: "online_softmax"("/mnt/flash-attention/flash_attn/cute/softmax.py":80:0): see current operation: %1966 = "cute.make_fragment_like"(%3011#2) <{elem_type = f32}> : (!cute.memref<f32, rmem, align<32>, "2:1">) -> !cute.memref<f32, rmem, align<32>, "2:1">
 note: unknown: operand defined here (op is neither in a parent nor in a child region)
```

The error occurs somewhere in this function https://github.com/Dao-AILab/flash-attention/blob/5059fd53e602bcc00336bb5cc8a85e50940485cb/flash_attn/cute/softmax.py#L65 related to the uses of `self.row_max[r]` and `self.row_sum[r]`, but I'm not too familiar with MLIR and haven't looked too closely to fix.

Reproducible via same script I put in https://github.com/Dao-AILab/flash-attention/issues/1915 (script calls both fwd and bwd, but the nvidia-cutlass-dsl upgrade only affects fwd).

Error goes away if I switch back to 4.2.0 in [pyproject.toml](https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/cute/pyproject.toml#L23)

tridao:
This should be fixed now

Issue #1915, opened 2025-09-29
Cute-DSL bwd R2P masking doesn't work on H200


imbr92:
I get the wrong result when running the causal cute-dsl FA bwd on an H200 unless I swap out the R2P trick part for the original way it was computed in [mask.py](https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/cute/mask.py#L97). 

Changes I had to make (to L97-110):
```py
# if cutlass.const_expr(False):
#     cute.printf("unreachable")
#     # traverse column index.
for c in cutlass.range(cute.size(tScS_mn.shape[1]), unroll_full=True):
    acc_S_mn[r, c] = -cutlass.Float32.inf if t0ScS_mn[0, c][1] >= col_limit_right else acc_S_mn[r, c]
# else:  # R2P trick, see apply_mask_sm100
#     col_limit_transformed = col_limit_right // 8 * 2 + min(col_limit_right % 8, 2)
#     ncol = cutlass.const_expr(cute.size(tScS_mn.shape[1]))
#     for s in cutlass.range_constexpr(cute.ceil_div(ncol, 24)):
#         col_limit_right_s = max(col_limit_transformed - s * 24, 0)
#         mask = (1 << col_limit_right_s) - 1
#         for i in cutlass.range_constexpr(min(24, ncol - s * 24)):
#             in_bound = cutlass.Boolean(mask & (1 << i))
#             c = s * 24 + i
#             acc_S_mn[r, c] = acc_S_mn[r, c] if in_bound else -cutlass.Float32.inf
```

Verification Code (only works for me once I make the change above^ in `mask.py`):

```py
import torch
import torch.nn.functional as F
from math import sqrt
from torch.nn.attention import sdpa_kernel, SDPBackend
from flash_attn.cute import flash_attn_func

def _sdpa_flash(q, k, v, causal: bool, scale: float):
    """
    Expects 4D tensors [B, H, L, D] and will force the FLASH_ATTENTION backend.
    Raises RuntimeError if FLASH is not available for the given dtype/hardware/shapes.
    """
    with sdpa_kernel(backends=[SDPBackend.FLASH_ATTENTION]):
        return F.scaled_dot_product_attention(
            q, k, v, is_causal=causal, scale=scale
        )

def torch_flash_fwd(q, k, v, softmax_scale=None, causal=True):
    """
    q: [B, Sq, H, D]
    k: [B, Sk, H_kv, D]
[...]

PR #1919, opened 2025-09-30
Minor changes to get cute-dsl bwd running (divisibility assumptions, returning correct number of args, etc.)

imbr92:
# Summary

Some minor changes to clean up + get cute-dsl bwd running. 
- Some of the copy atoms used in the 3 bwd kernels use # copy bits = 128 > dtype bits, so jit compiler complains without marking divisibility (to 8 for the 16 bit dtypes and 4 for the 32 bit dtypes)
- Backward returns the wrong number of arguments
- Added an assert to warp reduce for width (tripped me up when I was modifying copy bits constants since that ended up making width > warp size which wasn't caught by any asserts)
- Minor comment corrections

# Tests

Above changes are enough to get cute-dsl bwd running (no jit compiler errors), but note that I also needed to make the changes described in https://github.com/Dao-AILab/flash-attention/issues/1918 to get fwd running and correct, and https://github.com/Dao-AILab/flash-attention/issues/1915 for bwd correctness since the R2P masking trick was not working. 

After doing that, I get the cute-dsl bwd matching SDPA with `SDPBackend.FLASH_ATTENTION`. See testing script in https://github.com/Dao-AILab/flash-attention/issues/1915.

### DIFF
--- a/flash_attn/cute/flash_bwd.py
+++ b/flash_attn/cute/flash_bwd.py
@@ -819,7 +819,7 @@ class FlashAttentionBackwardSm80:
         else:
             tdVrP = mma_params.tdVrP
 
-        # MMA dK
+        # MMA dV
         sm80_utils.gemm(
             mma_params.thr_mma_dkv, mma_params.acc_dV, tdVrP, mma_params.tdVrdO,
             smem_copy_params.tdVsPt,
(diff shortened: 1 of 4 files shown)


imbr92:
This was fixed in ea03e0644c22a282d2ccd2b75844c76e4acb436b (divisibility added to the 3 bwd kernels)

PR #1934, opened 2025-10-13
feat: Adding varlen support to cute-dsl sm80 bwd

imbr92:
## Summary

Added varlen support for cute-dsl SM80 bwd. Split this into 3 commits:
1. Turning R2P trick off so that mask works for bwd (at least until https://github.com/Dao-AILab/flash-attention/issues/1915 is resolved)
2. Pre-indenting to make final commit more readable: Switching to using a tile scheduler requires me to check if a tile is valid before going on to the main kernel body. Since we are currently not able to exit early from cute-dsl kernels, I pre-indent with if True in this separate commit to make the main commit with actual changes more readable.
5. Actual changes to interface + bwd kernels to support varlen along with tests.

## Tests

Kept similar style pytest tests in flash-attention/tests/cute/test_varlen_bwd.py. Currently supports varlen with causal masking and gqa/mqa.

To test, note that we need to switch interface to use the sm90 main bwd and post-process bwd kernels here: 

https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/cute/interface.py#L430 

and 

https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/cute/interface.py#L449 

Test Output: 

image

### DIFF
--- a/flash_attn/cute/flash_bwd.py
+++ b/flash_attn/cute/flash_bwd.py
@@ -17,6 +17,7 @@ from flash_attn.cute import ampere_helpers as sm80_utils
 from flash_attn.cute import utils
 from flash_attn.cute.mask import AttentionMask
 from flash_attn.cute.seqlen_info import SeqlenInfoQK
+from flash_attn.cute.tile_scheduler import ParamsBase, SingleTileScheduler, SingleTileVarlenScheduler, TileSchedulerArguments
 
 
 class FlashAttentionBackwardSm80:
@@ -31,6 +32,7 @@ class FlashAttentionBackwardSm80:
         num_stages_Q: int = 2,
         num_stages_dO: int = 2,
         num_threads: int = 256,
+        pack_gqa: bool = False,
         is_causal: bool = False,
         SdP_swapAB: bool = False,
         dKV_swapAB: bool = False,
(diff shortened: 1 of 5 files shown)


flash_attn/cute/flash_bwd.py
```diff
@@ -245,11 +260,22 @@ def _setup_attributes(self):
         self.gmem_tiled_copy_dK = cute.make_tiled_copy_tv(atom_universal_copy, tQK_layout, vQKVdO_layout)
         self.gmem_tiled_copy_dV = cute.make_tiled_copy_tv(atom_universal_copy, tVdO_layout, vQKVdO_layout)
         async_copy_elems_accum = universal_copy_bits // cutlass.Float32.width
-        atom_async_copy_accum = cute.make_copy_atom(
-            cpasync.CopyG2SOp(cache_mode=cpasync.LoadCacheMode.GLOBAL),
-            cutlass.Float32,
-            num_bits_per_copy=universal_copy_bits,
-        )
+
+        # I think we wouldn't require this with smarter padding
```
tridao, line 264: I think vectorized copy should still work even in varlen since the contiguous dimension is the headdim (not the seqlen dim). But we can merge as is and fix later.

imbr92: Thanks for the quick review! I might be missing something, but from my understanding `atom_async_copy_accum` is only used for LSE which is padded like `[seqlen_0, m_block_size pad, seqlen_1, m_block_size pad, ...]` so we lose alignment when seq lens are not sufficiently divisible.

Though now that you mention it, it seems that the atom used for `gmem_tiled_copy_dQaccum` can probably be vectorized

PR #1936, opened 2025-10-14
Fix backward preprocess kernel in FlashAttention3

rfbr:
Fix backward preprocess kernel sentinel value for padded LSE positions.

The backward preprocess kernel used +INFINITY as the sentinel value for padded thread positions, which caused NaN/inf gradients in ring attention backward passes when these values were written to the lse log2 buffer. This commit changes the sentinel value from +INFINITY to -INFINITY for padded positions which ensures padded positions write 0.f to the buffer, preventing NaN/inf propagation in ring attention gradient computations.

### DIFF
--- a/hopper/flash_bwd_preprocess_kernel.h
+++ b/hopper/flash_bwd_preprocess_kernel.h
@@ -164,7 +164,7 @@ public:
         Tensor mLSE = make_tensor(make_gmem_ptr(params.ptr_LSE), shape_LSE, params.stride_LSE)(_, bidh, !is_varlen ? bidb : 0);
         Tensor gLSE = local_tile(cute::domain_offset(make_coord(seqlen_info.offset), mLSE), Shape<Int<kBlockM>>{}, make_coord(m_block));
         static_assert(kBlockM <= MaxThreadsPerBlock);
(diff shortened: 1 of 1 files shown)


tridao:
The +inf value is intentional, so that during the backward kernel, for rows that are out of bound, we can just subtract from lse (+inf), then exponentiate, to get zero. So we can avoid bound checking in the seqlen_q dimension.

If you want to change it for ring attn it should be a flag.

rfbr:
First of all, thank you for taking the time to look at my PR and for your work! I'm new to the repo and now understand better how the +inf is used in the mainloop backward pass, which I missed at first glance. However I'm curious about the choice of sentinel value, couldn't the same approach used in flash_bwd_preprocess_kernel (checking for -inf) be applied in the mainloop backward as well to set scores to 0?

tridao:
Sure, but we'd like to do as little extra work in the main backward kernel as possible, for performance reason. You'd see that in the backward mainloop we don't scale LSE by log2 (done in preprocess) and don't do bound checking on LSE (preprocess kernel pads LSE to multiples of `m_block_size`).

rfbr:
I understand, thank you for clarifying!

PR #1944, opened 2025-10-17
[ROCm] prepare CK sources for pytorch hipify v2 APIs

jeffdaily:
See https://github.com/pytorch/pytorch/pull/151845.
pytorch has removed caffe2, but hipify still contained work-arounds for caffe2 vs torch compatibility.
As a result of hipify v2 changes, some torch APIs are changing.


tridao:
Lgtm, we can merge when you think it’s ready

jeffdaily:
It’s ready.  Build tested in both v1 and v2 environments.

Issue #879, opened 2024-03-07
Has anyone successfully compiled this on ARM Linux (aarch64)?


elkay:
I want to use this on a newer AWS Graviton server but anything I've done to try to get it to compile from sources has completely hung the process until I abort. No error output. I'm used to longer compile times and don't believe this is the issue. CPU usage hovers around 1% so it's not doing anything. Not even sure how to go about debugging it.

elkay:
Actually, just noticed this PR.

https://github.com/Dao-AILab/flash-attention/pull/757

Seems like exactly what I'm looking for. Hopefully the PR can be approved.

tridao:
Does either #757 or #724 work for you?

elkay:
> Does either #757 or #724 work for you?

757 did end up working for me.

anubrata:
I’m unable to install this package on an aarch64 Slurm server with H100 GPUs using #757. I keep encountering wheel not found errors despite trying different version numbers.

I also attempted to compile from source following the Hopper GPU instructions but ran into compilation failures.

Any suggestions on how to work around this?

leo-liuzy:
I am running build from source on GH200 with aarch64 CPU. 
```
torch==2.5.1+12.4
triton==3.1.0
```
I tried using setup.py both in `hopper` and in root directory. The error seems to be from ninja. I wonder if @tridao have some guess on what the problem is.

# Hopper

command: `MAX_JOBS=72 python setup.py install`
The error occurs at 
```
[...]

PR #1949, opened 2025-10-21
Fix hopper cuda 13 build

kevmo314:
Updates the hopper build to use system CCCL headers instead of downloading the toolchain.

Without this patch, I get errors in the build about CUB headers being missing. From what I can tell, it seems like the toolchain download is incompatible with cuda 13? Not completely sure, but this implements the fix described in https://github.com/Dao-AILab/flash-attention/issues/879#issuecomment-3410395783.


tridao (approved):

tridao:
Thank you!

Issue #1444, opened 2025-01-14
Wheel names and version inconsitency.


sfc-gh-mhazy:
Hi,
Thank you for uploading wheels in the recent release.

**Overview** 

I was experimenting with them a bit and I think they have a slight inconsistency between wheel name and version in metadata.

Based on https://peps.python.org/pep-0491/#file-name-convention the wheel name is  `{distribution}-{version}(-{build tag})?-{python tag}-{abi tag}-{platform tag}.whl.`

where IIUC, version is the one that is in `METADATA` file. 

Sample wheel from `v2.7.3` release: `flash_attn-2.7.3+cu11torch2.1cxx11abiFALSE-cp310-cp310-linux_x86_64.whl`, so version is `2.7.3+cu11torch2.1cxx11abiFALSE`. However, version `METADATA`  is `2.7.3`

**Example problem** 

Lets assume:
*  wheels are uploaded to some pip repository
*  One has a requirements file
```
torch==2.5.1+cu118
flash_attn=2.7.3+cu11torch2.5cxx11abiFALSE
```
* we point pip to appropriate repositories (https://download.pytorch.org/whl/cu118) 

The resolution of flash_attn fails. 

If we change the constraint to `flash_attn=2.7.3` it will succesfully download wheel. 

**Proposal**

I think the fix would be to change the versioning in https://github.com/Dao-AILab/flash-attention/blob/main/setup.py . Happy to make the change if this sounds reasonable.

To add some justification:

If you check pytorch wheels (example https://download.pytorch.org/whl/torch/), they follow the convention:
`torch-2.0.0+cu118-cp310-cp310-linux_x86_64.whl` has version in `METADATA` set to `2.0.0+cu118`

tridao:
What should the versioning in `setup.py` look like?

ZiyaCu:
@tridao I have the same problem. Please fix the metadata so we can use the wheels. Thanks.

PR #1956, opened 2025-10-22
Make wheel name and version consistent

bobingm:
Fix the version according to this https://github.com/Dao-AILab/flash-attention/issues/1444


bobingm:
@tridao

PR #1970, opened 2025-10-28
BlockSparse Tweaks

drisspg:
# Summary

Enables: https://github.com/pytorch/pytorch/pull/166359

What does this do:
* Adds better error messages / expand -> I will not tell you how long it took me to figure out why there was an IMA lol
* In Flex integration for score_mods we standardized on SSA form it makes it reallly nice to codegen made the same change for mask_mods
* I remove the seqlen vars to align w/ flex impl. I went back and fourth on this because at least for the offset based mods this means we have to compile + seqlen instantation. With makes test stupidly slow so idk either we change the mask mods or we revert this and leave as dead params for PT integration.
* For Triton impl -> we basically let dyanmic shapes handles this and we update the kernel siganture with additional symbols.
I think that we will ultmately need something like 
`aux_tensors`
`aux_ints`
`aux_floats`

Which act as containers that get unpacked in the score and mask mods. For now I am just baking in these in

### DIFF
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -42,7 +42,11 @@ from flash_attn.cute.flash_bwd_sm100 import FlashAttentionBackwardSm100
 from flash_attn.cute.flash_bwd_postprocess import FlashAttentionBackwardPostprocess
 from flash_attn.cute.flash_fwd_combine import FlashAttentionForwardCombine
 
-from flash_attn.cute.block_sparsity import BlockSparseTensorsTorch, to_cute_block_sparse_tensors
+from flash_attn.cute.block_sparsity import (
+    BlockSparseTensorsTorch,
+    to_cute_block_sparse_tensors,
+    validate_block_sparse_tensors,
+)
 
 
 def maybe_contiguous(x):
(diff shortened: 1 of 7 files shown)


flash_attn/cute/mask.py
```diff
@@ -135,17 +135,23 @@ def apply_mask(
                     # Convert to absolute column index
                     global_col_idx = thr_col_offset + col_idx_local + n_block * self.tile_n
 
-                    cond = cutlass.Boolean(
-                        mask_mod(
-                            batch_idx,
-                            head_idx,
-                            tScS_mn[r, 0][0] + m_block * self.tile_m,
-                            thr_col_offset + t0ScS_mn[0, col][1] + n_block * self.tile_n,
-                            self.seqlen_q,
```
drisspg, line 144: Semantic change, I remove seqlen_q and seqlen_kv and instead switch to the flex mask_mod api

flash_attn/cute/interface.py
```diff
@@ -348,7 +361,7 @@ def _flash_attn_fwd(
         score_mod_hash,
         mask_mod_hash,
         use_block_sparsity,
-        aux_tensors is not None,
+        aux_tensor_shapes,
```
drisspg: This really shouldn't be necessary IMO and I dont really get this. But for the document mask case if we reuse the mask_mod (which we should be able to) we get bad numerics and wrong outputs..

drisspg: probs cause we didnt mark dynamic ..

drisspg: this is the way

tridao (approved):

PR #1980, opened 2025-11-02
Add blocksparse support to sm100 and try to put blocksparse stuff in central file 

drisspg:
# Almost ready

need to add aux_tensor wrapping

### DIFF
--- a/flash_attn/cute/flash_fwd.py
+++ b/flash_attn/cute/flash_fwd.py
@@ -2114,22 +2114,29 @@ class FlashAttentionForwardSm90(FlashAttentionForwardBase):
                 # ==========================================
                 # Block sparsity
                 # ==========================================
-                kv_consumer_state, O_should_accumulate = self.process_blocksparse_mainloop(
-                    kv_consumer_state=kv_consumer_state,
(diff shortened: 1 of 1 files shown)


flash_attn/cute/block_sparsity.py
```diff
@@ -151,6 +154,635 @@ def to_cute_block_sparse_tensors(tensors: BlockSparseTensorsTorch) -> Optional[B
...
+    kv_producer_state,
+    load_Q,
+    load_K,
+    load_V,
+    pipeline_kv,
+):
+    """SM100 version of process_block_list (no intra_wg_overlap, no extra_tx_count)."""
+    if block_count > 0:
+        # First iteration: load Q alongside K if requested
+        n_block_first = block_indices[block_count - 1]
```
drisspg: Todo, I need to figure this one out, since each cta handles 2 slices of query, in the base case w/ block_sparse_q = 128 and stages_2 we need to actually load the block_indices incremented 1 row down. 

I think we can jump  by 2*blockM sizes as we iterate chunks I just need to double check how we do the predicates and handle not reading past the len(block-indices) 


cc @reubenconducts if you want to sanity check me here

PR #1982, opened 2025-11-03
[Cute] Fix main

drisspg:
Fixes main

we were always checking blocksparse and not None


jduprat:
This will fix currently broken benchmarking over at [pytorch-integration-testing](https://github.com/pytorch/pytorch-integration-testing)

tridao (approved):

PR #1985, opened 2025-11-05
[Cute] Block sparse support Sm100

drisspg:
# Summary
- Implement block-sparse attention in flash_fwd_sm100.py
- Update interface.py to handle SM100 block size calculations
  (2x multiplier for m_block_size since 1 CTA handles 2*tile_m rows)
- Add mask_mod parameter support in mask.py for block-sparse masking
- Add SM100 test fixtures and tile size handling in test_mask_mod.py

## Fast follow
Do the aux_tensor fastdivmod wrapping to avoid OOB reads

Also we should land:
https://github.com/Dao-AILab/flash-attention/pull/1984
Before and rebase so its easier to review

## Perf
Alot of perf wins (not universal for document mask ) but the delta from sol is much higher than what was found on hopper impl

Not autotuning the flex blocksparse impl gives this:
combined_comparison

And autotuning the triton impl:
combined_comparison_autotune

### Possible problems

Looking at the Pm samples we can see a long tail:
<link image>
For causal_mask with the default StaticPersistentSchedule. (We need to build a generic version of this) but we already have a better schedule for causal. If hard code  the LPT schedule 

we go from :
Screenshot 2025-11-04 at 5 39 33 PM

to:

Screenshot 2025-11-04 at 5 39 14 PM

## Tests
Screenshot 2025-11-10 at 8 26 43 PM
Screenshot 2025-11-10 at 8 37 51 PM
Screenshot 2025-11-10 at 8 43 15 PM

### DIFF
--- a/flash_attn/cute/block_sparse_utils.py
+++ b/flash_attn/cute/block_sparse_utils.py
@@ -7,12 +7,14 @@ These utilities are used by CUTE DSL kernels to produce and consume block-sparse
 
 from typing import Callable
 from functools import partial
+import math
 import cutlass
 import cutlass.cute as cute
-from cutlass import const_expr
+from cutlass import Float32, Int32, const_expr
 
 # Import data structures from block_sparsity
 from flash_attn.cute.block_sparsity import BlockSparseTensors
+from flash_attn.cute import utils
 
 
 @cute.jit
@@ -143,8 +145,13 @@ def produce_block_sparse_loads(
 
     curr_mask_block_cnt = mask_block_cnt[batch_idx, head_idx, m_block]
     curr_mask_block_idx = mask_block_idx[batch_idx, head_idx, m_block, None]
-    curr_full_block_cnt = full_block_cnt[batch_idx, head_idx, m_block]
-    curr_full_block_idx = full_block_idx[batch_idx, head_idx, m_block, None]
+
+    if const_expr(full_block_cnt is not None):
+        curr_full_block_cnt = full_block_cnt[batch_idx, head_idx, m_block]
+        curr_full_block_idx = full_block_idx[batch_idx, head_idx, m_block, None]
+    else:
+        curr_full_block_cnt = Int32(0)
+        curr_full_block_idx = None
 
     mask_empty = curr_mask_block_cnt == 0
     full_empty = curr_full_block_cnt == 0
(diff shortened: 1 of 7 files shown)


flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -870,6 +882,9 @@ def kernel(
         )
         TileSchedulerCls = partial(self.tile_scheduler_cls.create, tile_sched_params)
 
+        # Constexpr pipeline flags for block sparsity (SM90 style)
```
drisspg: delete

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1125,30 +1145,48 @@ def load(
...
-            load_K(block=n_block_max - 1, producer_state=kv_producer_state, page_idx=page_idx)  # K0
-            kv_producer_state.advance()
-            if const_expr(self.q_stage == 2):
-                load_Q(block=self.q_stage * m_block + 1, stage=1)  # Q1
-            q_producer_phase ^= 1
-            load_V(block=n_block_max - 1, producer_state=kv_producer_state, page_idx=page_idx)  # V0
-            kv_producer_state.advance()
-            for i in cutlass.range(n_block_max - 1 - n_block_min, unroll=1):
-                n_block = n_block_max - 2 - i
+            if const_expr(self.use_block_sparsity):
```
drisspg: I hate that github renders it this way, ill change to ` not use` so its easier to review

flash_attn/cute/block_sparse_utils.py
```diff
@@ -417,3 +417,344 @@ def consume_block_sparse_loads(
...
+
+    return (
+        softmax_corr_consumer_phase,
+        o_corr_consumer_phase,
+        corr_epi_producer_phase,
+    )
+
+
+@cute.jit
+def consume_block_sparse_loads_sm100(
```
reubenconducts: I find this name a bit confusing, since "consume" is ambiguous -- my $0.02 would be to have `softmax_block_sparse_sm100` and `load_block_sparse_sm100`, though that leads to a bit of an overlap with `load_block_list_sm100`. Very minor nitpick, though

drisspg: great point, will update

flash_attn/cute/block_sparse_utils.py
```diff
@@ -417,3 +422,349 @@ def consume_block_sparse_loads(
...
+        return (
+            mask_block_cnt[batch_idx, head_idx, m_block]
+            + full_block_cnt[batch_idx, head_idx, m_block]
+        )
+    else:
+        return mask_block_cnt[batch_idx, head_idx, m_block]
+
+
+@cute.jit
+def handle_block_sparse_empty_tile_correction_sm100(
```
drisspg, line 578: had to dupe alot here but still think its better, having very large IF Else indents makes it harder to rebase / iterate on all the constexpr tree

tridao (approved):

drisspg:
@tridao Okay, finally rebased, perf looks good 
and tests are green
Screenshot 2025-11-13 at 7 26 18 PM

drisspg:
Screenshot 2025-11-17 at 4 28 12 PM

tridao (approved):

PR #1992, opened 2025-11-06
[Cute,Fwd,Sm90] Support paged attention

imbr92:
### Summary

Adding support for KV cache in sm90 fwd. Requires page size to be a multiple of `n_block_size`

### Testing

I updated the kv cache test to enforce `seq_len_q <= seq_len_k` and ran `tests/cute/test_flash_attn.py::test_flash_attn_kvcache` on an H200. 

Test Output:
image

### DIFF
--- a/flash_attn/cute/flash_fwd.py
+++ b/flash_attn/cute/flash_fwd.py
@@ -1880,7 +1880,6 @@ class FlashAttentionForwardSm90(FlashAttentionForwardBase):
                 work_tile = tile_scheduler.get_current_work()
                 # End of persistent scheduler loop
 
-    # How to inline?
     @cute.jit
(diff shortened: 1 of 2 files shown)


tridao:
Is `self.tiles_per_page` a compile time constant? 
If so, we should add it to the `compile_key` in the interface

imbr92:
> Is `self.tiles_per_page` a compile time constant? If so, we should add it to the `compile_key` in the interface

Yes, it's a compile time constant. I left it out of the `compile_key` since it is completely determined by `n_block_size` and `page_size` which are both already in the `compile_key`. Should I instead include all three in the `compile_key` and pass `tiles_per_page` directly to the constructor?

Issue #2005, opened 2025-11-12
[Cute] Splitkv PR cut perf  by 2/3s


drisspg:
# Summary

Context although it says flex-attention, I was rerunning  perf scripts to re-verify numbers for this PR: https://github.com/Dao-AILab/flash-attention/pull/1985. I noticed the Noop which doesn't pass in  any sparse tensor info took a big hit. Bisected it to: https://github.com/Dao-AILab/flash-attention/commit/e724e2588cbe754beb97cf7c011b5e7e34119e62

cc @timmy-feng, @tridao 
Looks like we might need some more constexpr paths 

#### Current main:
```Shell
❯ python benchmarks/flex_perf.py --config benchmarks/configs/config_basic.yaml
100%|████████████████████████████████████████████████████████████████████████████████████| 16/16 [00:12<00:00,  1.30it/s]
| attn_type   | dtype          | shape(B,Hq,M,Hkv,N,D)          |   flex_attn_fwd_time |   flex_attn_fwd_TFlops |   flex_attn_fwd_mem_bw (TB/s) |
|-------------|----------------|--------------------------------|----------------------|------------------------|-------------------------------|
| noop        | torch.bfloat16 | (2, 16, 4096, 16, 4096, 128)   |              355.296 |                776.681 |                         0.378 |
| noop        | torch.bfloat16 | (2, 16, 8192, 16, 8192, 128)   |             1207.264 |                914.304 |                         0.222 |
| noop        | torch.bfloat16 | (2, 16, 16384, 16, 16384, 128) |             4781.056 |                923.484 |                         0.112 |
| noop        | torch.bfloat16 | (2, 16, 32768, 16, 32768, 128) |            19089.409 |                925.168 |                         0.056 |
| noop        | torch.bfloat16 | (2, 32, 4096, 32, 4096, 128)   |              612.512 |                901.049 |                         0.438 |
| noop        | torch.bfloat16 | (2, 32, 8192, 32, 8192, 128)   |             2404.416 |                918.149 |                         0.223 |
| noop        | torch.bfloat16 | (2, 32, 16384, 32, 16384, 128) |             9566.176 |                923.091 |                         0.112 |
| noop        | torch.bfloat16 | (2, 32, 32768, 32, 32768, 128) |            38171.646 |                925.342 |                         0.056 |
| noop        | torch.bfloat16 | (4, 16, 4096, 16, 4096, 128)   |              612.320 |                901.331 |                         0.438 |
[...]

PR #2006, opened 2025-11-12
[Cute,Fwd,Sm100] fix major regression with split kv

jayhshah:
Hotfix for #2005


tridao (approved):

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1162,7 +1162,7 @@ def load(
 
             n_block_min, n_block_max = block_info.get_n_block_min_max(seqlen, m_block, split_idx, num_splits)
 
-            if n_block_min < n_block_max:
+            if const_expr(not self.is_split_kv) or n_block_min < n_block_max:
```
Edenzzzz, line 1165: Hi can I ask why this is a regression? e.g. for the 1st query tile, this will load most of the subsequent kv tiles  (instead of skipping) only to be masked out later? Thanks

PR #1996, opened 2025-11-08
[NVIDIA] bump github actions

johnnynunez:
- Add support for python 3.14
- Put official 2.9.1 pytorch version(skipping 2.9.0 that is bugged for some nvidia devices).

PR #2007, opened 2025-11-12
add support for torch 2.9.0 and deprecate 3.9

Jackmin801:
torch 2.9.0 makes min supported python version 3.10 and thus 3.9 is dropped from the matrix


Jackmin801:
Testing that the build and deploy action will work in a fork here: https://github.com/Jackmin801/flash-attention/actions/runs/19301136888

It seems some of the matrix elements dont build. Will look into it further

johnnynunez:
> Testing that the build and deploy action will work in a fork here: https://github.com/Jackmin801/flash-attention/actions/runs/19301136888
> 
> It seems some of the matrix elements dont build. Will look into it further

https://github.com/Dao-AILab/flash-attention/pull/1996

i recommend you to jump to 2.9.1 and triton 3.5.1, 2.9.0 and 3.5.0 are unstable

PR #2014, opened 2025-11-15
[Cute,Sm100,Fwd] use correction warps for epi when not using TMA

jayhshah:
This PR switches the sm100 fwd epilogue over to using the 4 correction warps when not using TMA store (e.g., when query seqlen is variable), instead of adding the empty warp. This also enables Paged KV in general with varlen since we eliminate the contention issue between producer and epilogue warps.

Will look at perf before merging.

### DIFF
--- a/tests/cute/test_flash_attn.py
+++ b/tests/cute/test_flash_attn.py
@@ -100,8 +100,8 @@ def test_flash_attn_output(
     mha_type,
     dtype,
 ):
-    if (causal or local) and seqlen_k < seqlen_q:
-        pytest.skip("Causal attention requires seqlen_k >= seqlen_q")
+    # if (causal or local) and seqlen_k < seqlen_q:
+    #     pytest.skip("Causal attention requires seqlen_k >= seqlen_q")
     device = "cuda"
     # set seed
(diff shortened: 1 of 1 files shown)


tridao:
thanks, we can merge when you think it's ready

jayhshah:
I don't see any perf regression, indeed using the 4 correction warps for varlen epilogue over 2 shows a very minor gain, e.g.

```
NEW

### headdim = 128, causal = False, seqlen = 4096, batch_size = 4, nheads = 16 ###
FA Python fwd: 0.396ms, 1387.7 TFLOPS

### headdim = 128, causal = True, seqlen = 4096, batch_size = 4, nheads = 16 ###
FA Python fwd: 0.230ms, 1194.9 TFLOPS

OLD

### headdim = 128, causal = False, seqlen = 4096, batch_size = 4, nheads = 16 ###
FA Python fwd: 0.398ms, 1382.5 TFLOPS

### headdim = 128, causal = True, seqlen = 4096, batch_size = 4, nheads = 16 ###
FA Python fwd: 0.233ms, 1180.6 TFLOPS
```

PR #2017, opened 2025-11-16
[Cute,Fwd/Bwd,Sm12x] [WIP/DRAFT/HELP] cute FA for sm12x

johnnynunez:
Baseline: cutlass Cutlass v4.3.0 is running on spark and consumer devices
following image is an error, so we are fixing it in upcoming releases.
image

PR summary(trying to implement cute dsl thanks to your structure code).
this PR can contain a lot of mistakes ( i am beginner level trying to understand tcgen05 and cuda technologies). It can be a baseline for expert people.

sm11x/sm12x:
SM120 is essentially a stripped-down version of SM100: it lacks Tensor Memory, and UTCMMA instructions, and only supports the smaller-tile “SuperMMA” operations.

cc @drisspg @tridao

### DIFF
--- a/flash_attn/cute/flash_fwd_sm100.py
+++ b/flash_attn/cute/flash_fwd_sm100.py
@@ -2394,8 +2394,6 @@ class FlashAttentionForwardSm100:
             tOcO = gmem_thr_copy_O.partition_S(cO)
             t0OcO = gmem_tiled_copy_O.get_slice(0).partition_S(cO)
             tOpO = utils.predicate_k(tOcO, limit=mO_cur.shape[1])
-            # TODO: the packgqa case isn't correct rn (sometimes IMA), disabling it
-            assert not self.pack_gqa
             pack_gqa = PackGQA(
                 self.m_block_size,
                 self.head_dim_v_padded,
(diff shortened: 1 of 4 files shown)


johnnynunez:
@tridao @drisspg is tests from cute updated? i've got error for num_splits
image

johnnynunez:
seems that tests_varlen is compatible with thor
💡 Suggestions:
 Ensure env CUTE_DSL_ARCH matches your GPU architecture
FAILED tests/cute/test_flash_attn_varlen.py::test_varlen[mha-dtype1-None-True-8-1-128-1-20] - cutlass.cute.nvgpu.common.OpError: OpError: expects arch to be one of [Arch.sm_100a, Arch.sm_100f, Arch.sm_101a, Arch.sm_101f, Arch.sm_110a, Arch.sm_110f], but got Arch.sm_121a
Error Code: MmaF16BF16Op error

johnnynunez:
@drisspg @tridao 
image

flash_attn/cute/flash_bwd_postprocess.py
```diff
@@ -47,8 +47,8 @@ def __init__(
         """
         self.dtype = dtype
         self.tile_m = tile_m
-        assert arch in [80, 90, 100], (
-            "Only Ampere (80), Hopper (90), and Blackwell (100) are supported"
+        assert arch in [80, 90, 100, 110], (
+            "Only Ampere (80), Hopper (90), Blackwell (100), and Thor (110) are supported"
```
thakkarV:
```suggestion
            "Only Ampere (80), Hopper (90), Blackwell Datacenter (100), and Thor (110) are supported"
```

johnnynunez:
@thakkarV thanks! i have to refactor sm12 because i was thinking that i could use tcgen05, but only sm10x and sm11x can...
dgx spark/rtx50 is more similar to the GeForce Ampere style MMAs

reubenconducts:
> @drisspg @tridao image

@johnnynunez This was a silly limitation in this test -- PyTorch's `BlockMask` counts fully-computed blocks that go out of bounds as masked, whereas we do not in FlashAttention (seqlen masking is handled independent of `mask_mod` application). So the full block count is potentially one less and the mask block count is potentially one more than expected in the tests where block size does not evenly divide sequence length.

PR #2020, opened 2025-11-19
add fastdivmod for oob reads in mask_mods

drisspg:
# Summary

Follow up PR , replays same logic as score_mod

If I manually set this

```Shell
                if False or const_expr(wrap_aux_indices):
                    _, global_col_for_mod = fastdiv_mods[1].divmod(global_col)
```
Screenshot 2025-11-18 at 7 51 33 PM

## With this PR I get:
Screenshot 2025-11-18 at 7 49 25 PM

### DIFF
--- a/flash_attn/cute/block_sparse_utils.py
+++ b/flash_attn/cute/block_sparse_utils.py
@@ -283,6 +283,7 @@ def consume_block_sparse_loads(
     score_mod_fn,
     O_should_accumulate,
     mask_mod,
+    fastdiv_mods,
     intra_wg_overlap: cutlass.Constexpr,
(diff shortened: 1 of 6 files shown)


drisspg:
cc @reubenconducts Should be easy to apply to sm90

drisspg:
@tridao easy one

tridao (approved):

tridao:
Btw cutlass 4.3 (out today) also has FastDivmod now with a more pythonic syntax, we can use it when we bump version to 4.3

tridao:
we can merge when you think it's ready

PR #2025, opened 2025-11-22
Bump pin

drisspg:
# Summary
* *Bump Cutlass DSL pin to 4.3.0 -> needed the fill_swizzled to store to constant val to which is now asserted with any swizzle layout in 4.3.0
* Small tweak to the benchmark to lets us run with only cute installed, perf looks good before and after
* Varlen test now skips SM100 (Blackwell) backward and runs fwd 
* Swapped custom FastDivmod for Cutlass’s FastDivmodDivisor, I had to add some `divisor` args directly since the main diff besides just calling the builtin in `divmod` is no divisor attribute to grab easily

### DIFF
--- a/flash_attn/cute/flash_bwd_sm100.py
+++ b/flash_attn/cute/flash_bwd_sm100.py
@@ -414,8 +414,7 @@ class FlashAttentionBackwardSm100:
             assert mdK_semaphore is not None
             assert mdV_semaphore is not None
             mdK_semaphore, mdV_semaphore = [
-                utils.select(t, mode=semaphore_transpose)
-                for t in (mdK_semaphore, mdV_semaphore)
+                utils.select(t, mode=semaphore_transpose) for t in (mdK_semaphore, mdV_semaphore)
             ]
         else:
             mdK_semaphore = None
@@ -562,7 +561,7 @@ class FlashAttentionBackwardSm100:
             cute.size(mQ.shape[2]),  # num_heads = num_query_heads
             cute.size(mK.shape[3]),
             1,  # num_splits
-            cute.size(mQ.shape[0]), # pass seqlen_q for seqlen_k
+            cute.size(mQ.shape[0]),  # pass seqlen_q for seqlen_k
             mQ.shape[1],
             mV.shape[1],
             total_q=cute.size(mQ.shape[0]),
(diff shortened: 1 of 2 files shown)


flash_attn/cute/paged_kv.py
```diff
@@ -173,4 +173,13 @@ def load_KV(self, n_block: Int32, sX: cute.Tensor, K_or_V: str):
                     )
             elif const_expr(K_or_V == "V"):
                 # Don't need to clear out the rest of the smem for K since we'll mask out the scores anyway.
-                tXsX[None, m, None].fill(0)
+                fill_tensor_swizzled(tXsX[None, m, None], 0)
+
+
+def fill_tensor_swizzled(tensor, value=0):
```
drisspg: from @fengxie

```py
@cutlass.dsl_user_op
@cute.jit
def fill_swizzled(tensor, value: cutlass.Numeric, *, loc=None, ip=None) -> None:
    # Naively process all elements one-by-one
    for _ in range(cute.size(tensor.shape, loc=loc, ip=ip)):
        ptr = _cute_ir.apply_swizzle(tensor.iterator.value, loc=loc, ip=ip)
        # 1-element tensor
        tmp = cute.make_tensor(
            cute.recast_ptr(ptr, swizzle_=None, loc=loc, ip=ip), (1,), loc=loc, ip=ip
        )
        tmp[0] = value
```

is better but may be a perf hit

fengxie: Thanks fore reporting. This one probably better for perf. 

```python
@cutlass.dsl_user_op
def fill_swizzled(tensor, value: cutlass.Numeric, *, loc=None, ip=None) -> None:
    """Fill tensor with a constant value.

    Fills all elements of the tensor with the specified value, assuming static size
    and supported memory space.
    """
    rTmp = cute.make_rmem_tensor_like(tensor, tensor.element_type)
    rTmp.fill(value)
    cute.autovec_copy(rTmp, tensor)
```

drisspg: updated

flash_attn/cute/flash_bwd_sm100.py
```diff
@@ -562,7 +561,7 @@ def __call__(
             cute.size(mQ.shape[2]),  # num_heads = num_query_heads
             cute.size(mK.shape[3]),
             1,  # num_splits
-            cute.size(mQ.shape[0]), # pass seqlen_q for seqlen_k
+            cute.size(mQ.shape[0]),  # pass seqlen_q for seqlen_k
```
drisspg, line 564: this is all ruff formatting

flash_attn/cute/tile_scheduler.py
```diff
@@ -284,11 +287,13 @@ def create(
             num_hb_remainder = (args.num_head * args.num_batch) % swizzle
             return SingleTileLPTScheduler.Params(
                 total_blocks=args.num_block * args.num_head * args.num_batch,
-                num_block_divmod=FastDivmod.create(args.num_block),
-                num_head_divmod=FastDivmod.create(args.num_head),
-                l2_minor_divmod=FastDivmod.create(swizzle),
-                l2_major_divmod=FastDivmod.create(swizzle * args.num_block),
-                l2_minor_residual_divmod=FastDivmod.create(
+                num_block=args.num_block,
```
drisspg, line 290: no public `divisor` attr on the divmods

tridao (approved):

PR #2027, opened 2025-11-22
[Cute,Fwd] Add scheduler metadata kernel for varlen and dynamic splits into sm100 fwd

jayhshah:
We add initial support for precomputing scheduler metadata, following the FA3 C++ implementation. Currently, we compute dynamic splits per batch for varlen q and/or k (e.g., when passing in `cache_seqlens`). `num_splits` itself then plays the role of a static allocation for the partial output buffers, capping the allowable number of dynamic splits. We enable the associated tests for scheduler metadata in `test_flash_attn_kvcache`, which all validate.

We observe no perf regression versus main. This initial implementation builds in use of dynamic splits into all of the forward tile schedulers. Note that this doesn't yet include the varlen persistent tile scheduler; we plan to add this in a future PR.

We also only added for sm100 but plan to extend support to sm90 and 8x kernels later.

Work done by @reubenconducts with a little cleanup from myself.

### DIFF
--- a/tests/cute/test_flash_attn.py
+++ b/tests/cute/test_flash_attn.py
@@ -776,6 +776,7 @@ def test_flash_attn_varlen_output(
         # (128, 128),
         # (256, 512),  # To test appending KV with more than 1 block
         # (2048, 3577),  # Enough tile to test persistent scheduler
+        (1024, 1024),
     ],
 )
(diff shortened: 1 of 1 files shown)


jayhshah:
Moving this to a new branch.

PR #2034, opened 2025-11-25
Disable abi3 for free-threaded python

kevmo314:


hopper/setup.py
```diff
@@ -67,6 +67,12 @@
 DISABLE_HDIMDIFF64 = os.getenv("FLASH_ATTENTION_DISABLE_HDIMDIFF64", "FALSE") == "TRUE"
 DISABLE_HDIMDIFF192 = os.getenv("FLASH_ATTENTION_DISABLE_HDIMDIFF192", "FALSE") == "TRUE"
 
+# Auto-disable the Python limited API (abi3) for free-threaded Python (3.13t+)
+# Free-threaded Python does not support the limited API
+# See: https://docs.python.org/3/howto/free-threading-extensions.html
+FREE_THREADED = sysconfig.get_config_var("Py_GIL_DISABLED")
+DISABLE_ABI3 = FREE_THREADED or os.getenv("FLASH_ATTENTION_DISABLE_ABI3", "FALSE") == "TRUE"
```
rgommers, line 74: Adding the extra environment variable here is potentially useful, but not related to free-threading support. I'd personally not include it unless a maintainer asks for it.

janeyx99:
@rgommers for context, xformers maintainers cc @danthe3rd wanted to be able to build a single wheel (since FA3 didn't publish wheels so the onus is on the users to produce the wheels) https://github.com/Dao-AILab/flash-attention/pull/1662.

rgommers:
@janeyx99 sure that is perfectly fine. The point of this PR isn't to change what wheels are distributed nor what wheels are even built for a default (with-GIL) Python interpreter. However, free-threaded interpreters don't support the stable ABI at all yet, and once they do (expected for Python 3.15, see PEP 803) then that will be a new ABI rather than `abi3`. So `pip install .` right now will produce a hard error under a free-threaded interpreter like:
```
#error "The limited API is not currently supported in the free-threaded build"
```
The goal of this PR is to fix that, while changing nothing for with-GIL interpreteres.

This vLLM PR has a more extensive description to explain: https://github.com/vllm-project/vllm/pull/29241.

It'd have been useful for the PR description to articulate this. Note that I just happened to notice this PR and decided to add some context that I thought may be useful, I'm not otherwise involved. @kevmo314 may be able to add more context (maybe they want to experiment with free-threading and this package?).

PR #2035, opened 2025-11-26
[Cute] Fix head dim 64 bwd

drisspg:
# Summary

Adds support for headdim64, a myriad of small fixes:
1. https://github.com/Dao-AILab/flash-attention/pull/2035/files#diff-ef5e5c95ec953948e93c912b7d92841bf839138364dc065c4d567bcc8bd7a3c5R326 we had a divison that made the num_stages -> 0 
2. https://github.com/Dao-AILab/flash-attention/pull/2035/files#diff-ef5e5c95ec953948e93c912b7d92841bf839138364dc065c4d567bcc8bd7a3c5R527 we were using the wrong mma_tiler for the dV matmul that happend to work because tile sizes were equal
3. https://github.com/Dao-AILab/flash-attention/pull/2035/files#diff-ef5e5c95ec953948e93c912b7d92841bf839138364dc065c4d567bcc8bd7a3c5R606-R612; This one was crazy. @v0i0 and I spent 4 hours debugging this and tried many many things. Eventually narrowed down to a race condition in epilogue dK that would cause the smem tensor to overlap with what was needed for sdS and this made dQ have bad values but dk was good since it was not using shared memory. 

A really helpful technique was basically to permute the sdO allocation to the end of the struct and this made racey overlap cause an OOB read/write and hard error in compute sanitizer. Basically in the non MHA path we were allocating in fp16 and we should have been using fp32. TBH not totally sure why this hasn't showed up before, could be different timing / overlap w/ headdim =128.

### DIFF
--- a/flash_attn/cute/flash_bwd_sm100.py
+++ b/flash_attn/cute/flash_bwd_sm100.py
@@ -320,11 +320,11 @@ class FlashAttentionBackwardSm100:
         )
         self.sdKV_epi_tile = (
             self.tile_n,
-            128 // (self.dk_dtype.width // 8),  # 64 or 32
+            min(128 // (self.dk_dtype.width // 8), self.tile_hdim // 2),  # 64 or 32
         )  # subtiles mma_tiler_dsq[:2] = mma_tiler_pdo[:2]
-        self.num_epi_stages = (self.tile_hdim // 2) // self.sdKV_epi_tile[1]
+        # headdim_64 gets 1 stage
+        self.num_epi_stages = max(1, (self.tile_hdim // 2) // self.sdKV_epi_tile[1])
         self.sdKV_flat_epi_tile = self.tile_n * (self.tile_hdim // 2) // self.num_epi_stages
-
         # TODO: dK and dV could have different shapes
         if const_expr(self.qhead_per_kvhead == 1):
             self.sdKV_layout = sm100_utils_basic.make_smem_layout_epi(
@@ -402,7 +402,7 @@ class FlashAttentionBackwardSm100:
         else:
             layout_dKV_transpose = LSE_dPsum_dQaccum_transpose
         mdK, mdV = [utils.select(t, mode=layout_dKV_transpose) for t in (mdK, mdV)]
-        dO_transpose = [1, 0, 2, 3]  # (s, h, n, b) --> (h, s, n, h)
+        dO_transpose = [1, 0, 2, 3]  # (s, h, n, b) --> (h, s, n, b)
         mdO = utils.select(mdO, mode=dO_transpose)
 
         semaphore_transpose = [2, 3, 1, 0]  # (b, n, block, stage) -> (block, stage, n, b)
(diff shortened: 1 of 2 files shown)


tridao:
For hdim64, do we have enough tmem to avoid some overlapping? 
Overlapping tmem is necessary for hdim128, but that increases the burden of synchronization (i.e. more likely to get race conditions)

tridao:
Tmem planning (assuming 128 x 128 tile):
- 128 cols for S accumulator (S = K @ Q)
- 128 cols for dP accumulator (dP = V @ dO)
- 64 cols for dK
- 64 cols for dV
- 64 cols for dQ
- 64 cols for dS (not overlapping w anything)
P will take 64 cols but can be overlapped w S (that's how it currently is).

Does this look feasible?

drisspg:
Yeah I think that makes sense, I will note that w/ the change to actually view the smem tensors correctly;
Screenshot 2025-12-01 at 7 44 56 PM
All tests are green

flash_attn/cute/flash_bwd_sm100.py
```diff
@@ -900,12 +921,10 @@ def kernel(
         else:
             sdV = storage.sdO.get_tensor(sdKV_layout, dtype=self.dv_dtype)
             sdK = storage.sQ.get_tensor(sdKV_layout, dtype=self.dk_dtype)
-        assert cute.size_in_bytes(self.do_dtype, sdO_layout) >= cute.size_in_bytes(
```
v0i0, line 903: keep the asserts, testing against buffer size?

flash_attn/cute/flash_bwd_sm100.py
```diff
@@ -320,11 +320,11 @@ def _setup_smem_layout(self):
         )
         self.sdKV_epi_tile = (
             self.tile_n,
-            128 // (self.dk_dtype.width // 8),  # 64 or 32
+            min(128 // (self.dk_dtype.width // 8), self.tile_hdim // 2),  # 64 or 32
         )  # subtiles mma_tiler_dsq[:2] = mma_tiler_pdo[:2]
-        self.num_epi_stages = (self.tile_hdim // 2) // self.sdKV_epi_tile[1]
+        # headdim_64 gets 1 stage
+        self.num_epi_stages = max(1, (self.tile_hdim // 2) // self.sdKV_epi_tile[1])
```
v0i0, line 326: if this was still 2 stages at headdim64 for gqa, we would not run into buffer issues i.e. `epi_tile = ... self.tile_hdim // (self.dk_dtype.width // 8)`, right?

flash_attn/cute/flash_bwd_sm100.py
drisspg: just moved them closer to allocation

flash_attn/cute/flash_bwd_sm100.py
drisspg: uhmm hmm so I think because we set `self.dK_reduce_ncol = 32` this wont naively work, and instead it would need to be 16. and then we might be able to split the tile and do each sub write. 

Which buffer issue though are you referring to though? there has been a few 😂

v0i0: the dO is not large enough for dV issue. if headdim_64 has two stages, it will all naturally fit

drisspg: ahhh you are right, did some experiments for headdim 64
For `Config: batch=4, seq=8192, head_dim=64, q_heads=16, kv_heads=4, gqa_multiplier=4, dtype=bf16, is_causal=False, `
I get:   `│ Backward │     3121.46 │   887.49`
time + tflops w/ the current stage allocation
and If I bump the q_stage: 3, and do_stage: 2, I get -> ` Backward │     3178.43 │   871.58 ` 


So doesn't seem like allocating more smem  elseewhere would be too helpful in this case

tridao (approved):

jayhshah:
@drisspg Can you enable hdim 64 checks in `tests/cute/test_flash_attn_race_condition.py`?

drisspg:
@jayhshah ill put up a new PR

PR #2038, opened 2025-12-02
[ROCm] Track latest `composable_kernel`

a4lg:
The latest version of `composable_kernel` supports more versatile architectures and no longer assumes wavefront size of 64 (where many compiler errors come from).

This commit updates the `composable_kernel` submodule along with necessary interface changes.

This is a part of my attempt to support Flash Attention for Strix Halo (`gfx1151`) and it seems... CK portions used by Flash Attention already supports this architecture (not entire CK, though).  I believe that my interface changes are fine (only adds/removes defaults).

p.s.
If someone has an AMD hardware already supported by Flash Attention (i.e. AMD Instinct), can you check the test results *before and after* this PR?  In my Strix Halo environment, about half of the tests fail due to high arithmetic errors and I'd like to see whether this behavior is Strix Halo-specific.
If that's not Strix Halo-specific, I'll submit a follow-up PR to allow versatile AMD GPU architectures (to `setup.py`; possibly RDNA 2 or later?).

### DIFF
--- a/csrc/flash_attn_ck/mha_bwd.cpp
+++ b/csrc/flash_attn_ck/mha_bwd.cpp
@@ -135,7 +135,10 @@ fmha_bwd_args get_ck_fmha_bwd_args(const mask_info &mask,
                          dq_acc.data_ptr(), // dq_acc
                          nullptr, // seqstart_q
                          nullptr, // seqstart_k
+                         nullptr, // seqlen_q_ptr
                          nullptr, // seqlen_k_ptr
+                         nullptr, // cu_seqlen_q_ptr
+                         nullptr, // cu_seqlen_k_ptr
                          seqlen_q,
                          seqlen_k,
                          b,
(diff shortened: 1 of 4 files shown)


tridao:
@rocking5566 does this interface change affect any existing case?

rocking5566:
@a4lg could you also change c++ version in setup.py from c++17 to c++20?
Because lastest CK use c++20 by default

rocking5566:
Actually, we are doing the similiar thing. 
just finish the testing and about to send the PR from this branch.
https://github.com/ROCm/flash-attention/tree/ck_improve_v0.1.8

a4lg:
@rocking5566
As long as CK is correctly updated, I don't stick to my changes.
BTW, my updated branch with suggested changes will be ready tomorrow (because I'm on a trip).

rocking5566:
> @rocking5566 As long as CK is correctly updated, I don't stick to my changes. BTW, my updated branch with suggested changes will be ready tomorrow (because I'm on a trip).

I submit the PR and change the c++ version.
https://github.com/Dao-AILab/flash-attention/pull/2052

We also test correctness and performance in this version (commit id) of CK in both MI300 and MI350.

a4lg:
Closing in favor of #2052.

PR #2040, opened 2025-12-02
ruff all the smaller files

drisspg:


tridao (approved):

tridao:
LGTM,  should we merge this before or after the hdim 64 bwd?

drisspg:
I think before is fine, since they should be disjoint file sets

PR #2044, opened 2025-12-03
Ko3n1g/ci/torch2.9 for cuda129

ko3n1g:


tridao:
Do we need to compile 2 separate wheels for torch 2.9, one with cuda 12.9 and one with cuda 13.0?

ko3n1g:
Potentially. Some projects are on 12.9 but want to upgrade torch. This PR was more a draft to allow building wheels one-off. Let me know what you think

PR #2048, opened 2025-12-04
Add hash attr to shortcut expensive check

drisspg:
# Summary

I noticed that on the hot path this I very expensive, and since in inductor we can ensure that nothing changes under the hoood we can always map to the right cache entry.

I tried just swapping to id(func) and we cache missed everytime, this feels robust enough.

From microbenchmarks hashing score_mod or mask_mod on my system was ~40us each and on __cute__hash hit we are down to nanoseconds

PyTorch PR: https://github.com/pytorch/pytorch/pull/169601

### DIFF
--- a/flash_attn/cute/utils.py
+++ b/flash_attn/cute/utils.py
@@ -29,10 +29,20 @@ sub_packed_f32x2 = partial(
 
 
 def hash_callable(func: Callable) -> str:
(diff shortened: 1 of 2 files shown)


tridao (approved):

tridao:
We can merge when it's ready

drisspg:
This shud be good :)

PR #2056, opened 2025-12-08
fixing cute bwd func def

liangel-02:
**Summary** 

fixes the error below when running `pytest tests/cute/test_flash_attn.py::test_flash_attn_output`

Screenshot 2025-12-08 at 5 17 32 PM


drisspg:
cc @tridao

tridao (approved):

tridao:
Thanks. Can you fix the lint error then we can merge

PR #2063, opened 2025-12-12
Fix use-after-free in FA3 deterministic mode.

skarupke:
The pytorch caching allocator actually saves us here, but if you turn it off, then compute-sanitizer will detect this.

A simple reproducer is below. Run it with compute_sanitizer and without the caching allocator. e.g like this:

PYTORCH_NO_CUDA_MEMORY_CACHING=1 compute-sanitizer python reproduce_bug.py

Where reproduce_bug.py has these contents:

import torch
import flash_attn_interface

batch_size = 1
seq_len = 16
hidden = 16

q = torch.randn(batch_size, seq_len, 2, hidden, device="cuda", dtype=torch.bfloat16)
k = torch.randn(batch_size, seq_len, 1, hidden, device="cuda", dtype=torch.bfloat16)
v = torch.randn(batch_size, seq_len, 1, hidden, device="cuda", dtype=torch.bfloat16)

q.requires_grad = True

o = flash_attn_interface.flash_attn_func(q, k, v, causal=True, deterministic=True, softmax_scale=1.0 / (hidden**0.5))

o.sum().backward()


tridao:
Thank you!

PR #2067, opened 2025-12-15
[Cute,Fwd,SM100] Add softmax precision control parameters: rescale_threshold and disable_e2e

ssxuwinter:
## 1. Motivation
While testing flash-attention (specifically the Cute-based implementation) on NVIDIA B200 (Blackwell) hardware using BF16 (bfloat16) on real-world data, we observed significant accuracy degradation compared to:
- FlashAttention-3 running on Hopper architectures.
- PyTorch's native SDPA (Scaled Dot Product Attention).

## 2. Root Cause Analysis
Upon investigation, we identified two internal Softmax configurations that negatively impact precision on this hardware/dtype combination:
- rescale_threshold:
  - The current implementation defaults this threshold to 8.0 for 16-bit precision (FP16/BF16).
  - Finding: For BF16 on B200, this default value causes noticeable precision loss. Setting rescale_threshold to 0.0 (effectively disabling the threshold check) significantly recovers accuracy.

- e2e:
  - This optimization is enabled by default.
  - Finding: In our unmasked attention test cases, explicitly disabling e2e pipelining further reduced numerical error.

## 3. Proposed Changes
To allow users to tune these precision-critical settings based on their specific hardware and data requirements, I have modified the top-level interface flash_attn_func:
- Exposed rescale_threshold (float) as an optional argument.
- Exposed e2e (bool) as an optional argument.
This change allows users to override the hardcoded defaults without modifying the kernel source code.

## 4. Benchmark Results (Accuracy)
We compared the numerical accuracy against PyTorch SDPA (Ground Truth).
Hardware: NVIDIA B200
Configuration: Batch Size: 1, Num Heads: 8, Head Dim: 128, Dtype: torch.bfloat16
As shown in the table below, configuring rescale=0.0 and e2e=Off reduces the MeanAbs error by an order of magnitude (from ~10^-4 to ~10^-5) compared to the default settings.
### Accuracy Comparison Summary (All Configs vs. Torch SDPA)
- **Batch Size**: 1
- **Num Heads**: 8
- **Head Dim**: 128
- **Data Type**: `torch.bfloat16`

| Seq Len | Metric | `rescale=None(8.0), e2e=On` | `rescale=None(8.0), e2e=Off` | `rescale=0.0, e2e=On` | `rescale=0.0, e2e=Off` |
| :---: | :---: | :---: | :---: | :---: | :---: |
| 128 | **MaxAbs** | 3.125000e-02 | 3.125000e-02 | 3.125000e-02 | 3.125000e-02 |
|  | **MeanAbs** | 6.675720e-05 | 6.675720e-05 | 6.675720e-05 | 6.675720e-05 |
| 256 | **MaxAbs** | 6.250000e-02 | 6.250000e-02 | 3.125000e-02 | 3.125000e-02 |
|  | **MeanAbs** | 2.613068e-04 | 2.555847e-04 | 3.147125e-05 | 2.563000e-05 |
| 512 | **MaxAbs** | 6.250000e-02 | 6.250000e-02 | 3.125000e-02 | 3.125000e-02 |
|  | **MeanAbs** | 3.089905e-04 | 3.070831e-04 | 5.507469e-05 | 4.959106e-05 |
| 1024 | **MaxAbs** | 6.250000e-02 | 6.250000e-02 | 6.250000e-02 | 6.250000e-02 |
|  | **MeanAbs** | 2.841949e-04 | 2.803802e-04 | 3.409386e-05 | 2.789497e-05 |
| 2048 | **MaxAbs** | 6.250000e-02 | 6.250000e-02 | 6.250000e-02 | 6.250000e-02 |
|  | **MeanAbs** | 3.376007e-04 | 3.356934e-04 | 4.482269e-05 | 3.886223e-05 |
| 4096 | **MaxAbs** | 6.250000e-02 | 6.250000e-02 | 6.250000e-02 | 3.125000e-02 |
|  | **MeanAbs** | 3.890991e-04 | 3.852844e-04 | 4.243851e-05 | 3.600121e-05 |
| 8192 | **MaxAbs** | 6.250000e-02 | 6.250000e-02 | 6.250000e-02 | 6.250000e-02 |
|  | **MeanAbs** | 4.787445e-04 | 4.768372e-04 | 4.148483e-05 | 3.504753e-05 |
| 16384 | **MaxAbs** | 6.250000e-02 | 6.250000e-02 | 6.250000e-02 | 6.250000e-02 |
|  | **MeanAbs** | 4.920959e-04 | 4.882812e-04 | 3.981590e-05 | 3.337860e-05 |
| 25634 | **MaxAbs** | 6.250000e-02 | 6.250000e-02 | 3.125000e-02 | 3.125000e-02 |
|  | **MeanAbs** | 4.749298e-04 | 4.730225e-04 | 4.673004e-05 | 4.100800e-05 |

### DIFF
--- a/flash_attn/cute/flash_fwd_sm100.py
+++ b/flash_attn/cute/flash_fwd_sm100.py
@@ -86,6 +86,8 @@ class FlashAttentionForwardSm100:
         has_aux_tensors: cutlass.Constexpr = False,
         paged_kv_non_tma: bool = False,
         is_varlen_q: bool = False,
+        rescale_threshold: Optional[float] = None,
+        disable_e2e: bool = False,
     ):
         self.use_tma_KV = not paged_kv_non_tma
         # self.dtype = dtype
@@ -145,6 +147,8 @@ class FlashAttentionForwardSm100:
             "Paged KV does not support irregular head dim"
         )
 
+        self.rescale_threshold = rescale_threshold
+        self.disable_e2e = disable_e2e
         self.softmax0_warp_ids = (0, 1, 2, 3)
         self.softmax1_warp_ids = (4, 5, 6, 7)
         self.correction_warp_ids = (8, 9, 10, 11)
@@ -1644,7 +1648,11 @@ class FlashAttentionForwardSm100:
 
             softmax = SoftmaxSm100.create(
                 softmax_scale_log2,
-                rescale_threshold=8.0 if const_expr(self.q_dtype.width == 16) else 0.0,
+                rescale_threshold = (
+                    self.rescale_threshold 
+                    if self.rescale_threshold is not None 
+                    else (8.0 if const_expr(self.q_dtype.width == 16) else 0.0)
+                ),
                 softmax_scale=softmax_scale,
             )
             softmax.reset()
@@ -1910,7 +1918,7 @@ class FlashAttentionForwardSm100:
         softmax.apply_exp2_convert(
             tSrS_t2r,
             tSrP_r2t,
-            e2e=mask_fn is None and self.head_dim_padded <= 128,
+            e2e=mask_fn is None and self.head_dim_padded <= 128 and not (self.disable_e2e),
             e2e_freq=self.e2e_freq,
         )
         # Sequence barrier arrive
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -107,6 +107,8 @@ def _flash_attn_fwd(
     out: Optional[torch.Tensor] = None,
     lse: Optional[torch.Tensor] = None,
     aux_tensors: Optional[list[torch.Tensor]] = None,
+    rescale_threshold: Optional[float] = None,
+    disable_e2e: bool = False,
 ) -> Tuple[torch.Tensor, torch.Tensor]:
     """Forward pass for FlashAttention.
 
@@ -411,6 +413,8 @@ def _flash_attn_fwd(
         pack_gqa,
         compute_capability,
         page_size not in [None, 128],  # paged KV non-TMA
+        rescale_threshold, #softmax rescale
+        disable_e2e, #softmax e2e
     )
     if compile_key not in _flash_attn_fwd.compile_cache:
         if compute_capability == 9:
@@ -459,6 +463,8 @@ def _flash_attn_fwd(
                 paged_kv_non_tma=page_size not in [None, 128],
                 is_varlen_q=cu_seqlens_q is not None
                     or seqused_q is not None,
+                rescale_threshold=rescale_threshold,
+                disable_e2e=disable_e2e,
             )
         else:
             raise ValueError(
@@ -1081,6 +1087,8 @@ class FlashAttnFunc(torch.autograd.Function):
         full_block_idx: Optional[torch.Tensor] = None,
         mask_block_cnt: Optional[torch.Tensor] = None,
         mask_block_idx: Optional[torch.Tensor] = None,
+        rescale_threshold: Optional[float] = None,
+        disable_e2e: bool = False,
     ):
         # Only create block sparse tensors if at least one block sparse parameter is provided
         block_sparse_tensors = None
@@ -1104,7 +1112,9 @@ class FlashAttnFunc(torch.autograd.Function):
             num_splits=num_splits,
             pack_gqa=pack_gqa,
             mask_mod=mask_mod,
-            block_sparse_tensors=block_sparse_tensors
+            block_sparse_tensors=block_sparse_tensors,
+            rescale_threshold=rescale_threshold,
+            disable_e2e=disable_e2e,
         )
         ctx.save_for_backward(q, k, v, out, lse)
         ctx.softmax_scale = softmax_scale
@@ -1154,6 +1164,8 @@ class FlashAttnVarlenFunc(torch.autograd.Function):
         num_splits: int = 1,
         pack_gqa: Optional[bool] = None,
         deterministic: bool = False,
+        rescale_threshold: Optional[float] = None,
+        disable_e2e: bool = False,
     ):
         out, lse = _flash_attn_fwd(
             q,
@@ -1172,6 +1184,8 @@ class FlashAttnVarlenFunc(torch.autograd.Function):
             softcap=softcap,
             num_splits=num_splits,
             pack_gqa=pack_gqa,
+            rescale_threshold=rescale_threshold,
+            disable_e2e=disable_e2e,
         )
         ctx.save_for_backward(q, k, v, out, lse, cu_seqlens_q, cu_seqlens_k, seqused_q, seqused_k)
         ctx.softmax_scale = softmax_scale
@@ -1223,6 +1237,8 @@ def flash_attn_func(
     full_block_idx: Optional[torch.Tensor] = None,
     mask_block_cnt: Optional[torch.Tensor] = None,
     mask_block_idx: Optional[torch.Tensor] = None,
+    rescale_threshold: Optional[float] = None,
+    disable_e2e: bool = False,
 ):
     return FlashAttnFunc.apply(
         q,
@@ -1241,6 +1257,8 @@ def flash_attn_func(
         full_block_idx,
         mask_block_cnt,
         mask_block_idx,
+        rescale_threshold,
+        disable_e2e,
     )
 
 
@@ -1261,6 +1279,8 @@ def flash_attn_varlen_func(
     num_splits: int = 1,
     pack_gqa: Optional[bool] = None,
     deterministic: bool = False,
+    rescale_threshold: Optional[float] = None,
+    disable_e2e: bool = False,
 ):
     return FlashAttnVarlenFunc.apply(
         q,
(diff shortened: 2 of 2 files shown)


tridao:
How do you measure numerical error? You said "PyTorch SDPA (Ground Truth)", is that done in fp32 or bf16?

ssxuwinter:
> How do you measure numerical error? You said "PyTorch SDPA (Ground Truth)", is that done in fp32 or bf16?

We measured the numerical error using Mean Absolute Error (MeanAbs) between the FlashAttention output and the PyTorch SDPA baseline.

The baseline (Ground Truth) is computed using PyTorch SDPA in BF16. We deliberately chose BF16 instead of FP32 because we wanted the benchmark to reflect the numerical behavior of a realistic training setup, where BF16 SDPA is the standard configuration.

tridao:
Can you compare against SDPA fp32? That should be "ground truth". Then you can also see how far is SDPA bf16 (and the rest) away from SDPA fp32.

ssxuwinter:
Thank you for the incredibly fast response. Below is the comparison against the FP32 baseline.As observed in the data, the deviation does indeed increase as the sequence length grows. While the magnitude of the error might make the trend less visually obvious at first glance, the raw data confirms that a gap exists.

We explicitly chose BF16 as the Ground Truth to ensure Training-Inference Alignment. Since our actual model training is performed in BF16, using BF16 as the reference best reflects our real-world production environment.
### Accuracy Comparison Summary (All Configs vs. Torch SDPA(fp32))

- **Batch Size**: 1
- **Num Heads**: 8
- **Head Dim**: 128
- **Data Type**: `torch.bfloat16`

| Seq Len | Metric | `rescale=None(8.0), e2e=On` | `rescale=None(8.0), e2e=Off` | `rescale=0.0, e2e=On` | `rescale=0.0, e2e=Off` |
| :---: | :---: | :---: | :---: | :---: | :---: |
| 128 | **MaxAbs** | 3.014469e-02 | 3.014469e-02 | 3.014469e-02 | 3.014469e-02 |
|  | **MeanAbs** | 1.507348e-03 | 1.507348e-03 | 1.507348e-03 | 1.507348e-03 |
| 256 | **MaxAbs** | 3.386784e-02 | 3.386784e-02 | 3.386784e-02 | 3.386784e-02 |
|  | **MeanAbs** | 1.511154e-03 | 1.510795e-03 | 1.474344e-03 | 1.474374e-03 |
| 512 | **MaxAbs** | 4.688931e-02 | 4.688931e-02 | 3.129482e-02 | 3.129482e-02 |
|  | **MeanAbs** | 1.489657e-03 | 1.489466e-03 | 1.442515e-03 | 1.442468e-03 |
| 1024 | **MaxAbs** | 4.387951e-02 | 4.387951e-02 | 3.254223e-02 | 3.254223e-02 |
|  | **MeanAbs** | 1.476606e-03 | 1.476383e-03 | 1.432680e-03 | 1.432648e-03 |
| 2048 | **MaxAbs** | 4.810715e-02 | 4.810715e-02 | 3.991699e-02 | 3.991699e-02 |
|  | **MeanAbs** | 1.444933e-03 | 1.444974e-03 | 1.390630e-03 | 1.390568e-03 |
| 4096 | **MaxAbs** | 4.009533e-02 | 4.009533e-02 | 3.389740e-02 | 3.389740e-02 |
|  | **MeanAbs** | 1.383549e-03 | 1.383332e-03 | 1.314834e-03 | 1.314808e-03 |
| 8192 | **MaxAbs** | 4.508495e-02 | 4.508495e-02 | 3.218174e-02 | 3.197956e-02 |
|  | **MeanAbs** | 1.336420e-03 | 1.336409e-03 | 1.243122e-03 | 1.243113e-03 |
| 16384 | **MaxAbs** | 4.065895e-02 | 4.065895e-02 | 3.468227e-02 | 3.468227e-02 |
|  | **MeanAbs** | 1.292120e-03 | 1.292084e-03 | 1.192783e-03 | 1.192755e-03 |
| 25634 | **MaxAbs** | 6.317711e-02 | 6.317711e-02 | 3.313923e-02 | 3.313923e-02 |
|  | **MeanAbs** | 1.203765e-03 | 1.203734e-03 | 1.108327e-03 | 1.108309e-03 |

tridao:
These numbers make sense to me. There's little difference between the options (at most 8-9%).
The "ground truth" for numerical error should be SDPA done in higher precision (fp32).
SDPA in bf16 uses the flash-attention algorithm from the FA(1) and FA-2 papers, so when you measure differences against SDPA in bf16, you're measuring how well the new implementation matches that exact algorithm. Not matching that algorithm is not necessarily bad (or good). Indeed, the algo is in FA4 is slightly different from the algo in FA2 (due to softmax thresholding).

I'm not yet convinced that changing the attn algorithm (while maintaining similar numerical error) leads to train-inference divergence. 

I'm hesitant to add these as options, since those options constrain the implementation (and we might have to keep supporting these options in the future). An analogy: for a kernel that adds 3 numbers a + b + c, it should not be constrained to add (a + b) first before adding c, or adding (b + c) first before adding a.

That said, if you have numbers showing changing the attn algorithm leads to train-inference divergence, we can consider adding softmax threshold as an option.

ssxuwinter:
I have conducted a deeper analysis into `cute/softmax.py`, specifically the `update_row_max` function, and identified why the default `rescale_threshold=8` causes precision issues in BF16.

### 1. Theoretical Derivation

In `update_row_max`, the logic for **skipping** the max update is as follows:
- `row_max_old` $\rightarrow m_i^{j-1}$
- `row_max_new` $\rightarrow m_i^{j} = \max(\text{rowsum}(S_i^j), m_i^{j-1})$
- `acc_scale_` $\rightarrow (m_i^{j-1} - m_i^{j}) \cdot \frac{1}{\sqrt{D}} \cdot \log_2(e)$

The code condition to skip the update (i.e., keep `row_max_old`) is:

```python
if cutlass.const_expr(self.rescale_threshold > 0.0):
    if acc_scale_ >= -self.rescale_threshold:
        row_max_new = row_max_old  # Max is NOT updated
        acc_scale = 1.0

```

Substituting the formula, the condition `acc_scale_ >= -8` implies:

$$(m_i^{j-1} - m_i^{j} ) \cdot \frac{1}{\sqrt{D}} \cdot \log_2(e) \ge -8$$

Rearranging this to find the difference required to trigger an update:

$$m_i^{j} - m_i^{j - 1} \le \frac{8 \cdot \sqrt{D}}{\log_2(e)} \approx 5.55 \cdot \sqrt{D}$$

For a standard head dimension $D=128$, this threshold is quite loose: $\approx 5.55 \cdot 11.31 \approx 62.7$. This means unless the new block's max score exceeds the old running max by more than 62.7, the algorithm will not update the row_max.

### 2. Empirical Evidence (Traces)

B=1,H=1,Seq=128*8, D=128

I instrumented the kernel to trace the behavior with real data (). The logs confirm that the condition is almost always met, causing `row_max` to "stick" to the initial value, even when the real local max increases significantly.

```text
[m_block=0, is_first=1] | row_max_old=0.000000, row_max_real=6.577559, row_max=6.577559, rescale_work=0
[m_block=0, is_first=0] | row_max_old=6.577559, row_max_real=8.285974, row_max=6.577559, rescale_work=1 (Skip Update)
[m_block=0, is_first=0] | row_max_old=6.577559, row_max_real=7.996893, row_max=6.577559, rescale_work=1 (Skip Update)
[m_block=0, is_first=0] | row_max_old=6.577559, row_max_real=9.219412, row_max=6.577559, rescale_work=1 (Skip Update)
[m_block=0, is_first=0] | row_max_old=6.577559, row_max_real=11.668362, row_max=6.577559, rescale_work=1 (Skip Update)
[m_block=0, is_first=0] | row_max_old=6.577559, row_max_real=8.158500, row_max=6.577559, rescale_work=1 (Skip Update)
[m_block=0, is_first=0] | row_max_old=6.577559, row_max_real=13.831820, row_max=6.577559, rescale_work=1 (Skip Update)
[m_block=0, is_first=0] | row_max_old=6.577559, row_max_real=13.520465, row_max=6.577559, rescale_work=1 (Skip Update)

```
In the logs above, `row_max_old` corresponds to $m_i^{j-1}$, `row_max_real` represents the true $m_i^{j}$, and `row_max` denotes the effective (approximated) $m_i^{j}$ used for calculation after the threshold check.
### 3. Conclusion

As seen in the logs, even when the real max jumps from `~6.5` to `~13.8`, the difference is within the threshold (), so `row_max` remains locked at `6.57`.

For **BF16**, this failure to update `row_max` means subsequent exponential operations are handling larger positive values:

Instead of the standard stable form where the exponent is .

Accumulating these larger values in BF16 (which has limited mantissa precision) likely leads to significant precision loss in the l and o accumulators. Setting rescale_threshold=0.0 forces the row_max to update correctly at every step, which aligns with the observed accuracy recovery.

tridao:
The large values of threshold is intentional. The point is to avoid rescaling as much as possible. Exponential can handle larger positive values just fine as long as they're below a certain threshold, which is the point of this variable.
Accumulation of l and o are done in fp32 (not bf16)

PR #2070, opened 2025-12-15
Add score-mod bwd support 

drisspg:
# Summary
 
Adds the code paths to generate correct grads when fwd uses score_mod function. 

For some score_mods (score**2 case) we will need to have S live beyond when we would typically free and reuse tmem for P. For now I am keeping S around in rmem. I verified on toy that we are good on register usage, but rechecking now since its working e2e.

The hope is that for all score-mods that dont actually require score to live long enough they will get dced 

## Todo
1. Add block sparsity and mask replay
2. Support hopper 

### Example Perf:

Rel-bias 
Screenshot 2025-12-15 at 1 29 45 PM

And a version that spills in the bwds:

image

image

### DIFF
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -573,6 +573,9 @@ def _flash_attn_bwd(
     dq: Optional[torch.Tensor] = None,
     dk: Optional[torch.Tensor] = None,
     dv: Optional[torch.Tensor] = None,
+    score_mod: Optional[Callable] = None,
+    score_mod_bwd: Optional[Callable] = None,
+    aux_tensors: Optional[list[torch.Tensor]] = None,
 ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor]:
     compute_capability = _get_device_capability()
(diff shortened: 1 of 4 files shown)


flash_attn/cute/interface.py
```diff
@@ -664,6 +667,14 @@ def _flash_attn_bwd(
     if compute_capability != 10:
         assert deterministic is False, "bwd deterministic only supported for sm100 for now"
 
+    if score_mod is not None:
+        assert score_mod_bwd is not None, "score_mod_bwd is required when score_mod is provided"
+        assert softcap == 0.0, "softcap and score_mod are mutually exclusive (different log2 scaling)"
+        assert cu_seqlens_q is None and cu_seqlens_k is None, (
+            "varlen + score_mod not supported in bwd yet"
+        )
+        assert compute_capability == 10, "score_mod in bwd only supported on SM100 for now"
```
drisspg, line 697: ima do this one in a follow up to keep PR cleaner

tridao (approved):

tridao:
Yeah i think keeping S in rmem will fit

Issue #2051, opened 2025-12-06
Bug Report: Server freezes when installing/compiling flash-attn with torch==2.9.0+cu128


SharlotAway:
Hi, thanks for the great work on FlashAttention.

I encountered a server freeze issue when installing flash-attn under PyTorch 2.9.0 + CUDA 12.8.
The problem occurs during compilation, and the machine becomes completely unresponsive (SSH disconnects and the server needs a hard reboot).

Environment

OS: Ubuntu 20.04
CUDA: 12.8
PyTorch: 2.9.0+cu128 (official wheels)
flash-attn version: latest from GitHub (also tried pip install)
GPU: NVIDIA RTX 5880 Ada Generation
Driver Version: 570.133.07
Compiler: default nvcc from CUDA 12.8
Python: 3.12

Build command:
```bash 
MAX_JOBS=4 pip install flash-attn --no-build-isolation
```
also tried installing from source
```bash
MAX_JOBS=4 python setup.py install
```
What happened

* During flash-attn compilation, CPU load spikes, and after some time the entire server freezes.
* SSH connection drops abruptly.
* No kernel logs are available because the machine becomes unresponsive.
* The only recovery method is hardware reboot.

I have successfully built flash-attn many times in other environments, including CUDA 11.x and PyTorch ≤ 2.7, so this issue seems specifically related to PyTorch 2.9 + CUDA 12.8.

rjmehta1993:
+1

Yanko-7:
+1

Romanio1997:
+1

Qubitium:
@SharlotAway Most likely your system ran out of memory due to too many threads. You can check this by compiling in a docker or vm and limiting the cpu cores/memory assigned to the docker and see if it crashes. Once there is no more memory, kernel is forced to kill processes left and right and usually unpredicatble what happens next.

Ration-Lee:
+1

qmzpg:
+1

qmzpg:
@Qubitium My computer has 56 CPU cores, I used 1/4 of them，but the compilation still freezes .

Qubitium:
@qmzpg Can you monitor the ram usage using top/htop and post the screenshot before it crashes? This will give a sense if nvcc is running away with threads (nvcc spawns threads outside the num_jobs control).

SharlotAway:
[...]

PR #2079, opened 2025-12-17
Reduce Chance of Build OOM

Qubitium:
Fix: https://github.com/Dao-AILab/flash-attention/issues/2051

When the previous auto `MAX_JOBS` code was written long ago, it was observed that 4 total nvcc threads (MAJ_JOBS=1, NVCC_THREADS=4) consumed peak ~9GB of memory during compile stage or around ~2.25GB per thread. 

However, my latest test with Torch 2.9.0 + Cuda 12.8 shows that the actual range is between 3-5GB per `NVCC_THREAD` during peak. The peak happens in the first 2 minutes of compile as the nvcc tooling is doing aggregration work of some kind.  

Latest  Observed on my dev machine:

| MAX_JOBS | NVCC_THREADS | Total Threads | Peak Memory Observed | Memory Per Thread |
|---|---|---|---|---|
| 4 | 4 | 16 | ~81 GB | 5.0 GB |
| 48 | 4 | 192 | ~616 GB | 3.2 GB |

The new code uses the worst case scenario of 5GB per nvcc thread to auto-assign a more resonable `MAX_JOBS` based on host free memory.

In issue #2051, many users were experiencing the side-effects of oom in a vm/docker env where a burst of nvcc/cicc ram grab caused ssh or even system processes to be killed. The problem is worsend by the fact that over time, cpu cores have maybe quad-druped but memory per vm actually stayed flat and not grew linearly with vm cores. 

@tridao


Qubitium:
@drisspg Please review. This simple build fix solves many users having strange oom effects due to having more threads than memory during forced kernel builds.

tridao:
Thank you!

Issue #2082, opened 2025-12-17
[Cute, Fwd] Flash forward produces incorrect row max in certain cases resulting in zeroed heads


imbr92:
### Description
The softmax computation used in the sm90 forward produces incorrect output for situations such as the one I have attached to this issue. In this case, we have seqlen_k = 448, seqlen_q = 88, batch_size = 1, q_heads_per_kv_head = 7 (GQA) with 2 total KV heads and we are doing causal attention. With the defaults, sm90 forward uses tile_m = tile_n = 128 and processes n_blocks in decreasing order (from right to left).

Currently, in `softmax.py`, when `check_inf=True`, we do the following

```
                row_max_cur = 0.0 if row_max_cur == -Float32.inf else row_max_cur
```

From my understanding, this is so that the `acc_S_row * scale_log2 - row_max_cur_scaled` computation we do right below doesn't result in nans (since -inf - -inf is nan). For something like token 2 in q, we have that the first processed n_block is fully masked out and we have check_inf = True, so we hit the above check and row_max_cur = 0.0. Then, in the subsequent n_blocks due to the input data, all of our values happen to end up < 0.0, so row_max_cur stays at 0.0 causing us to incorrectly treat that as the true row max. If this is too far from the true row_max, this shows up as effectively zeroing out the entire head for that token. This affected some short training runs I did nontrivially so its not super rare.

I haven't tested this on sm100/FA4 since I don't have easy access to a blackwell machine at the moment, but I'd assume it happens there as well unless it is accounted for elsewhere in that code.

### Repro
I'm not sure what the best way of attaching binary data to an issue is (seems `.pt` is not supported as an attachment type) so I've attached the data for a simple test case with batch size = 1 at https://github.com/adaptive-ml/flash-attn/blob/softmax_files/softmax_repro.pt to reproduce this. I ran this on an H200. (Note that I disabled compiling with `options="--enable-tvm-ffi"` to get it working with my input format but the error has nothing to do with tvm). 

Simple runner for the attached data:
https://gist.github.com/imbr92/f61d270e80fc61419b9e6c471571d55e 

#### Output
```
Q: [1, 88, 14, 64], K: [1, 448, 2, 64], V: [1, 448, 2, 64]
softmax_scale=0.125, causal=True

Ref:  min=-3.1875, max=2.7969, mean=0.0038
Cute: min=-3.1875, max=2.7969, mean=0.0039
Diff: max=1.898438, mean=0.001213

[...]

PR #2083, opened 2025-12-17
Fix softmax incorrect row_max issue

imbr92:
See https://github.com/Dao-AILab/flash-attention/issues/2082


imbr92:
Will be fixed in a different way (see discussion in https://github.com/Dao-AILab/flash-attention/issues/2082)

PR #2085, opened 2025-12-17
Add blocksparse support for bwd on blackwell

drisspg:
# Summary

Some design decisions;
* Due to softmax pipelining in the fwd we are using a block_sparse size of 2 * tile_m and by default (at least in flex in core) we produce the bwds sparse info based off of the forward info. I added subtiling with a default of 2. So that we iterate 2 for all loaded q_indices. This can be loosened / inferred from input data and aligned w/ the triton impl so that we basically just assert sparse_q > tile_m and sparse_q%tile_m == 0 and have arbitrary subtile factor

The rest of it is basically replay of what we did in the fwd. I didnt totally extract out the compute iteration order because its a lil more mixed. And instead inlined into compute and store warps. Could refactor and think harder if needed

### Perf
combined_comparison_bwd
swa size of 2048

Perf is really good, one note is that this is comparing flash (5 dot) vs triton (7 dot) and using the 2.5 multiplier in the bwd so a little unfair but still. 

My measuring stick has been to compare causal masking implemented w/  `causal = true` vs blocksparse data. The drop from sparse iteration is much smaller than what we see in the fwd (need to take another look at the fwd)

PT changes: https://github.com/pytorch/pytorch/pull/170611

### DIFF
--- a/flash_attn/cute/mask.py
+++ b/flash_attn/cute/mask.py
@@ -440,9 +440,24 @@ class AttentionMask:
         mask_seqlen: cutlass.Constexpr,
         mask_causal: cutlass.Constexpr,
         mask_local: cutlass.Constexpr,
+        mask_mod: cutlass.Constexpr[Optional[Callable]] = None,
+        batch_idx: Int32 = None,
+        head_idx: Int32 = None,
+        aux_tensors: Optional[list] = None,
+        fastdiv_mods=(None, None),
+        is_full_block: bool = False,
+        check_m_boundary: bool = True,
     ) -> None:
         """
         Backward pass: mask S = K @ Q.T where n_block tiles seqlen_k and m_block tiles seqlen_q.
+
+        Coordinate conventio:
+        - ROW corresponds to Q (m_block)
+        - COL corresponds to KV (n_block)
+
+        is_full_block: If True, skip mask_mod (all elements valid). Only apply seqlen masking.
+        check_m_boundary: If False, skip seqlen_q boundary check (optimization for non-boundary m_blocks).
+                          When iterating m_blocks in forward order, only the last m_block may be partial.
         """
         assert not (mask_causal and mask_local), "mask_causal and mask_local cannot be both True"
         ROW = 0 if const_expr(not self.swap_AB) else 1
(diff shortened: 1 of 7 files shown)


tests/cute/test_score_mod.py
```diff
@@ -614,6 +614,16 @@ def score_mod_bwd_identity(grad, score, b_idx, h_idx, q_idx, kv_idx, seqlen_info
     return grad
 
 
+@cute.jit
+def score_mod_bwd_causal(grad, score, b_idx, h_idx, q_idx, kv_idx, seqlen_info, aux_tensors):
```
drisspg, line 618: for completeness I missed in the score_mod pr

tests/cute/test_mask_mod.py
```diff
@@ -510,6 +606,91 @@ def test_sm100_block_sparse_sink_all_masked():
...
+        q_ref = q.transpose(1, 2).requires_grad_(True)
+        k_ref = k.transpose(1, 2).requires_grad_(True)
+        v_ref = v.transpose(1, 2).requires_grad_(True)
+        grad_out_ref = grad_out.transpose(1, 2)
+
+    softmax_scale = 1.0 / math.sqrt(headdim)
+
+    # Use flex_attention directly without torch.compile for backward tests
+    # torch.compile can hang on certain mask patterns (e.g., mini_causal with float32)
+    out_ref = flex_attention(q_ref, k_ref, v_ref, block_mask=block_mask, scale=softmax_scale)
```
drisspg: I need to look into this again, this was weird but on my machine the compiled verison was basically hanging

tests/cute/test_mask_mod.py
```diff
@@ -510,6 +606,91 @@ def test_sm100_block_sparse_sink_all_masked():
...
+        k_ref = k.transpose(1, 2).to(dtype).requires_grad_(True)
+        v_ref = v.transpose(1, 2).to(dtype).requires_grad_(True)
+        grad_out_ref = grad_out.transpose(1, 2).to(dtype)
+    else:
+        q_ref = q.transpose(1, 2).requires_grad_(True)
+        k_ref = k.transpose(1, 2).requires_grad_(True)
+        v_ref = v.transpose(1, 2).requires_grad_(True)
+        grad_out_ref = grad_out.transpose(1, 2)
+
+    softmax_scale = 1.0 / math.sqrt(headdim)
```
drisspg: remove let default handle

tests/cute/test_mask_mod.py
```diff
@@ -510,6 +606,91 @@ def test_sm100_block_sparse_sink_all_masked():
...
+        grad_out: Gradient of output
+        mask_mod_cute: CuTE mask modification function
+        block_sparse_mask_bwd: Block sparse tensors for backward pass
+        tile_m, tile_n: Tile sizes
+        aux_tensors: Auxiliary tensors for mask_mod (e.g., doc_ids for document masking)
+
+    Returns (dq, dk, dv) all in BSHD format.
+    """
+    headdim = q.shape[-1]
+    softmax_scale = 1.0 / math.sqrt(headdim)
```
drisspg: ditto here

flash_attn/cute/mask.py
```diff
@@ -440,17 +440,114 @@ def apply_mask_sm100_transposed(
...
         """
         assert not (mask_causal and mask_local), "mask_causal and mask_local cannot be both True"
         ROW = 0 if const_expr(not self.swap_AB) else 1
         COL = 1 if const_expr(not self.swap_AB) else 0
         assert t0ScS_t2r[0][COL] == 0, "col0 == 0"
         thr_col_offset = tScS_t2r[0][COL]
         seqlenk_col_limit = self.seqlen_k - n_block * self.tile_n - thr_col_offset
-        if const_expr(not mask_causal and not mask_local):
+
+        if const_expr(not mask_causal and not mask_local and mask_mod is not None):
```
drisspg, line 469: nit: I dont how these if els are setup really want something like
```
match mode;

blocksparse,
local/causal
_
```

tridao (approved):

PR #2101, opened 2025-12-27
[Cute] Fix: arg pass in cute flash-attn inferface

SeanLi-OI:
# Summary
Fix incorrect gradients by passing pack_gqa, mask_mod, and block_sparse_tensors into FlashAttnFunc, ensuring they are available for the backward pass in _flash_attn_bwd.


SeanLi-OI:
@drisspg @tridao minor fix when calling FlashAttnFunc with mask_mod and etc. , but not directly to _flash_attn_bwd.

SeanLi-OI:
Add more args pass for backward, I believe these updates are useful for mask_mod and block_sparsity users. :smile:

cc. @drisspg @tridao

drisspg:
@SeanLi-OI I wrote up a related proposal here https://github.com/Dao-AILab/flash-attention/issues/2154

For direct access today you can go through flex attention but agree there is some cleanup needed for the sparse path

Issue #2019, opened 2025-11-18
Cute DSL Thor (sm11.x) RTX PRO/Spark (sm12.x)


johnnynunez:
I was implementing CuTE DSL for DGX Spark https://github.com/Dao-AILab/flash-attention/pull/2017 to learn like a hobby, and i've got this issue:
```bash
FAILED tests/cute/test_mask_mod.py::test_parameterized_masks[64-128-document-None-dtype0-128-mha-16-113-203] - cutlass.cute.nvgpu.common.OpError: OpError: expects arch to be one of [Arch.sm_100a, Arch.sm_100f, Arch.sm_101a, Arch.sm_101f, Arch.sm_110a, Arch.sm_110f], but got Arch.sm_121a
Error Code: MmaF16BF16Op error
```

Investigating i was found that Jetson Thor has TMEM and DGX Spark/RTX50 no.
TMEM: https://github.com/NVIDIA/cutlass/blob/a2439551c765c5393aebe557ee75d3a0412d2211/python/CuTeDSL/cutlass/cute/nvgpu/tcgen05/copy.py#L101

So... @tridao @drisspg 
1. Thor is comptible with the code from sm10x and sm12x it is similar to Ampere code.
2. I have some tests passed in cute dsl for dgx spark like Block Sparsity.

<img width="3114" height="488" alt="Image" src="https://github.com/user-attachments/assets/8fb412ef-4666-4641-8ffa-53276fed213c" />

<img width="2774" height="730" alt="Image" src="https://github.com/user-attachments/assets/36a2581c-0ad8-4b18-a485-f80ac04373e1" />

<img width="1153" height="875" alt="Image" src="https://github.com/user-attachments/assets/6406c39a-fba3-428e-b12e-1f89330c8e93" />

wrmedford:
Can confirm, and if anyone needs access to a AGX Thor, I'm happy to provide access.

Functionally, sm110a is identical to sm100a with the small exception that it does not support stochastic rounding after poring over PTX ISA docs, in case I can save anyone else the same time.

johnnynunez:
> Can confirm, and if anyone needs access to a AGX Thor, I'm happy to provide access.
> 
> Functionally, sm110a is identical to sm100a with the small exception that it does not support stochastic rounding after poring over PTX ISA docs, in case I can save anyone else the same time.

I have one, and I have dgx spark remotely. Try my PR. Maybe I have to modify some parameters, but some tests passed from Thor

[...]

PR #2108, opened 2025-12-29
[NVIDIA] Enable Jetson Thor FA4

johnnynunez:
Jetson AGX Thor comes from gb200, it has tcgen05 and TMEM.

Functionally, sm110a is identical to sm100a with the small exception that it does not support stochastic rounding after poring over PTX ISA docs, in case I can save anyone else the same time, comment from: https://github.com/Dao-AILab/flash-attention/issues/2019#issuecomment-3630007248

@drisspg @tridao

### DIFF
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -252,7 +252,7 @@ def _flash_attn_fwd(
         else _compute_capability
     )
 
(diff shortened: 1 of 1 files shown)


johnnynunez:
Thor has the same structure as Blackwell, so it has 512 cols

tridao:
Awesome, thanks!

PR #2138, opened 2026-01-05
[Cute] API update to include FlexAttention parameters

reubenconducts:
The public `flash_attn_func` and `flash_attn_varlen_func` APIs are up to date with features currently implemented. For blocksparsity, `flash_attn_func` accepts flexattention `BlockMask` or `BlockSparseTensorsTorch` tuples. 

CC @drisspg @tridao

### DIFF
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -24,6 +24,7 @@ from functools import lru_cache
 from typing import Optional, Tuple, Callable
 
 import torch
+from torch.nn.attention.flex_attention import BlockMask
 
 
 @lru_cache(maxsize=None)
@@ -1193,20 +1194,40 @@ class FlashAttnFunc(torch.autograd.Function):
         num_splits: int = 1,
         pack_gqa: Optional[bool] = None,
         deterministic: bool = False,
+        score_mod: Optional[Callable] = None,
+        score_mod_bwd: Optional[Callable] = None,
         mask_mod: Optional[Callable] = None,
-        full_block_cnt: Optional[torch.Tensor] = None,
-        full_block_idx: Optional[torch.Tensor] = None,
-        mask_block_cnt: Optional[torch.Tensor] = None,
-        mask_block_idx: Optional[torch.Tensor] = None,
+        aux_tensors: Optional[list[torch.Tensor]] = None,
+        block_sparse_tensors: Optional[BlockSparseTensorsTorch] = None,
+        block_sparse_tensors_bwd: Optional[BlockSparseTensorsTorch] = None,
+        block_mask: Optional[BlockMask] = None,
     ):
-        # Only create block sparse tensors if at least one block sparse parameter is provided
-        block_sparse_tensors = None
-        if any(t is not None for t in [full_block_cnt, full_block_idx, mask_block_cnt, mask_block_idx]):
+        if block_mask is not None:
+            (
+                _, _,
+                # fwd blocksparse tensors
+                kv_mask_cnt,
+                kv_mask_idx,
+                full_kv_cnt,
+                full_kv_idx,
+                # bwd blocksparse tensors
+                q_mask_cnt,
+                q_mask_idx,
+                full_q_cnt,
+                full_q_idx,
+                *_,
+            ) = block_mask.as_tuple()
             block_sparse_tensors = BlockSparseTensorsTorch(
-                full_block_cnt=full_block_cnt,
-                full_block_idx=full_block_idx,
-                mask_block_cnt=mask_block_cnt,
-                mask_block_idx=mask_block_idx,
+                mask_block_cnt=kv_mask_cnt,
+                mask_block_idx=kv_mask_idx,
+                full_block_cnt=full_kv_cnt,
+                full_block_idx=full_kv_idx,
+            )
+            block_sparse_tensors_bwd = BlockSparseTensorsTorch(
+                mask_block_cnt=q_mask_cnt,
+                mask_block_idx=q_mask_idx,
+                full_block_cnt=full_q_cnt,
+                full_block_idx=full_q_idx,
             )
         out, lse = _flash_attn_fwd(
             q,
@@ -1220,7 +1241,9 @@ class FlashAttnFunc(torch.autograd.Function):
             softcap=softcap,
             num_splits=num_splits,
             pack_gqa=pack_gqa,
+            score_mod=score_mod,
             mask_mod=mask_mod,
+            aux_tensors=aux_tensors,
             block_sparse_tensors=block_sparse_tensors
         )
         ctx.save_for_backward(q, k, v, out, lse)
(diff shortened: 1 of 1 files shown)


flash_attn/cute/interface.py
```diff
@@ -1190,20 +1194,40 @@ def forward(
         num_splits: int = 1,
         pack_gqa: Optional[bool] = None,
         deterministic: bool = False,
+        score_mod: Optional[Callable] = None,
```
drisspg, line 1197: One small note is that this will cause some trouble w/ flash compile support

drisspg (commented):
.

drisspg (changes requested):
one sec

flash_attn/cute/interface.py
```diff
@@ -1193,20 +1194,40 @@ def forward(
...
+        score_mod_bwd: Optional[Callable] = None,
         mask_mod: Optional[Callable] = None,
-        full_block_cnt: Optional[torch.Tensor] = None,
-        full_block_idx: Optional[torch.Tensor] = None,
-        mask_block_cnt: Optional[torch.Tensor] = None,
-        mask_block_idx: Optional[torch.Tensor] = None,
+        aux_tensors: Optional[list[torch.Tensor]] = None,
+        block_sparse_tensors: Optional[BlockSparseTensorsTorch] = None,
+        block_sparse_tensors_bwd: Optional[BlockSparseTensorsTorch] = None,
+        block_mask: Optional[BlockMask] = None,
```
niyunsheng, line 1203: Subject: Prefer generic interface over specific `BlockMask` object for flexibility

While `FlashAttnFunc` is inherently a PyTorch class, I suggest strictly accepting a generic struct (like `BlockSparseTensorsTorch`) instead of the `BlockMask` object here.

Reasoning:
1. **Customization**: The implementation in this repo has specific optimizations (e.g., how `aux_tensors` are handled) that differ from the standard `flex_attention` design. Enforcing `BlockMask` in the signature restricts us to its specific structure and might hinder these custom efficiency gains.
2. **Abstraction Layer**: The Autograd function acts as a low-level primitive. It should operate on the data of the mask (indices/counts), not be coupled to the high-level logic of how `flex_attention` constructs it.

**Suggestion**: Let's handle the `BlockMask` unpacking in the python wrapper, and keep this Autograd interface pure and flexible.

niyunsheng: > Subject: Prefer generic interface over specific `BlockMask` object for flexibility
> 
> While `FlashAttnFunc` is inherently a PyTorch class, I suggest strictly accepting a generic struct (like `BlockSparseTensorsTorch`) instead of the `BlockMask` object here.
> 
> Reasoning:
> 
> 1. **Customization**: The implementation in this repo has specific optimizations (e.g., how `aux_tensors` are handled) that differ from the standard `flex_attention` design. Enforcing `BlockMask` in the signature restricts us to its specific structure and might hinder these custom efficiency gains.
> 2. **Abstraction Layer**: The Autograd function acts as a low-level primitive. It should operate on the data of the mask (indices/counts), not be coupled to the high-level logic of how `flex_attention` constructs it.
> 
> **Suggestion**: Let's handle the `BlockMask` unpacking in the python wrapper, and keep this Autograd interface pure and flexible.

Following up on the interface design, here is a concrete suggestion to support both modes cleanly without strictly depending on BlockMask.

We can define a superset structure for training (e.g., BlockSparseTensorsTrainTorch) that contains the extra Q-major indices required for backward, alongside the existing BlockSparseTensorsTorch (KV-only).

1. Proposed Class Definition:
```python
class BlockSparseTensorsTrainTorch(NamedTuple):
    # Forward pass data (KV-major, consistent with BlockSparseTensorsTorch)
    kv_num_blocks: torch.Tensor
    kv_indices: torch.Tensor
    full_kv_num_blocks: Optional[torch.Tensor] = None
    full_kv_indices: Optional[torch.Tensor] = None
    
    # Backward pass data (Q-major / Transposed, needed for dK/dV)
    q_num_blocks: torch.Tensor
    q_indices: torch.Tensor
    full_q_num_blocks: Optional[torch.Tensor] = None
    full_q_indices: Optional[torch.Tensor] = None
```

**2. Interface Usage:**
```python
def forward(
    ...,
    # Accepts either the lightweight inference struct or the full training struct
    block_sparse: Union[BlockSparseTensorsTorch, BlockSparseTensorsTrainTorch] = None,
):
    # ctx.block_sparse = block_sparse
```
* Inference: Users pass `BlockSparseTensorsTorch`. It's lightweight and efficient.
* Training: Users pass `BlockSparseTensorsTrainTorch`. The backward pass can then strictly check for this type to ensure q_indices are available.

This approach keeps the kernel interface strictly typed and decoupled from PyTorch's internal utility classes.

cc @reubenconducts @drisspg WDYT?

flash_attn/cute/interface.py
```diff
@@ -1300,6 +1336,9 @@ def forward(
         ctx.window_size = window_size
         ctx.softcap = softcap
         ctx.deterministic = deterministic
+        ctx.score_mod = score_mod
+        ctx.score_mod_bwd = score_mod_bwd
+        ctx.aux_tensors = aux_tensors
```
niyunsheng, line 1341: Just a heads up: assigning a list of Tensors directly to ctx can be risky (it often causes memory leaks and bypasses Autograd safety checks).

It might be safer to unpack them into `save_for_backward`, and maybe add a small pytest case to verify the gradients flow correctly through them?

flash_attn/cute/interface.py
reubenconducts: Thanks for the input. I like this, though my inclination would be to define 
```python
class BlockSparseTensorsTrainTorch(NamedTuple):
	blocksparse_tensors_torch: BlockSparseTensorsTorch
	blocksparse_tensors_torch_bwd: BlockSparseTensorsTorch
```
and then in the interface usage have
```python
def forward(
	...,
	block_sparse: Union[BlockSparseTensorsTorch, BlockSparseTensorsTrainTorch] = None, 
):
	if isinstance(block_sparse, BlockSparseTensorsTrainTorch):
		block_sparse_fwd = block_sparse.block_sparse_tensors_torch
		block_sparse_bwd = block_sparse.block_sparse_tensors_torch_bwd
	else:
		block_sparse_fwd = block_sparse
		block_sparse_bwd = None
	
	ctx.block_sparse_bwd = block_sparse_bwd
```
I will defer to @drisspg for his opinions, though.

flash_attn/cute/interface.py
reubenconducts: Sure, something like
```python
# in forward
ctx.save_for_backward(
    q,
    k,
    v,
    out,
    lse,
    *aux_tensors if aux_tensors is not None else None,
)

# in backward
q, k, v, out, lse, *_ = ctx.saved_tensors
aux_tensors = None if ctx.saved_tensors[-1] is None else ctx.saved_tensors[5:]
```
?

drisspg: yeah that looks right

flash_attn/cute/interface.py
drisspg: This is an interesting one, obviously Im a little biased but I do think the best way to currently use this API is directly through flex-attention with ` partial(flex_attention, kernel_options={"BACKEND": "FLASH"}))` which handles all the details for you.

However I do think that these APIs should be directly usable without needing to go though this path as well.


It might make sense to create a separate autograd func for this, analogous to what is done for varlen api. If I remember correctly for FA1 we had dense, varlen, and block-sparse top level APIs. cc @tridao if you have any thoughts. There are also going to be some annoying details with passing in callables if care about full graph compilation where having things separated would let us implement piecemeal

niyunsheng: Thanks @drisspg. I will try the partial(flex_attention, kernel_options={"BACKEND": "FLASH"}) best practice you mentioned.

Issue #2024, opened 2025-11-21
flash-attn hopper build fails on latest NVIDIA PyTorch images


antony-frolov:
Hi! It seems like flash attention 3 builds fail on latest NVIDIA PyTorch images. I think it might be related to recent PyTorch updates or CUDA 13.0 which latest images come with.

Here's how to reproduce the error:

`docker run --rm nvcr.io/nvidia/pytorch:25.09-py3 bash -c 'git clone https://github.com/Dao-AILab/flash-attention && cd flash-attention/hopper && MAX_JOBS=1 python setup.py install'`

```
[1/293] c++ -MMD -MF /workspace/flash-attention/hopper/build/temp.linux-x86_64-cpython-312/flash_api_stable.o.d -fno-strict-overflow -Wsign-compare -DNDEBUG -g -O2 -Wall -fPIC -I/workspace/flash-attention/hopper -I/workspace/flash-attention/csrc/cutlass/include -I/usr/local/lib/python3.12/dist-packages/torch/include -I/usr/local/lib/python3.12/dist-packages/torch/include/torch/csrc/api/include -I/usr/local/cuda/include -I/usr/include/python3.12 -c -c /workspace/flash-attention/hopper/flash_api_stable.cpp -o /workspace/flash-attention/hopper/build/temp.linux-x86_64-cpython-312/flash_api_stable.o -O3 -std=c++17 -DPy_LIMITED_API=0x03090000 -DTORCH_STABLE_ONLY -DTORCH_API_INCLUDE_EXTENSION_H -DPy_LIMITED_API=0x03090000 -DTORCH_EXTENSION_NAME=_C
FAILED: [code=1] /workspace/flash-attention/hopper/build/temp.linux-x86_64-cpython-312/flash_api_stable.o
[...]

PR #2140, opened 2026-01-05
Add FLASH_ATTENTION_FORCE_NON_STABLE_API option to allow building on NVidia Pytorch 25.09 image

jp-gr:
# Summary

This issue here comes from the handling of dev vs alpha versions of libraries in Python. 

The existing code is comparing the version of `torch` already in the 25.09 NGC image (i.e. `2.9.0a0+50eac811a6` to `2.9.0.dev20250830`). From PEP440, dev is ordered before alpha, so we fall into the `flash_api_stable.cpp` case. However the `50eac811a6` tag included in the NGC image is from 2025-08-04 (see https://github.com/pytorch/pytorch/commit/50eac811a68e63e96ad56c11c983bfe298a0bb8a ), before the `20250830` referenced in the `target_version` variable, and crucially before the `accelerator.h` file was added on 14th August: https://github.com/pytorch/pytorch/commits/release/2.9/torch/csrc/stable/accelerator.h

See https://github.com/Dao-AILab/flash-attention/issues/2024 for more discussion.

This solution does pollute the external API with yet another option but also means we make a minimally invasive change which avoids layering on more heuristics to try to figure out the desires of the user, and gives the user a bit more power as the stable API is fleshed out.

### DIFF
--- a/hopper/setup.py
+++ b/hopper/setup.py
@@ -43,6 +43,7 @@ FORCE_BUILD = os.getenv("FLASH_ATTENTION_FORCE_BUILD", "FALSE") == "TRUE"
 SKIP_CUDA_BUILD = os.getenv("FLASH_ATTENTION_SKIP_CUDA_BUILD", "FALSE") == "TRUE"
 # For CI, we want the option to build with C++11 ABI since the nvcr images use C++11 ABI
 FORCE_CXX11_ABI = os.getenv("FLASH_ATTENTION_FORCE_CXX11_ABI", "FALSE") == "TRUE"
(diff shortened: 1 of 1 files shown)


hopper/setup.py
```diff
@@ -576,7 +577,7 @@ def nvcc_threads_args():
     target_version = parse("2.9.0.dev20250830")
```
janeyx99, line 577: This is an okay solution though I wonder if changing this to the official 2.9 release would also fix? If so that might just make more sense for the long term without needing a new flag. What do you think?

jp-gr: That would ask work fine for us, but that approach might cause diffs for those using the 25.10 NGC image (perhaps that is not a concern since FA3 is beta anyway?).

We're happy either way, and appreciate that a new flag is more clutter to deal with in the future!

PR #2137, opened 2026-01-05
score-mod backward SM90

drisspg:
Stacked PRs:
 * #2158
 * #2145
 * #2146
 * #2143
 * __->__#2137
 * #2136

--- --- ---

### score-mod backward SM90

stack-pr keeps deleting my damn summary, sigh

FWD:
image

BWD:
image

insert pic, of ci being all green :) which I no longer have in my clipboard

PR #2136, opened 2026-01-05
block-sparse backward SM90

drisspg:
Stacked PRs:
 * #2158
 * #2145
 * #2146
 * #2143
 * #2137
 * __->__#2136

--- --- ---

### block-sparse backward SM90

PR #2143, opened 2026-01-06
[Cute] Clarify and fix subtle cachekey bug

drisspg:
Stacked PRs:
 * #2158
 * #2145
 * #2146
 * __->__#2143
 * #2137
 * #2136

--- --- ---

### [Cute] Clarify and fix subtle cachekey bug

### DIFF
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -868,6 +868,12 @@ def _flash_attn_bwd(
         current_stream,
     )
 
+    # NB num_threads application for 3 kernels
+    # There are pre, main, post processing kernels, currenlty num_threads is only actually
+    # used for the pre proc, and then we hard code to 384 for the main and post proc, and we do
+    # before cache key gen
+    num_threads = 384
+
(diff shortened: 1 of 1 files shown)


flash_attn/cute/interface.py
```diff
@@ -842,6 +842,12 @@ def _flash_attn_bwd(
         current_stream,
     )
 
+    # NB num_threads application for 3 kernels
+    # There are pre, main, post processing kernels, currenlty num_threads is only actually
+    # used for the pre proc, and then we hard code to 384 for the main and post proc, and we do
+    # before cache key gen
+    num_threads = 384
```
v0i0, line 875: i don't get how this changes anything, feels like a no-op edit

drisspg: TLDR is that num_threads was gettting set below where we building the cache key using the wrong num_threads. So if we end up changing the num_thread (alot of things are hardcoded for this value so things would break) the cache val would be wrong

PR #2145, opened 2026-01-07
[CUTE][SM90]Enable pack-gqa with broadcasted maskmods

drisspg:
Stacked PRs:
 * #2158
 * __->__#2145
 * #2146
 * #2143
 * #2137
 * #2136

--- --- ---

[CUTE][SM90]Enable pack-gqa with broadcasted maskmods

looks good just small tolerance issues with small N case
image

### DIFF
--- a/flash_attn/cute/block_sparse_utils.py
+++ b/flash_attn/cute/block_sparse_utils.py
@@ -116,6 +116,17 @@ def finish_overlap_v_load(
 
 
 @cute.jit
+def sparse_tensor_m_block(
+    m_block,
+    qhead_per_kvhead: cutlass.Constexpr[int],
+):
+    """Map packed m_block indices to block-sparse tensor indices."""
+    if const_expr(qhead_per_kvhead != 1):
+        return m_block // qhead_per_kvhead
+    return m_block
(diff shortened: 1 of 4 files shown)


flash_attn/cute/block_sparse_utils.py
```diff
@@ -115,6 +115,17 @@ def finish_overlap_v_load(
     return kv_producer_state
 
 
+@cute.jit
+def sparse_tensor_m_block(
+    m_block,
+    qhead_per_kvhead: cutlass.Constexpr[int],
```
v0i0, line 121: would we ever want this to be a fast divmod? we might already have one flying around for gqa right?

drisspg: let me look, not a bad idea

drisspg: so we have one for bwd but not fwd ..

drisspg: image

we dont have one in the fwd surprisngly, wired one up perf seems a lil negligible will leave as a follow up

PR #2150, opened 2026-01-08
[Cute, Bwd, Sm100] Add varlen for sm100 bwd

jayhshah:
We add varlen to sm100 backward pass and expose this capability through the `flash_attn_varlen_func` API.

Head-to-head benchmark against non-varlen sm100 backward for equal sequence lengths shows minimal overhead:

```
### headdim = 128, causal = False, seqlen = 8192, batch_size = 4, nheads = 16, nheads_kv = 16 ###
FA Python fwd: 1.445ms, 1522.0 TFLOPS
FA Python varlen fwd: 1.485ms, 1481.1 TFLOPS
FA Python bwd: 4.461ms, 1232.2 TFLOPS
FA Python varlen bwd: 4.577ms, 1201.0 TFLOPS
FA Python bwd (deterministic): 6.007ms, 915.1 TFLOPS
FA Python varlen bwd (deterministic): 6.020ms, 913.3 TFLOPS

### headdim = 128, causal = True, seqlen = 8192, batch_size = 4, nheads = 16, nheads_kv = 16 ###
FA Python fwd: 0.775ms, 1419.4 TFLOPS
FA Python varlen fwd: 0.792ms, 1388.1 TFLOPS
FA Python bwd: 2.447ms, 1123.4 TFLOPS
FA Python varlen bwd: 2.542ms, 1081.4 TFLOPS
FA Python bwd (deterministic): 2.923ms, 940.4 TFLOPS
FA Python varlen bwd (deterministic): 3.012ms, 912.6 TFLOPS
```

To fix an alignment issue with loading padded LSE in the backward kernel, we change padded offsets to FA3 style, e.g.:

```
padded_offset_q = seqlen.offset_q + batch_idx * self.m_block_size
if cutlass.const_expr(self.arch >= 90):
    padded_offset_q = padded_offset_q // self.m_block_size * self.m_block_size
```

### DIFF
--- a/flash_attn/cute/seqlen_info.py
+++ b/flash_attn/cute/seqlen_info.py
@@ -46,8 +46,6 @@ class SeqlenInfoQK:
     has_cu_seqlens_k: cutlass.Constexpr[bool]
     has_seqused_q: cutlass.Constexpr[bool]
     has_seqused_k: cutlass.Constexpr[bool]
-    tile_m: cutlass.Constexpr[cutlass.Int32]
-    tile_n: cutlass.Constexpr[cutlass.Int32]
 
     @staticmethod
     def create(
@@ -104,8 +102,6 @@ class SeqlenInfoQK:
             has_cu_seqlens_k,
             has_seqused_q,
             has_seqused_k,
-            tile_m,
-            tile_n,
         )
 
     def offset_batch_Q(


kiddyboots216:
varlen_fwd (and bwd) training matches FA2 on Blackwell 
image

flash_attn/cute/flash_bwd_sm100.py
```diff
@@ -405,7 +402,12 @@ def __call__(
         self.dv_dtype = mdV.element_type
         self.ds_dtype = self.q_dtype
 
-        if const_expr(self.qhead_per_kvhead > 1):
+        self.is_varlen_k = mCuSeqlensK is not None or mSeqUsedK is not None
+        self.is_varlen_q = mCuSeqlensQ is not None or mSeqUsedQ is not None
+        self.use_tma_store = not (self.qhead_per_kvhead == 1 and self.is_varlen_k)
```
v0i0: is that meant to be an `or`? what is the logic here?

flash_attn/cute/flash_bwd_sm100.py
```diff
@@ -718,6 +734,22 @@ class SharedStorage:
             fastdiv_mods = (seqlen_q_divmod, seqlen_k_divmod)
         self.use_block_sparsity = cutlass.const_expr(blocksparse_tensors is not None)
 
+        if const_expr(self.use_block_sparsity or aux_tensors is not None):
+            assert all(x is None for x in (mCuSeqlensQ, mCuSeqlensK, mSeqUsedQ, mSeqUsedK)), (
+                "Variable sequence length is not supported yet for blocksparse or aux tensors in bwd"
+            )
+
+        # cute.printf("mQ = {}", tma_tensor_Q.layout)
```
v0i0: remove debug prints

flash_attn/cute/flash_bwd_sm100.py
```diff
@@ -1294,12 +1336,22 @@ def load(
...
+            if const_expr(not seqlen.has_cu_seqlens_q):
+                mdO_cur = mdO[None, None, head_idx, batch_idx]
+            else:
+                mdO_cur = cute.domain_offset((0, seqlen.offset_q), mdO[None, None, head_idx])
+            if const_expr(not seqlen.has_cu_seqlens_q):
+                mLSE_cur = mLSE[None, head_idx, batch_idx]
+                mdPsum_cur = mdPsum[None, head_idx, batch_idx]
+            else:
+                padded_offset_q = (
+                    (seqlen.offset_q + batch_idx * self.tile_m) // self.tile_m * self.tile_m
```
v0i0: can we just stick this into `seqlen` or something and not repeat it a bunch of times? e.g. a `seqlen.offset_batch_padded(...)`

flash_attn/cute/interface.py
```diff
@@ -793,19 +796,29 @@ def _flash_attn_bwd(
     current_stream = cuda.CUstream(torch.cuda.current_stream().cuda_stream)
 
     if deterministic:
+        seqlen_q_rounded = (seqlen_q + m_block_size - 1) // m_block_size * m_block_size
```
v0i0: maybe just pull these out of all the conditionals

flash_attn/cute/flash_bwd_sm100.py
jayhshah: You're right to call this out, I only need to not use tma store dK/dV for cu_seqlens_k. I'll change this.

In general:
1. varlen k is the condition for the varlen scheduler
2. varlen q is a condition for checking process tile (since mblocks processed may then equal 0 if length for that batch is 0)

flash_attn/cute/interface.py
jayhshah: I'll move it up and delete redundant instance

flash_attn/cute/flash_bwd_sm100.py
jayhshah: Though we also disable tma store for seqused_q only in the forward kernel, so there will be an inconsistency here (albeit with rarely used settings)

jayhshah: To address your other implicit question: note that since we use a special padded intermediate tensor for doing TMA reduce add with dK/dV accum when gqa, we are free to use tma store then without worrying about the usual problem of overwriting other batch's outputs. So it should be 'and' and not 'or'.

For the same reason we can use TMA reduce add for dQ accum.

v0i0: so the idea is: if we have gqa we post-process, so we can always use tma, even with seqlen_k. and we can use it without seqlen_k. so the only case where we need to not use tma is not seqlen_k and not gqa.

jayhshah: Yes (assuming you mean not tma = cu_seqlens_k and mha). I also tried using postprocess with mha and cu_seqlens_k to allow for tma store (hence why I had separated out the `dKV_postprocess` boolean) but that was slightly slower.

PR #1769, opened 2025-07-22
Add torch.compile support to flash attention 3

guilhermeleobas:
Enable torch.compile support for FlashAttention and improve testing
* Add support for torch.compile to recognize FlashAttention forward/backward functions.
* Update tests to use torch._subclasses.fake_tensor.fake_check to validate the FakeTensor implementation.
* Creates a file exposing the flags used to build FlashAttention at runtime (`flash_attn_config.py`)

PR #2153, opened 2026-01-08
Update schema in `test_flash3_bw_compatibility`

guilhermeleobas:
xref: https://github.com/Dao-AILab/flash-attention/pull/1769#issuecomment-3724833664

Pull request https://github.com/Dao-AILab/flash-attention/pull/1769 breaks backward compatibility of the backward function as the C++ `mha_bwd` function returns some of its inputs (`dq`, `dk`, `dv`). 

cc: @janeyx99

### DIFF
--- a/hopper/test_flash_attn.py
+++ b/hopper/test_flash_attn.py
@@ -1222,7 +1222,6 @@ def test_flash_attn_combine(num_splits, seqlen, d, dtype):
     # pytorch_profiler(flash_attn_combine, out_partial, lse_partial)
     # pytorch_profiler(torch.sum, out_partial)
 
-@pytest.mark.xfail
 def test_flash3_bw_compatibility() -> None:
     # Let's try to always stay backward compatible! This will make life easier
     # for downstream libaries, users, and exported models.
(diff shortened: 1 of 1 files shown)


janeyx99:
Thanks, I do think the longterm solution if we were to fix the test is to update the schema.

guilhermeleobas:
Isn't the test testing for backwards compatibility with Flash Attention 2?

janeyx99:
@guilhermeleobas do you mean this schema is from FA2? https://github.com/Dao-AILab/flash-attention/blob/main/hopper/test_flash_attn.py#L1244-L1250

I wasn't aware they were tied.

guilhermeleobas:
Let me double-check that info, if FA3 schema is backward compatible with FA2.

I'm not sure if that's feasible, but I can try to return empty tensors for `dq`, `dk`, `dv` in the backward function instead of breaking backward compatibility.
https://github.com/Dao-AILab/flash-attention/blob/13696f2e5e235696a6851eada1780f7753226a68/hopper/flash_api_stable.cpp#L1645

guilhermeleobas:
> I wasn't aware they were tied.

I think you're right, FA2 and FA3 schemas are not tied.

Issue #2154, opened 2026-01-08
[RFC] FA4 pt compile plan


drisspg:
# RFC: FA4 + PT2 full graph

Driss (HUMAN NOTE): I had claude write this but I came up with this design. I just did a few rounds with it to add the details I wanted.

## Summary

Make `flash_attn_func` and `flash_attn_varlen_func` fully PT2-compliant by registering the forward/backward kernels as custom ops.

**Key changes:**
- Register custom ops for forward: `flash_attn::flash_attn_fwd`, `flash_attn::flash_attn_varlen_fwd`
- Register custom ops for backward: `flash_attn::flash_attn_bwd`, `flash_attn::flash_attn_varlen_bwd`
- Use `autograd.Function` that calls these custom ops (simpler than `register_autograd`, per Richard Zou)
- Blocksparse/callable attention stays eager-only (use `flex_attention` with Flash backend for PT2 + callables)

> **PT2 + Callables:** Use `flex_attention(..., kernel_options={"BACKEND": "FLASH"})` for full `torch.compile` support with `score_mod`/`mask_mod`. See [flex_flash_attention.py](https://github.com/meta-pytorch/attention-gym/blob/main/examples/flex_flash_attention.py).

Driss: There are some much newer apis that might make this possible without flex but this is fully fleshed out and we should recommend, you need nightly PT though.

## Current vs Proposed

**Before:**
```
flash_attn_func() → FlashAttnFunc.apply() → _flash_attn_fwd()  (not a custom op)
```

**After:**
```
flash_attn_func() → FlashAttnFunc.apply() → torch.ops.flash_attn.flash_attn_fwd  (custom op!)
                                          ↳ torch.ops.flash_attn.flash_attn_bwd  (custom op!)
```

The `autograd.Function` stays, but now it calls registered custom ops instead of raw functions.

## API Summary

| Function | Callables | torch.compile | Use Case |
|----------|-----------|---------------|----------|
| `flash_attn_func()` | ❌ | ✅ | Production inference, AOTInductor |
| `flash_attn_varlen_func()` | ❌ | ✅ | Packed sequences, production |
| `flash_attn_blocksparse_func()` | ✅ | ❌ | Blocksparse, callables (eager-only) |
| `flex_attention(kernel_options={"BACKEND": "FLASH"})` | ✅ | ✅ | PT2 + callables |

## Migration

1. **No callables**: No changes needed - same API, now PT2-compliant
[...]

PR #2164, opened 2026-01-09
[Cute] Add torch.compile support for FA4

gilfordting:
# Summary
This PR implements `torch.compile` support for the CuTe implementations of FlashAttention. See [FA4 pt compile plan](https://github.com/Dao-AILab/flash-attention/issues/2154) posted by @drisspg.
- Custom ops are registered along with their fake variants for the following functions:
	- regular fwd, bwd
	- varlen fwd, bwd
	- combine
- `flash_attn_blocksparse_func`, new eager-only function for blocksparse/callable (`score_mod`, `mask_mod`) attention; see `flex_attention` with CuTe-Flash backend for torch.compile support with callables.
- New tests to check that custom ops compile properly (see below).
# Compile tests
New `pytest` tests in `tests/cute/test_compile.py` verify that the top-level functions can be compiled and traced through properly for fwd/bwd. These should work out of the box for regular (and varlen now!); I also implemented and tested torch.compile integration for `flash_attn_combine`; this one is not yet exported in `__init__.py`, though.
# Small changes
Other small changes/standardizations include:
- `softcap` —> non-`Optional` `float` value, with default of `0.0`
- `pack_gqa` —> `Optional[bool]` (auto-detect when `None`, otherwise explicitly set)
- Removed `learnable_sink` from function signatures, since there is not yet support for backwards
- Changed `out_dtype` in `flash_attn_combine` to a non-optional parameter, since both input tensors are partials and are likely higher precision for accumulation
- Fixed an incorrect `score_mod` function used for softcap
- With the addition of split-KV, the kernel is not necessarily always batch-invariant/deterministic for the forward pass. I marked with a TODO where this would need to be fixed; will work on this in another PR

### DIFF
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -1390,22 +1390,7 @@ class FlashAttnFunc(torch.autograd.Function):
         pack_gqa: Optional[bool] = None,
         return_lse: bool = False,
         deterministic: bool = False,
-        # mask_mod: Optional[Callable] = None,
-        # full_block_cnt: Optional[torch.Tensor] = None,
-        # full_block_idx: Optional[torch.Tensor] = None,
-        # mask_block_cnt: Optional[torch.Tensor] = None,
-        # mask_block_idx: Optional[torch.Tensor] = None,
     ):
-        # Only create block sparse tensors if at least one block sparse parameter is provided
-        # block_sparse_tensors = None
-        # if any(t is not None for t in [full_block_cnt, full_block_idx, mask_block_cnt, mask_block_idx]):
-        #     block_sparse_tensors = BlockSparseTensorsTorch(
-        #         full_block_cnt=full_block_cnt,
-        #         full_block_idx=full_block_idx,
-        #         mask_block_cnt=mask_block_cnt,
-        #         mask_block_idx=mask_block_idx,
-        #     )
-
         # Always compute LSE when gradients are needed, even if not requested; this is required for the backward pass
         requires_grad = q.requires_grad or k.requires_grad or v.requires_grad
         out, lse = torch.ops.flash_attn_cute.flash_attn_fwd(


gilfordting:
Cleaning up this PR; new max_seqlen_q/k params need to be folded in and varlen bwd just got added. Need wait for tests to run as well

gilfordting:
Currently running all (83604) tests on a B200; will update this PR when they are all done. I don't think there should be any regressions, though (changes do not touch the core implementation), but will make sure everything passes before merging. Opening it up for feedback for now

flash_attn/cute/interface.py
```diff
@@ -551,7 +557,7 @@ def _flash_attn_bwd(
     m_block_size: int = 64,
     n_block_size: int = 128,
     num_threads: int = 256,
-    pack_gqa: bool = False,
+    pack_gqa: Optional[bool] = None,
```
drisspg, line 560: ?

flash_attn/cute/interface.py
```diff
@@ -96,13 +96,10 @@ def _flash_attn_fwd(
     page_table: Optional[torch.Tensor] = None,
     softmax_scale: Optional[float] = None,
     causal: bool = False,
-    softcap: Optional[float] = None,
+    softcap: float = 0.0,
```
drisspg, line 99: ?

flash_attn/cute/interface.py
```diff
@@ -1264,68 +1383,228 @@ def forward(
...
         num_splits: int = 1,
         pack_gqa: Optional[bool] = None,
+        return_lse: bool = False,
         deterministic: bool = False,
-        mask_mod: Optional[Callable] = None,
-        full_block_cnt: Optional[torch.Tensor] = None,
-        full_block_idx: Optional[torch.Tensor] = None,
-        mask_block_cnt: Optional[torch.Tensor] = None,
-        mask_block_idx: Optional[torch.Tensor] = None,
+        # mask_mod: Optional[Callable] = None,
```
drisspg: full delete

flash_attn/cute/interface.py
```diff
@@ -1667,14 +1939,61 @@ def _flash_attn_fwd_combine(
...
+        _, batch_size, seqlen, num_heads, head_size = out_partial.shape
+        out = out_partial.new_empty(batch_size, seqlen, num_heads, head_size, dtype=out_dtype)
+        lse = out_partial.new_empty(batch_size, num_heads, seqlen, dtype=torch.float32).transpose(
+            1, 2
+        )
+
+    return out, lse
+
+
+def flash_attn_combine(
```
drisspg, line 1976: Is this exposed to the dispatcher?

flash_attn/cute/interface.py
gilfordting: see the "Small changes" section of the PR; this + the pack_gqa typing are not consistent with the other function signatures so I just standardized everything

flash_attn/cute/interface.py
gilfordting: see comment from R99

flash_attn/cute/interface.py
gilfordting: done

PR #2167, opened 2026-01-10
[Cute,Bwd,Sm100] fix seqused in varlen bwd

jayhshah:
We fix errors related to using seqused_{q,k} instead of cu_seqlens_{q,k} in the sm100 backward kernel and enable additional tests for this in `test_flash_attn_varlen_output`.

Note that we currently don't store zero to the unused region for seqused_{q,k}, so the tests fill zero for this region.


jayhshah:
There are still some errors with seqused_q and zero length q batches; will fix.

Edit: done

PR #2169, opened 2026-01-11
Fix comment in tma_get_copy_fn

Edenzzzz:
The last dim here is L, because the mainloop iterates over L (m_block). K is done in one `gemm_ptx_partial`.


tridao:
No RestK here is right, as i'm thinking of matmul where the stages correspond to the K dimension (matmul of M x K @ N x K).

Edenzzzz:
Got it--it just seems to be L in FA4

PR #2180, opened 2026-01-15
[Cute][Flex]Add pack-gqa divmod

drisspg:
[Cute][Flex]Add pack-gqa divmod

This really helps with gqa fwd perf for causal via blocksparse:

image

With this change:
image

Before:
image


flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1706,6 +1716,7 @@ def softmax_loop(
                 seqlen=seqlen,
                 aux_tensors=aux_tensors,
                 fastdiv_mods=fastdiv_mods,
+            head_divmod=head_divmod,
```
v0i0: indent

PR #2195, opened 2026-01-20
ci: Use 1 ninja job for cu13

ko3n1g:
The ARM build failed on cuda13, so trying if less workers fix it.


ko3n1g:
Can confirm this is helpful and required for cuda13-arm builds. Lets please get this merged when time is good

PR #2210, opened 2026-01-25
[README.md] Update README to include `psutil` package as build dependency

wanglc02:
Added `psutil` as a build dependency in the README.
This matches the `setup` function call in `setup.py`:
```
    ...
    setup_requires=[
        "packaging",
        "psutil",
        "ninja",
    ],
    ...
```


tridao:
Thanks!

PR #2218, opened 2026-01-31
[Ai-assisted] CLC work stealing

drisspg:
Stacked PRs:
 * #2346
 * __->__#2218

--- --- ---

CLC work stealing

Not for land yet; the scheduling leaking abstraction all over the place is bad, im going to find a better way to encapsulate that. 

I had to help alot on this one even though claude and codex did most of the setup. 

Example work steal "trace" recreated from printf logs 
image

## Perf run

We would expect the highest gain for the most imbalanced workloads with the current scheduling for flex; We kind of see that e.g. alibi + causal are the same and dont currently have lpt schedule set. Document mask also sees a nice boost which makes sense

image

I dont really know why noop (fully dense fa4 with no sparsity and no scoremod) takes a hit for hdim 64 but only on  NON - GQA path, the pattern looks too regular to be chance

### What is needed to figure out before land
1. A more unified API and a mechanism for turning on with the env var. think we should universally enable for Flex use cases. I think fwd only is fine for now but likely will want bwd integration
2. I have been debugging the weirdest race condition for the 128x128 test. I have narrowed it down somewhat. My current working theory is that repsonse_ptr gets allocated with random smem data. We have num_tiles < num_sms, so only initial work is needed. We query clc, if we print after the consumer_wait in warp15 we see that we clc says no more work all invalid. The other consumer warps are not properly syncing and some of them end up pulling the random reponse data before its actually beeen populated. Racecheck shows me some error but not helping find the source of this race..
3. The register spills for no-op MHA case are weird. I also spent some time debugging. NCU points pretty much to a huge registor spill. However there is not really any good reason for this to be happening in this case and not others (AFAIK). I dumped the ptx and then compiled with 13.1 ptxas and it showed no spills in the sass (claude was helping here). Im not 100% convinced this is just random ptxas edge case but leaning that way. Also the ptxas patch thing didnt seem to be realy working .. something else to look at.

### DIFF
--- a/flash_attn/cute/flash_fwd_sm100.py
+++ b/flash_attn/cute/flash_fwd_sm100.py
@@ -14,7 +14,7 @@
 # https://github.com/NVIDIA/cutlass/blob/main/examples/python/CuTeDSL/blackwell/fmha.py
 
 import math
-from typing import Type, Tuple, Callable, Optional, Literal
+from typing import Tuple, Callable, Optional, Literal
 from functools import partial
 
 import cuda.bindings.driver as cuda
@@ -27,6 +27,7 @@ import cutlass.cute.nvgpu.tcgen05 as tcgen05
 import cutlass.utils.blackwell_helpers as sm100_utils_basic
 from cutlass import pipeline
 from cutlass.pipeline import pipeline_init_arrive, pipeline_init_wait
+from cutlass.utils import ClcDynamicPersistentTileScheduler
 from cutlass.base_dsl.arch import Arch
 from cutlass.cutlass_dsl import BaseDSL
 
@@ -36,6 +37,7 @@ from flash_attn.cute.paged_kv import PagedKVManager
 from flash_attn.cute.cute_dsl_utils import assume_tensor_aligned
 from flash_attn.cute import utils
 import flash_attn.cute.pipeline as pipeline_custom
+import cutlass.pipeline as cutlass_pipeline
 from flash_attn.cute.mask import AttentionMask
 from flash_attn.cute.softmax import SoftmaxSm100, apply_score_mod_inner
 from flash_attn.cute.seqlen_info import SeqlenInfoQK
@@ -54,12 +56,17 @@ from flash_attn.cute.named_barrier import NamedBarrierFwdSm100
 from cutlass.cute import FastDivmodDivisor
 from quack.cute_dsl_utils import ParamsBase
 from flash_attn.cute.tile_scheduler import (
+    ClcState,
+    SchedulingMode,
     TileSchedulerArguments,
+    TileSchedulerProtocol,
     SingleTileScheduler,
     StaticPersistentTileScheduler,
     SingleTileLPTScheduler,
     SingleTileVarlenScheduler,
 )
+from flash_attn.cute.fa_logging import fa_log, fa_printf
+from flash_attn.cute.utils import smid
 
 # === TUNING KNOBS (agent-editable) ===
 # Keys: (use_2cta_instrs: bool, is_causal: bool, head_dim_padded: int, is_sm103: bool)
@@ -106,6 +113,7 @@ class FlashAttentionForwardSm100:
         paged_kv_non_tma: bool = False,
         is_varlen_q: bool = False,
         use_2cta_instrs: bool = False,
+        use_clc_scheduler: bool = False,
     ):
         self.use_tma_KV = not paged_kv_non_tma
         # self.dtype = dtype
@@ -179,6 +187,32 @@ class FlashAttentionForwardSm100:
             "Paged KV does not support irregular head dim"
         )
 
+        self.use_clc_scheduler = (
+            use_clc_scheduler
+            and self.use_tma_KV
+            and not self.overlap_sO_sQ
+        )
+        self.sched_stages = 1
+        if self.use_clc_scheduler:
+            assert self.cluster_shape_mn[1] == 1, f"CLC requires cluster N == 1: {self.cluster_shape_mn}"
+            assert self.cluster_shape_mn[0] in (1, 2), f"bad CLC cluster M: {self.cluster_shape_mn}"
+            assert self.cluster_shape_mn[0] == self.cta_group_size, (
+                f"CLC cluster M != cta_group_size: {self.cluster_shape_mn}, {self.cta_group_size}"
+            )
+
+        self.scheduling_mode = SchedulingMode.CLC if self.use_clc_scheduler else SchedulingMode.STATIC
+
+        if is_varlen_q:
+            self.TileScheduler = SingleTileVarlenScheduler
+        elif self.is_causal or self.is_local or self.use_clc_scheduler:
+            self.TileScheduler = SingleTileLPTScheduler
+        elif self.is_persistent:
+            self.TileScheduler = StaticPersistentTileScheduler
+        else:
+            self.TileScheduler = SingleTileScheduler
+
+        fa_log(1, f"TileScheduler={self.TileScheduler.__name__}, scheduling_mode={self.scheduling_mode.name}, USE_2CTA={self.use_2cta_instrs}")
+
         self.softmax0_warp_ids = (0, 1, 2, 3)
         self.softmax1_warp_ids = (4, 5, 6, 7)
         self.correction_warp_ids = (8, 9, 10, 11)
@@ -218,6 +252,8 @@ class FlashAttentionForwardSm100:
             self.epilogue_warp_ids = self.correction_warp_ids
         elif self.is_varlen_q: # fallback
             self.epilogue_warp_ids = (13, 14)
+
+        self.clc_scheduler_warp_id = self.empty_warp_ids[0] if self.use_clc_scheduler else None
 
         self.tmem_s_offset = [0, self.n_block_size]  # e.g., 0, 128
         self.tmem_o_offset = [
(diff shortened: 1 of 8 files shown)


tridao:
I like the direction, CLC is the right thing to do

tridao:
Cc @tzadouri who’s thinking about scheduling and persistence

jayhshah:
In your causal benchmark the absolute speeds look lower than I expect, for instance on main I measure on B200 with benchmark_attn.py

```
### headdim = 128, causal = True, seqlen = 16384, batch_size = 4, nheads = 16, nheads_kv = 4, varlen = False, deterministic = False ###
FA Python fwd: 2.801ms, 1570.3 TFLOPS
FA Python bwd: 9.715ms, 1131.8 TFLOPS
```

With 100 repeats fwd declines to ~1520 but this is still higher than the chart. Can you share more about the setup?

tridao:
my guess is the number of repeats and how long to sleep between measurements but i'll let @drisspg clarify

drisspg:
@jayhshah Ohh so this is using the flex-bench script: https://github.com/meta-pytorch/attention-gym/blob/main/benchmarks/flex_perf.py. So this isn't the builtin causal path but causal through flex. My hope for this work is that we wont need to design a optimal schedule AOT and just let the hardware do the work. Ideally this could help to close the gap to something like the LPT schedule for blocksparse impls. And in fact we are getting pretty close

drisspg:
@jayhshah this is what im getting with the benchmark:
image


This is an aws node, and im somewhat dubious of how badly it thermally throttles. But I also have some ncu compares of dense and you can potentially less tail latency with the clc  in the pm samples


Normal schedule: 
image

CLC:
image


NCU reported roughly 3% faster in this particular case

jayhshah:
I threw up a branch [dynamic-persistent-with-semaphore](https://github.com/Dao-AILab/flash-attention/tree/jshah/dynamic-persistent-with-semaphore) with the classic (pre CLC) way of doing dynamic persistent scheduler with a semaphore, it should be useful for ablations and eventually the sm90 kernel. Benchmark shows the expected improvement for small sequence length (holding batch * seqlen constant) but regression against SingleTileLPTScheduler for large seqlen; this is more pronounced for MHA than GQA. It's also clear that the LPT swizzle needs to be tweaked for large seqlen, at least batch and head should be treated differently.

[dynamic_persistent_bench.txt](https://github.com/user-attachments/files/25110606/dynamic_persistent_bench.txt)

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -2678,8 +2811,66 @@ def epilogue_s2g(
...
+        work_tile = tile_scheduler.initial_work_tile_info()
+        while work_tile.is_valid_tile:
+            clc_pipeline.producer_acquire(clc_producer_state)
+            mbarrier_addr = clc_pipeline.producer_get_barrier(clc_producer_state)
+            tile_scheduler.advance_to_next_work(mbarrier_addr=mbarrier_addr)
+            clc_producer_state.advance()
+
+            clc_pipeline.consumer_wait(clc_consumer_state)
+            work_tile = tile_scheduler.get_current_work()
+            if cute.arch.thread_idx()[0] == 480:
```
drisspg: make this not magic number

drisspg:
Alright got 2cta working, still just forward. One thing I noticed is we are using 2cta even for M<256 which doesnt seem right

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -158,6 +164,28 @@ def __init__(
...
+        elif self.is_persistent:
+            self.TileScheduler = StaticPersistentTileScheduler
+        else:
+            self.TileScheduler = SingleTileScheduler
+
+                # todo remove b4 land
+        if FA4_DEBUG:
+            print(f"[FA4 DEBUG] use_clc_scheduler={self.use_clc_scheduler}, TileScheduler={self.TileScheduler.__name__}, USE_2CTA={self.use_2cta_instrs}")
+        # TODO: for moi Remove Before Land
+        print(f"[FA4 DEBUG] use_clc_scheduler={self.use_clc_scheduler}, TileScheduler={self.TileScheduler.__name__}, USE_2CTA={self.use_2cta_instrs}")
```
drisspg: this is the one to remove since it always prints

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -196,6 +224,12 @@ def __init__(
         elif self.is_varlen_q: # fallback
             self.epilogue_warp_ids = (13, 14)
 
+        self.clc_scheduler_warp_id = 15 if self.use_clc_scheduler else None
+        if self.clc_scheduler_warp_id is not None:
+            assert self.clc_scheduler_warp_id in self.empty_warp_ids
```
drisspg: this is a little too belt and suspenders

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1658,8 +1765,14 @@ def mma(
                 # End of GEMM_PV1(i_end) (P1 * Vi_end -> O1)
 
             # Advance to next tile
-            tile_scheduler.advance_to_next_work()
-            work_tile = tile_scheduler.get_current_work()
+            if const_expr(self.use_clc_scheduler):
```
drisspg: still want to come up with a DRY'r way to do this

flash_attn/cute/flash_fwd_sm100.py
drisspg: e.g. empty_warp_id * num_warps

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -2678,8 +2815,66 @@ def epilogue_s2g(
...
+        while work_tile.is_valid_tile:
+            clc_pipeline.producer_acquire(clc_producer_state)
+            mbarrier_addr = clc_pipeline.producer_get_barrier(clc_producer_state)
+            tile_scheduler.advance_to_next_work(mbarrier_addr=mbarrier_addr)
+            clc_producer_state.advance()
+
+            clc_pipeline.consumer_wait(clc_consumer_state)
+            work_tile = tile_scheduler.get_current_work()
+            if cute.arch.thread_idx()[0] == 480:
+                debug_printf(
```
drisspg: So I have found these debug prints really helpful, and think we shud maybe leave them in? but  im cool if we dont want to

flash_attn/cute/interface.py
```diff
@@ -146,6 +146,8 @@ def _flash_attn_fwd(
     out: Optional[torch.Tensor] = None,
     lse: Optional[torch.Tensor] = None,
     aux_tensors: Optional[list[torch.Tensor]] = None,
+    use_clc_scheduler: bool | None = None,
```
drisspg: remove from top level interface

flash_attn/cute/interface.py
```diff
@@ -288,6 +290,9 @@ def _flash_attn_fwd(
     else:
         causal, local = False, False
 
+    if use_clc_scheduler is None:
+        use_clc_scheduler = utils._get_use_clc_scheduler_default() and not causal and not local
```
drisspg: actually I think I can completely move into kernel init

drisspg:
Okay a few follow ups; one thing I found really helpful is these log messages + script to construct the trace from logs. I think having a `FA4_LOG=0,1,2,3` and different types of logging, e.g 1 is host side log should have no perf impacts, and 3 is full debug prints in kernel using cute.printf make sense to me. I can remove from this PR but curious if others agree / think its worth adding this mechanism

Being able to explicitly disable 2cta and CLC is quite convenient, currently you need to opt in . 

Also I was bad with naming and use `FA4_<var>` instead of `FLASH_ATTENTION_<var>` it is much easier to dtype :)

image
I think there are wins on dense cases.

But regardless I think this is an a mostly landable state and worth some input

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -83,6 +87,8 @@ def __init__(
         paged_kv_non_tma: bool = False,
         is_varlen_q: bool = False,
         use_2cta_instrs: bool = False,
+        use_clc_scheduler: bool = False,
+        clc_stages: int = 1,
```
tridao: should we just hardcode clc_stages to be 1, for simplicity?
cute-dsl example asserts it to be 1
https://github.com/NVIDIA/cutlass/blob/73c59c055c0fec87792470dbf33325158113db5e/examples/python/CuTeDSL/blackwell/dense_gemm_persistent_dynamic.py#L404

tridao: Nvm let's call it `sched_stage` since eventually we'll have a unified interface for all schedulers

drisspg: good call

drisspg:
Okay 
image


back to better state -> install the 13 verison of cutedsl explicitly

flash_attn/cute/interface.py
```diff
@@ -316,6 +316,7 @@ def _flash_attn_fwd(
     out: Optional[torch.Tensor] = None,
     lse: Optional[torch.Tensor] = None,
     aux_tensors: Optional[list[torch.Tensor]] = None,
+    sched_stages: int = 1,
```
drisspg: drop

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -159,6 +166,33 @@ def __init__(
             "Paged KV does not support irregular head dim"
         )
 
+        self.use_clc_scheduler = (
```
drisspg, line 190: you need to opt in (env var for now since I didnt wire up to interface) just yet but even if you request there are cases where you cant use so we guard against

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -590,6 +622,13 @@ class SharedStorage:
             # Smem tensors
             # store row max and row sum
             sScale: cute.struct.MemRange[Float32, self.q_stage * self.m_block_size * 2]
+            # CLC buffers placed here to utilize padding before sO's 1024-byte alignment.
+            # This avoids adding bytes at the end when we're at the smem limit.
+            # PipelineClcFetchAsync expects 2 * sched_stages mbarriers (full + empty).
+            clc_mbar_ptr: cute.struct.MemRange[cutlass.Int64, clc_mbar_size]
+            # CLC response storage (16 bytes per stage, stored as 4 Int32s).
```
drisspg, line 656: only keep this first arg the rest are not neeeded

tests/cute/conftest.py
```diff
@@ -59,6 +59,9 @@ def pytest_configure(config):
     os.environ["CUDA_VISIBLE_DEVICES"] = gpu_ids[worker_num % len(gpu_ids)]
 
 def pytest_collection_finish(session):
+    if not session.config.option.collectonly:
```
drisspg, line 62: cleans up logs a little

flash_attn/cute/flash_fwd_sm100.py
drisspg, line 190: you need to opt in (env var for now since I didnt wire up to interface so we we see if we can and is requested otherwise disable -> leave a comment, maybe we shoudl log to why we cant.. with a

tridao: is it just so that we can keep testing CLC for a while before flipping it to default?

drisspg: Yup exactly, I feel quite good about the correctness but the performance still has a few weird cliffs, that I want to understand better. At least in inductor it's an easy thing to autotune over

tridao:
LGTM let's merge when it's ready

PR #2222, opened 2026-02-02
[Draft][Cute,Fwd,Sm120] FA Cute DSL sm12x

johnnynunez:
With cutlass v4.4.0.dev we starting to get native behavior and support on sm12x devices. 
I know that is no the best way to code and develope and i had to do fallbacks.
The behavior is similar to sm89 MMA, without wargroup like Sm90, but with TMA. 
Remember sm12x not has tcgen05 due die space (rt cores) and DLSS algorithm.

 The idea is:
Hybrid flash attention 2/3/4 cuteDSL, working in  DGX Spark and RTX 50/PRO, i will publish and do clean code when cutlass v4.4 stable release come out. 
image

**I reiterate, we don't expose kernel, but the op is exposed, it is needed cutlass v4.4 and cuda 13.1 minimum.**
**We will jump in pytorch directly to CUDA 13.2, so sm12x will be full compatible soon** https://github.com/NVIDIA/cutlass/blob/main/python/CuTeDSL/cutlass/cute/nvgpu/warp/mma.py#L123

DGX Spark test:
image

cc @tridao @drisspg

### DIFF
--- a/flash_attn/cute/flash_fwd_sm100.py
+++ b/flash_attn/cute/flash_fwd_sm100.py
@@ -26,6 +26,8 @@ from cutlass import Float32, Int32, const_expr
 from cutlass.cute.nvgpu import cpasync
 import cutlass.cute.nvgpu.tcgen05 as tcgen05
 import cutlass.utils.blackwell_helpers as sm100_utils_basic
+from cutlass.base_dsl.arch import Arch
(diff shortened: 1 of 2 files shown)


johnnynunez:
image
checking error

image

Issue #2009, opened 2025-11-13
FlashAttention-3 Windows + CUDA 13.0 build fails with C2719 “formal parameter with requested alignment of 128 won’t be aligned”


windreamer:
**Problem**
Compilation aborts while generating the SM90 BF16 backward kernel (`flash_bwd_hdim128_bf16_softcapall_sm90.cu`).  
The MSVC front-end emits > 100 identical errors:

```
error C2719: 'unnamed-parameter': formal parameter with requested alignment of 128 won't be aligned
```

and finally:

```
fatal error C1003: error count exceeds 100; stopping compilation
```

All errors point to the auto-generated `*.cudafe1.stub.c` file, never to user code.

**Logs**

```
tmpxft_00002734_00000000-7_flash_bwd_hdim128_bf16_softcapall_sm90.cudafe1.cpp
C:\Users\runneradmin\AppData\Local\Temp\tmpxft_00002734_00000000-7_flash_bwd_hdim128_bf16_softcapall_sm90.cudafe1.stub.c(190): error C2719: 'unnamed-parameter': formal parameter with requested alignment of 128 won't be aligned
C:\Users\runneradmin\AppData\Local\Temp\tmpxft_00002734_00000000-7_flash_bwd_hdim128_bf16_softcapall_sm90.cudafe1.stub.c(217): error C2719: 'unnamed-parameter': formal parameter with requested alignment of 128 won't be aligned
C:\Users\runneradmin\AppData\Local\Temp\tmpxft_00002734_00000000-7_flash_bwd_hdim128_bf16_softcapall_sm90.cudafe1.stub.c(226): error C2719: 'unnamed-parameter': formal parameter with requested alignment of 128 won't be aligned
C:\Users\runneradmin\AppData\Local\Temp\tmpxft_00002734_00000000-7_flash_bwd_hdim128_bf16_softcapall_sm90.cudafe1.stub.c(235): error C2719: 'unnamed-parameter': formal parameter with requested alignment of 128 won't be aligned
C:\Users\runneradmin\AppData\Local\Temp\tmpxft_00002734_00000000-7_flash_bwd_hdim128_bf16_softcapall_sm90.cudafe1.stub.c(253): error C2719: 'unnamed-parameter': formal parameter with requested alignment of 128 won't be aligned
[...]

PR #2223, opened 2026-02-02
[FA3] Mark current main version as v3.0.0 stable

lw:
A collaboration between Flash-Attention, PyTorch and xFormers is trying to provide pre-built wheels for FA3 across as many platforms/environments as possible (e.g., ARM, Windows, CUDA 13, ...). To simplify the installation workflow, it would help to tag these packages as stable, but the current main version is tagged as beta.

FA3 hasn't received substantial updates in a while (the latest was a bugfix almost two months ago), and most new development is happening in FA4. Thus, in this PR, I propose we just claim that the current main version _is_ stable.

I have heard concerns that the feature set of FA3 doesn't currently match FA2 (e.g., dropout is missing). I think this concern is partly addressed by the fact that the new wheels will have a different name than the FA2 ones (`flash_attn_3` and `flash_attn` respectively), hence the former does _not_ claim to be a replacement for the latter, and the two can coexist (and they provide different modules).


johnnynunez:
> A collaboration between Flash-Attention, PyTorch and xFormers is trying to provide pre-built wheels for FA3 across as many platforms/environments as possible (e.g., ARM, Windows, CUDA 13, ...). To simplify the installation workflow, it would help to tag these packages as stable, but the current main version is tagged as beta.
> 
> FA3 hasn't received substantial updates in a while (the latest was a bugfix almost two months ago), and most new development is happening in FA4. Thus, in this PR, I propose we just claim that the current main version _is_ stable.
> 
> I have heard concerns that the feature set of FA3 doesn't currently match FA2 (e.g., dropout is missing). I think this concern is partly addressed by the fact that the new wheels will have a different name than the FA2 ones (`flash_attn_3` and `flash_attn` respectively), hence the former does _not_ claim to be a replacement for the latter, and the two can coexist (and they provide different modules).

cc @tridao @drisspg

tridao:
This seems fine to me, let's have it as the `flash_attn_3` package.

tridao (approved):

windreamer:
Hi @lw ,

Thanks for the context on FA3's stability status. I'm reaching out regarding the new `flash_attn_3` package on PyPI.

I've been providing **unofficial pre-built wheels** for FA3 at [https://windreamer.github.io/flash-attention3-wheels/](https://windreamer.github.io/flash-attention3-wheels/) to help users on various platforms (ARM, Windows, newer CUDA versions, etc.) who need FA3 but can't easily build from source.

However, I noticed that a new `flash_attn_3` package was recently published to PyPI that appears to be essentially an **empty package** (no actual wheels or functional code). This is causing a significant issue: users who have my unofficial wheels installed, or who try to install `flash_attn_3` expecting functional code, are now getting this empty package instead via `pip install flash_attn_3`.

**Would it be possible to address this?** Options could include:
- Removing or yanking the current empty package from PyPI
- Uploading actual functional wheels to the official `flash_attn_3` package
- Or coordinating so the unofficial wheels can coexist without conflict

Happy to discuss further if helpful. The goal is just to ensure users can actually get working FA3 installations without confusion.

Thanks for your time and all the work on Flash Attention!

lw:
Hey @windreamer,

Sorry, I wasn't aware of your work providing pre-built wheels for FA3! I'm a maintainer of xFormers and we found ourselves in a similar situation (many Windows users installed xFormers just to get the pre-built FA3 module). If I had been aware we should have coordinated earlier!

I'm working on building some "official" FA3 wheels in collaboration with @liangel-02, @drisspg and @janeyx99 from PyTorch, and we're currently putting our candidate wheels in https://download.pytorch.org/whl/test/cu126/flash-attn-3 (and the corresponding indices for the other CUDA versions). We're actually almost done and we'll likely move them to the stable index this week.

I did claim the `flash_attn_3` package on PyPI in preparation for this. It does currently contain an empty wheel, but we plan to replace it shortly with one of the wheels I mentioned above.

lw:
I "yanked" the 0.0.0 release of `flash-attn-3` from pypi.org. I had planned to upload the actual wheels there, but they exceed the current file size limit. My request to increase that limit (https://github.com/pypi/support/issues/9311) is not currently going through.

woct0rdho:
Hi, I'd like to ask a tangential question: How is it going with the libtorch stable ABI support in FA3?

I've been maintaining [SageAttention](https://github.com/woct0rdho/SageAttention) Windows wheels for a while, and I've ported it to both Python stable ABI (ABI3) and libtorch stable ABI. We no longer need to build a different wheel for each Python version and torch version.

AFAIK libtorch stable ABI support in FA3 is tracked in https://github.com/pytorch/pytorch/issues/154908 but there are still unresolved items. It should have progressed a lot with the changes in torch 2.10 . The wheels at https://windreamer.github.io/flash-attention3-wheels/ are not using libtorch stable ABI. Are the wheels at https://download.pytorch.org/whl/flash-attn-3/ using libtorch stable ABI? I can't see it from the filenames.

woct0rdho:
Ok, the wheels in https://download.pytorch.org/whl/flash-attn-3/ are ABI stable. Thanks for your work!

janeyx99:
@woct0rdho Thanks for the question!

Glad to hear that SageAttention's been ported and enjoying the benefits!

Thanks for bringing that issue to my attention--FA3 is now ABI stable with libtorch and the wheels in the pytorch download link are built to be ABI stable with torch 2.9+. If you build FA3 with torch 2.9 and above, there is logic in setup.py to default to using the ABI stable code. The old code is left in the repo for earlier support.

woct0rdho:
Do you know why there is no wheel for Windows in https://download.pytorch.org/whl/cu130/flash-attn-3/ , while there are wheels for Windows on cu126/cu128/cu129? Is there some workflow or CI log for us to see?

lw:
@woct0rdho See https://github.com/Dao-AILab/flash-attention/issues/2009 for why building for Windows with CUDA 13 fails. Please comment there if you have any insights.

PR #2228, opened 2026-02-04
Benchmark: O(1) Memory Waller Operator vs FlashAttention

RegularJoe-CEO:
## Summary

Adds a memory scaling benchmark comparing the Waller Operator against FlashAttention v2 and standard PyTorch attention.

## Results (NVIDIA H100 80GB HBM3)

| SeqLen | PyTorch | Flash2 | Waller | Reduction |
|--------|---------|--------|--------|-----------|
| 512 | 0.0706 | 0.0391 | 0.0010 | 97.4% |
| 1024 | 0.1025 | 0.0391 | 0.0010 | 97.4% |
| 2048 | 0.1680 | 0.0391 | 0.0010 | 97.4% |
| 4096 | 0.3047 | 0.0391 | 0.0010 | 97.4% |
| 8192 | 0.6016 | 0.0391 | 0.0010 | 97.4% |
| 16384 | 1.2891 | 0.0391 | 0.0010 | 97.4% |
| 32768 | - | 0.0470 | 0.0010 | 97.9% |
| 65536 | - | 0.0627 | 0.0010 | 98.4% |
| 131072 | - | 0.0942 | 0.0010 | 98.9% |
| 262144 | - | 0.1572 | 0.0010 | 99.4% |

Memory in GB. Config: heads=1, head_dim=64, causal=True, float16.

## Key Finding

Waller Operator maintains **constant 0.001 GB memory** regardless of sequence length, achieving **97-99% memory reduction** vs FlashAttention v2.

## Files

- `benchmarks/benchmark_waller_comparison.py` - Reproducible benchmark script

## Links

- Evaluation binary: https://github.com/RegularJoe-CEO/waller-eval
- Contact: e@ewaller.com
- Patent pending

### DIFF
--- a/benchmarks/benchmark_waller_comparison.py
+++ b/benchmarks/benchmark_waller_comparison.py
@@ -1,0 +1,59 @@
+#!/usr/bin/env python3
+import torch, math, subprocess, json, os, re
+from datetime import datetime
+try:
+    from flash_attn import flash_attn_qkvpacked_func
+    HAS_FLASH = True
+except:
+    HAS_FLASH = False
+WALLER_BINARY = os.path.expanduser("~/waller-eval-repo/waller_eval_cli_x86")
+def get_gpu_info():
+    return {"name": torch.cuda.get_device_name(0), "total_memory_gb": torch.cuda.get_device_properties(0).total_memory / (1024**3)}
+def measure_pytorch(bs, sl, nh, hd):
+    torch.cuda.reset_peak_memory_stats(); torch.cuda.empty_cache()
+    try:
+        q=torch.randn(bs,nh,sl,hd,device='cuda',dtype=torch.float16)
+        k=torch.randn(bs,nh,sl,hd,device='cuda',dtype=torch.float16)
+        v=torch.randn(bs,nh,sl,hd,device='cuda',dtype=torch.float16)
+        with torch.no_grad():
+            s=torch.matmul(q,k.transpose(-2,-1))/math.sqrt(hd)
+            m=torch.triu(torch.ones(sl,sl,device='cuda',dtype=torch.bool),diagonal=1)
+            s.masked_fill_(m,float('-inf'))
+            a=torch.softmax(s,dim=-1); o=torch.matmul(a,v)
+        torch.cuda.synchronize(); mem=torch.cuda.max_memory_allocated()/(1024**3)
+        del q,k,v,s,a,o; torch.cuda.empty_cache()
+        return {"status":"ok","peak_memory_gb":round(mem,4)}
+    except: torch.cuda.empty_cache(); return {"status":"OOM","peak_memory_gb":None}
+def measure_flash(bs, sl, nh, hd):
+    if not HAS_FLASH: return {"status":"N/A","peak_memory_gb":None}
+    torch.cuda.reset_peak_memory_stats(); torch.cuda.empty_cache()
+    try:
+        qkv=torch.randn(bs,sl,3,nh,hd,device='cuda',dtype=torch.float16)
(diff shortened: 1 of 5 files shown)


ManuelFay:
Hello !
I don't think anyone claims constant GPU  memory is impossible at inference (if you are okay with offloading everything on CPU / recomputing stuff), sequential operations through streaming are possible and that's the partly the point of FA. 
The whole issue here is that this will cause huge amounts of memory movements and it will be super slow.
The objective with efficient attention is to optimize throughput - saturate the available GPU memory and then make sure it goes fast.

RegularJoe-CEO:
Tahnk you for the comment. You are right that streaming/tiling is known. FA does this brilliantly, but it still scales the memory with the sequence length. FA grows, but this stays flat. Looking at the benchmarks, it isn't a marginal improvemnt, it is a different scaling curve. The memory movement concerns are valid in theory, but in practice the kernels sustain competitive throughput because fewer memory allocations means less synchronization overhead & better cache utilization. The bottleneck shifts. But please do not take my word for it, the binary is there and if you run it and you find it to be slow, I would really like to understand what you're seeing and discuss it further. I am extremely new to this. I have bench test tested it, but I don't understand how to fully test it and how to fully load it like you guys do. I'm very hopeful. Someone pulls it down and just tries to break it and finds that there are other weak spots that I have not been able to locate yet.

ManuelFay:
Wanted to give the benefit of the doubt - seems weird not to report any latencies here.

Weirder still is to post a PR that links to another repo, without the kernel code, just a binary. I don't want to assume the worst, but that + the fact you comment on random people's twitter just makes it look like a bad attempt at malware... I don't think anyone will run someone's binary like that.

I recommend closing the PR and wish you good luck if you were honest !

RegularJoe-CEO:
OK, that's a fair assessment. I spent most of my life in a different industry, so I have struggled to meet anyone in this industry, I should probably spend years getting to know people where it looks a little less spastic, but I just don't have that amount of time at my age. By the way, you were correct, when I went back, it was horrible and it was doing it one at a time, so I just worked on a different method and it is done. I now understand this looks like malware, but this is so valuable that I can't possibly expose any code or tell anyone how I did this whatsoever that would be insane for me personally.. there's nothing more I'd rather do than share this with the community but at this point, I would be starving my family and that's crazy. I dont know how to share itShould I close this pull request and open a new one with the updated latency numbers or should I do something differently? Thank you so much for your responses. I genuinely mean that. I'm not a fool. I'm just new to this tool and the platform where the scientist communicate in the scientific way that I'm not used to or familiar with because they didn't teach me that at the online university when I got my business degree unfortunately.. And also to be fair, I'm new with this, I just thought everybody that was doing this all day for a living and getting paid for it would know how to open it in a safe environment because I don't know that kind of thing. I just thought these multi billion dollar companies like had something that you know kind of helped with that. It's all just ignorance and it's all temporary ignorance because I will learn because I did do this. I also created a cross platform deterministic engine that delivers the same results on a Mac MacBook as it does a H2 hundred or an H 100 and these results are a bit exact every single time.. the engine completely eliminated floating point drift from non-linear expressions. I will do anything include give my phone number or email or anything where somebody will look at this because it actually works, and it can actually save an unbelievable amount of electricity. Attached are the latency numbers from 5 minutes ago. 

Seq Len | Standard           | Waller Op          | FlashAttn v2   | Win
65k        | 47-73ms/32GB | 4.8ms/0.09GB | 5-15ms/2GB   | Speed 1-3x Mem 20x
512k       | OOM 1TB           | 12ms/4MB       | OOM                | Infinite

PR #2239, opened 2026-02-07
Add `FLASH_ATTENTION_TRITON_AMD_CONFIG_JSON` env var support

alexheretic:
Add an env var for configuring `attn_fwd` triton.Config. This allows setting of more optimal config without having to enable autotuning, which has issues.

## Motivation
With autotuning disabled the current default configs are sub-optimal, at least for my gfx1100 but probably for other cards too.

I get roughly 2x faster wan2.2 workflow performance using config `FLASH_ATTENTION_FWD_TRITON_AMD_CONFIG_JSON='{"BLOCK_M":128,"BLOCK_N":64,"waves_per_eu":1,"PRE_LOAD_V":false,"num_stages":1,"num_warps":8}'`

Autotuning itself is very slow to run (e.g. takes 82 minutes) and needs to re-run on any param change (it also isn't persisting maybe?). So it is helpful to be able to override the !autotune config.

## Alternatives
Provide better default configs, e.g. use the above example for gfx1100. This is less flexible though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,9 +145,9 @@ FLASH_ATTENTION_TRITON_AMD_ENABLE="TRUE" pytest tests/test_flash_attn_triton_amd
 
 For better performance, enable autotune with `FLASH_ATTENTION_TRITON_AMD_AUTOTUNE="TRUE"`.
 
-Alternativly, if _not_ autotuning, `FLASH_ATTENTION_TRITON_AMD_CONFIG_JSON` may be used to set a single triton config overriding the hardcoded defaults. E.g.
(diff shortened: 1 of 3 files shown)


tridao:
Is this only for the forward pass? Would the backward pass need a different set of config?

alexheretic:
> Is this only for the forward pass? Would the backward pass need a different set of config?

My motivating use case is only for `attn_fwd`. Since bwd has different default config already it probably makes sense to name this as a forward only override, perhaps `FLASH_ATTENTION_FWD_TRITON_AMD_CONFIG_JSON`?

alexheretic:
I've updated it to be for `attn_fwd` only, now `FLASH_ATTENTION_FWD_TRITON_AMD_CONFIG_JSON`. I can't edit the PR title for some reason though :disappointed:

PR #2251, opened 2026-02-11
[Cute] Handle window_size=(-1, -1) for non-local attention

henrylhtsang:
In flash attention cute, we use (None, None) to handle not using local attention. This PR tries to treat (-1, -1) as the same to avoid some tricky debugging situation.

### DIFF
--- a/tests/cute/test_flash_attn.py
+++ b/tests/cute/test_flash_attn.py
@@ -168,6 +168,7 @@ def test_flash_attn_output(
             window_size = (-window_size[0], None)
         if local:
             print("window size = ", window_size)
+        # window_size = (-1, -1) if not local else (16, 0)
         if has_learnable_sink:
             learnable_sink = torch.randn(nheads, dtype=torch.bfloat16, device=device)
         else:
--- a/tests/cute/test_flash_attn_race_condition.py
+++ b/tests/cute/test_flash_attn_race_condition.py
@@ -138,6 +138,7 @@ def test_flash_attn_output(
             window_size = (-window_size[0], None)
         if local:
             print("window size = ", window_size)
+        # window_size = (-1, -1) if not local else (16, 0)
         if has_learnable_sink:
             learnable_sink = torch.randn(nheads, dtype=torch.bfloat16, device=device)
         else:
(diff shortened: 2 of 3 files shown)


henrylhtsang:
Well apparently the tests allow negative window sizes https://github.com/Dao-AILab/flash-attention/blob/c4d8b0630eb81cf88206e0cc9e9bff4e7806d88f/tests/cute/test_flash_attn.py#L165-L168

So maybe this PR is wrong..

tridao:
Yeah i realized that sometimes people use negative window size. In FA3 that wouldn't work at all. In FA4 it would work. But now we have this backward-breaking change. Not sure what to do here.
I'm open to just changing the behavior to "negative window size -> set to None"

henrylhtsang:
@tridao Thanks for the response! I modified the PR to do that and remove local_enum from the tests. Running the tests overnight so will get back in the morning.

jayhshah:
Is there a need to remove support for negative window sizes? I had added and checked that due to a request.

jayhshah:
How about if window_size[0] and window_size[1] are both provided, we check that the inequality
```
window_size[0] + window_size[1] >= 0
```
is satisfied, and if it is not, set window_size = (None, None) and disable local? In particular, this covers (-1, -1), but not the valid cases of (None, any) and (any, None), as well as window with any offset.

henrylhtsang:
> How about if window_size[0] and window_size[1] are both provided, we check that the inequality
> 
> ```
> window_size[0] + window_size[1] >= 0
> ```
> 
> is satisfied, and if it is not, set window_size = (None, None) and disable local? In particular, this covers (-1, -1), but not the valid cases of (None, any) and (any, None), as well as window with any offset.

I am happy with this. Let me update

henrylhtsang:
> Is there a need to remove support for negative window sizes? I had added and checked that due to a request.

I think people assumed this would work. I got burned by this two times so I submitted this PR.

henrylhtsang:
tested:
```
pytest -n 10 test_flash_attn.py::test_flash_attn_output -x -v -k "128-128-128-False"
```

```
pytest -n 10 test_flash_attn.py::test_flash_attn_output -x -v -k "128-128"
```

henrylhtsang:
@jayhshah maybe take another look?

PR #2253, opened 2026-02-12
Deterministic backward for blocksparse impl

drisspg:
Stacked PRs:
 * #2512
 * __->__#2253

--- --- ---

# Deterministic backward for blocksparse impl

PyTorch PR:  https://github.com/pytorch/pytorch/pull/174813

## Summary

This make it possible to run bwd with blocksparse tensors w/ determinism. 

At a very high level,  the current implementation needs two things

1.  For every M x N block, a way to figure out  for N_i is it this columns turn to accumulate DQ into the global memory. For dense it can infer this by predefined sparsity, I.e. dense, causal, sliding window. 
2. Some coordination with the tile schedule to ensure that we throws work tiles in the queue in a way that will not block any given row from making progress.

This is a fun little problem and I made some animations: https://x.com/drisspg/status/2022438130893426731?s=20

But that being said I feel as though there are two approaches  for block sparsity, we could either compute the semaphore lock value for every N tile at runtime based off of the sparse data, Or we can precompute the lock values ahead of time. I did not go with approach 1 and went with approach 2 because in terms of total work, since sparsity is typically consistent across layers we get to amortize the construction and I was concerned that doing the scan across kv_blocks and indices would be too much overhead.

### So what does this look like

We introduced 3 new attributes to `BlockSparseTensors` ( I wish this was 2 but we will come back to this):
New dq_write_order(partial|full) mirroring the q_block_idx(num_blocks) tensors.

They are used as such:

```Py
dq_write_order: int32[B, H, num_n_blocks, max_q_per_n]
```

Parallel to `q_block_idx[B, H, num_n_blocks, max_q_per_n]`. 
Each KV column worker iterates from 0 to max_q_per_n, so for the i-th M-block in the CTA's iteration list:

```Py
m_block       = q_block_idx[b, h, n_block, i]
my_lock_value = dq_write_order[b, h, n_block, i]
```

If we we can figure out via the total counts whether I corresponds to partial or full block and thats basically it. See animation :)

### Scheduling

I implemented this and perf was terrible since by default we would use the `non-spt` path. SPT is fancy word for right to left. `spt=shortest processing time` for most causal masks they look lower triangular which means the amount of work for a column is highest on the left, since q_idx = 0 which is >= to all kv_indices and thus all m_blocks will contribute - and smallest on the right since the only block that has any actual dQ to compute is the last block for Nth columns iteration. 

```
         n_block (KV)
         0    1    2    3    4
       ┌────┬────┬────┬────┬────┐
   0   │ ██ │    │    │    │    │
       ├────┼────┼────┼────┼────┤
   1   │ ██ │ ██ │    │    │    │
m      ├────┼────┼────┼────┼────┤
(Q) 2  │ ██ │ ██ │ ██ │    │    │
       ├────┼────┼────┼────┼────┤
   3   │ ██ │ ██ │ ██ │ ██ │    │
       ├────┼────┼────┼────┼────┤
   4   │ ██ │ ██ │ ██ │ ██ │ ██ │
       └────┴────┴────┴────┴────┘
work:    5    4    3    2    1
```

Bad scheduling is terrible for perf, you turn beautiful parallel GPU into terrible serialized turtle.  

So I also add 1 more flag which is SPT. And that says in stead of scheduling using the default left to right flip the lock values so that go from lowest to highest from right left. This does require coordination with the scheduler. Why? Again lets look at causal for simplicity -
Lets say your num_sms was 2 and your total KV blocks was 5 if you did the default scheduling we would throw work into the queue starting with kv_block 0 and kv_block 1 for m_tile 0 we would be good for m_tile 1 we would also be good since there is their lock values are 0 and then 1,0 but for m_row 2 we would have lock values (2, 1) and semaphore value 0 is at kv_block 2! So our workers will never be able to continue

```
         n_block
         0    1    2    3    4
       ┌────┬────┬────┬────┬────┐
   0   │ 0  │    │    │    │    │  ✓ only CTA 0, lock=0
       ├────┼────┼────┼────┼────┤
   1   │ 1  │ 0  │    │    │    │  ✓ CTA 1 lock=0 goes first, then CTA 0
m      ├────┼────┼────┼────┼────┤
(Q) 2  │ 2  │ 1  │ 0  │    │    │  ✗ lock=0 is at n_block=2 — not running!
       ├────┼────┼────┼────┼────┤
   3   │ 3  │ 2  │ 1  │ 0  │    │
       ├────┼────┼────┼────┼────┤
   4   │ 4  │ 3  │ 2  │ 1  │ 0  │
       └────┴────┴────┴────┴────┘
```

So long story short with SPT for most causally blocks we are seeing around the 20-30 perf drop from deterministic bwd -> see attached data.  

### Better Scheduling

I was toiling with ideas for more optimal scheduling. Since the distribution of tiles can be arbitrary for BlockSparse, I think if we could ahead of time compute a `least blocking schedule` and then we use this schedule to submit kv_tiles, again this requires coordination with the scheduler and would probably add some slow down since get_next_tile_info will need a global read to figure out where to go.  I think we would need to have 1 lock value per column in order to avoid any deadlocks and I thought of some good heuristics but afaik this is NP complete. Hard to say if all this AOT work would be worth the potentially faster bwd.  Sounds like a future problem.

Although lets say we did this I do kinda want `spt` attribute to really be `schedule` attribute which can either be non spt, spt or a permutation.  Couldnt think of anything really elegant here ** ideas from reviewer welcome**.

#### Fuzzing

Side note I did a bunch of fuzzing for this PR, one really nice aspect (or not lol) is that the ratio of compile time to gpu work is very high so both testing and fuzzing is very amenable to parallel workers.

image

image

This fuzzing lead to this PR: https://github.com/Dao-AILab/flash-attention/pull/2258

### Performance:

Sliding window size = 2048 and eventually starts to get majorly blocked
This is with SPT = True
image

And here is SPT = False
image

### DIFF
--- a/flash_attn/cute/block_sparsity.py
+++ b/flash_attn/cute/block_sparsity.py
@@ -19,9 +19,13 @@ class BlockSparseTensors(NamedTuple):
     mask_block_idx: cute.Tensor
     full_block_cnt: cute.Tensor | None
     full_block_idx: cute.Tensor | None
+    dq_write_order: cute.Tensor | None = None
+    dq_write_order_full: cute.Tensor | None = None
 
     def __new_from_mlir_values__(self, values):
         if len(values) == 2:
+            values = (*values, None, None, None, None)
+        elif len(values) == 4:
             values = (*values, None, None)
         return BlockSparseTensors(*values)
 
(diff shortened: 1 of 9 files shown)


flash_attn/cute/interface.py
```diff
@@ -915,6 +919,30 @@ def _flash_attn_bwd(
             block_size=(m_block_size, n_block_size),
             subtile_factor=subtile_factor,
         )
+        if deterministic:
```
drisspg, line 1557: this file is getting very complicated and hard to read from all the error checking.. I want to refactor jus so the main entrypoint flow is easier to grok

flash_attn/cute/flash_bwd_sm100.py
```diff
@@ -2522,7 +2586,10 @@ def dQacc_reduce(
                     pipeline_dQ.consumer_release(dQ_consumer_state)
                 dQ_consumer_state.advance()
 
-                gdQaccum_cur = gdQaccum[None, None, m_block]
+                m_block_store = m_block
+                if m_block_oob_upper and m_block_max > 0:
```
drisspg: hmm should make this closer to what it was

flash_attn/cute/flash_bwd_sm100.py
```diff
@@ -2401,6 +2406,48 @@ def compute_loop(
...
+            if sparse_iter < curr_q_cnt:
+                assert curr_dq_write_order is not None
+                lock_value = curr_dq_write_order[sparse_iter]
+            else:
+                assert curr_dq_write_order_full is not None
+                lock_value = curr_dq_write_order_full[sparse_iter - curr_q_cnt]
+        return lock_value
+
+    @cute.jit
+    def _release_dq_semaphore_if_in_bounds(
```
drisspg: inline this...

flash_attn/cute/block_sparsity.py
```diff
@@ -32,6 +36,138 @@ class BlockSparseTensorsTorch(NamedTuple):
...
+    valid = col_range[None, None, None, :] < num_blocks[:, :, :, None]
+    safe_indices = torch.where(valid, indices.long(), num_cols)
+    row_idx = torch.arange(num_rows, device=device)[None, None, :, None].expand_as(indices)
+    b_idx = torch.arange(B, device=device)[:, None, None, None].expand_as(indices)
+    h_idx = torch.arange(H, device=device)[None, :, None, None].expand_as(indices)
+    dense[b_idx, h_idx, row_idx, safe_indices] = 1
+    return dense[:, :, :, :num_cols]
+
+
+def compute_dq_write_order(
```
drisspg, line 72: can move this to PT call after we land there after this 😂

flash_attn/cute/block_sparsity.py
```diff
@@ -32,6 +36,138 @@ class BlockSparseTensorsTorch(NamedTuple):
     full_block_cnt: torch.Tensor | None = None
     full_block_idx: torch.Tensor | None = None
     block_size: tuple[int, int] | None = None
+    dq_write_order: torch.Tensor | None = None
+    dq_write_order_full: torch.Tensor | None = None
+    spt: bool | None = None
+
+
+def _ordered_to_dense_simple(
```
drisspg, line 44: ditto here

v0i0 (approved):

v0i0 (approved):

PR #2258, opened 2026-02-13
[Cute][Flex] Fix kernel hang w/ multiple empty tiles

drisspg:
[Cute][Flex] Fix kernel hang w/ multiple empty tiles

I think these werent surfaced because I didnt have any test that goes from empty row to empty row 

This fixes basically 3 bugs with our phasing related to empty rows

1. `handle_block_sparse_empty_tile_correction_sm100` for empty tiles we need to still flush the correct we were arriving on the rescaled P but the mma doesnt wait in the empty case
2. `softmax_block_sparse_sm100` for empty rows we were arrving on the rescaling bars again with no matching waiter, and for `mbar_softmax_corr_empty` we were arriving though we should only be waiting on and it gets arrived in the correction loop 
3. Always wait on `mbar_softmax_corr_empty_offset` if blocksparse since you can jump from row chunks that have work to ones that dont, and we want to ensure we arent flushing wrong values from previous iterations  
4. Also fixed a spurious phase flip `o_corr_consumer_phase` since for empty tiles again mma never arrives on this and correction never consumes (there is nothing from the mma warp to wait for) and if you had odd number of empty tiles then a tile with work the phase would have been flip flopped  and we hang

image

### DIFF
--- a/flash_attn/cute/block_sparse_utils.py
+++ b/flash_attn/cute/block_sparse_utils.py
@@ -18,6 +18,55 @@ from flash_attn.cute import copy_utils
 from flash_attn.cute.named_barrier import NamedBarrierBwd
 
 
+# NOTE [SM100 block-sparse empty tiles: mbarrier contract]
+#
+# For block-sparse SM100 forward, a given (m_block, stage) Q tile can have zero active
+# KV blocks (total_block_cnt == 0). In that case there is no seqlen_kv iteration, so
+# the softmax warp-group has no row stats to publish.
+#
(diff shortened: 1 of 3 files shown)


drisspg:
With this my fuzz tester for deterministic bwd passes for all tests;
image

tridao:
Do we have some convention here about which mbarriers are triggered regardless, and which mbarriers are triggered only if the tiles are non-empty? If so we should write it down explicitly as comments for next time.

tridao (approved):

drisspg:
This is a good call out, I'll document the semantics here before landing

PR #2260, opened 2026-02-14
[FA2] Fix int32 overflow

drisspg:
Fix int32 overflow

TLDR still possible to overflow since multiplication is done in `int` before index_t cast

No appreciable perf delta ( shouldn't be since we are fixing a cast)

Small repro:

```Py

@pytest.mark.parametrize("dtype", [torch.bfloat16])
def test_flash_attn_bwd_large_batch_no_overflow(dtype):
    if torch.cuda.get_device_capability("cuda")[0] < 8:
        pytest.skip()
    if torch.cuda.get_device_properties("cuda").total_memory < 24 * 2**30:
        pytest.skip()
    device = "cuda"
    torch.random.manual_seed(0)
    batch_size = 8300
    seqlen = 1
    nheads = 16
    d = 128
    q = torch.randn(
        batch_size, seqlen, nheads, d, device=device, dtype=dtype, requires_grad=True
    )
    k = torch.randn(
        batch_size, seqlen, nheads, d, device=device, dtype=dtype, requires_grad=True
    )
    v = torch.randn(
        batch_size, seqlen, nheads, d, device=device, dtype=dtype, requires_grad=True
    )
    out = flash_attn_func(q, k, v, dropout_p=0.0, causal=False)
    out.sum().backward()
    assert torch.isfinite(q.grad).all()
    assert torch.isfinite(k.grad).all()
    assert torch.isfinite(v.grad).all()

```

But didnt add since its a lil big in memory and feels like a one off

image

### DIFF
--- a/csrc/flash_attn/src/flash_bwd_kernel.h
+++ b/csrc/flash_attn/src/flash_bwd_kernel.h
@@ -117,8 +117,10 @@ inline __device__ void compute_dq_dk_dv_1colblock(const Params &params, const in
         + (m_block_max - 1) * kBlockM * params.o_row_stride + bidh * params.o_head_stride;
     const index_t row_offset_dq = binfo.q_offset(params.dq_batch_stride, params.dq_row_stride, bidb)
         + (m_block_max - 1) * kBlockM * params.dq_row_stride + bidh * params.dq_head_stride;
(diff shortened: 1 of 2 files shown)


tridao:
This makes sense, strides should generally be in int64. Thanks for fixing it.

PR #2270, opened 2026-02-24
[Cute,Sm100,Bwd] Add hdim 192 hdimv 128 backward for sm100

jayhshah:
Tested against all combinations of the following:

- MHA or MQA/GQA
- causal True or False
- deterministic True or False. 10000 iterations for bitwise equality with determinism 
- varlen: cu_seqlens_{q,k} x seqused_{q,k} + maybe zero_lengths_{q,k}

Speed looks good for a first version:

```
### headdim = 192, causal = False, seqlen = 8192, batch_size = 4, nheads = 16, nheads_kv = 16, varlen = False ###
use_2cta_instrs =  True
FA Python fwd: 1.767ms, 1555.7 TFLOPS
FA Python bwd: 6.406ms, 1072.8 TFLOPS
[deterministic]
FA Python bwd: 6.671ms, 1030.1 TFLOPS

### headdim = 192, causal = True, seqlen = 8192, batch_size = 4, nheads = 16, nheads_kv = 16, varlen = False ###
use_2cta_instrs =  True
FA Python fwd: 0.920ms, 1494.1 TFLOPS
FA Python bwd: 3.538ms, 971.1 TFLOPS
[deterministic]
FA Python bwd: 3.813ms, 901.1 TFLOPS
```

Local currently disabled for 2cta path, will fix this later along with hdim 128.

### DIFF
--- a/tests/cute/test_flash_attn.py
+++ b/tests/cute/test_flash_attn.py
@@ -119,7 +119,7 @@ def test_flash_attn_output(
     dtype_ref = torch.bfloat16 if dtype == torch.float8_e4m3fn else dtype
     # dv_vals = [128, d] if d > 128 and d <= 192 else ([256, 512, d] if d <= 64 else [d])
     dv_vals = [128] if d == 192 else ([d] if d != 128 else [64, d])
-    if dtype == torch.float8_e4m3fn or TEST_BWD_ONLY:
+    if dtype == torch.float8_e4m3fn:
         dv_vals = [d]
     # attention_chunk_vals = [torch.randint(1, seqlen_k * 2, (1,)).item(), 0]
     attention_chunk_vals = [0]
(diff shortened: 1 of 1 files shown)


flash_attn/cute/flash_bwd_preprocess.py
```diff
@@ -29,6 +29,7 @@ def __init__(
         self,
         dtype: Type[cutlass.Numeric],
         head_dim: int,
+        head_dim_v: int,
```
tridao, line 32: I was thinking for the preprocess kernel we just call the hdim of O and dO to be "head_dim" instead of "head_dim_v"?

tridao: do we ever need both head_dim and head_dim_v in this kernel?

jayhshah: I was thinking for clearing dQaccum we need head_dim

tridao: I see. Sounds great, then we need both head_dim and head_dim_v. Code is good as is.

PR #2271, opened 2026-02-25
[cute] Add return_lse

erikwijmans:
Adds `return_lse` as an explicit parameter to the cute interface. Currently the lse tensor is only not `None` if one of q/k/v requires grad.

Question: Should this be called `return_attn_probs` to fit with the FA2/FA3 naming of this parameter or `return_lse` to fit with existing name in `_flash_attn_fwd` for cute?

### DIFF
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -133,8 +133,8 @@ def _flash_attn_fwd(
         score_mod: A callable that takes the attention scores and applies a modification.
         mask_mod: A callable that takes token position information and selectively masks
         block_sparse_tensors: A tuple of tensors used for block sparsity.
-        return_lse: Whether to return the log softmax of the attention scores.
-            Note: the returned LSE currently does not support taking gradient. If set to True will always calculate
+        return_lse: Whether to return the log softmax of the attention scores. If set to True will always calculate
+            Note: the returned LSE currently does not support taking gradient.
         out: Optional pre-allocated output tensor. If None, will be allocated internally.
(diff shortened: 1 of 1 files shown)


flash_attn/cute/interface.py
```diff
@@ -1277,6 +1277,7 @@ def forward(
         mask_block_cnt: Optional[torch.Tensor] = None,
         mask_block_idx: Optional[torch.Tensor] = None,
         block_size: Optional[Tuple[int, int]] = None,
+        return_lse: bool = False,
```
tridao, line 1281: we should document that the returned LSE currently doesn't support taking gradient. And we should mark it as nondifferentiable for now.
We'll add support for LSE gradient later

flash_attn/cute/interface.py
```diff
@@ -1328,7 +1330,7 @@ def backward(ctx, dout, *args):
             window_size_right=ctx.window_size[1],
             deterministic=ctx.deterministic,
         )
-        return dq, dk, dv, *((None,) * 20)  # Extra Nones is fine
+        return dq, dk, dv, *((None,) * 21)  # Extra Nones is fine
```
tridao: returning 20 Nones is ok

tridao:
Thanks! `return_lse` is better. `return_attn_probs` was a confusing name (since we don't actually return the attention probabilities).

flash_attn/cute/interface.py
erikwijmans: Done!

PR #2273, opened 2026-02-25
Correct Cutlass Error Handling

ankutalev:
Hello!
There are some wrong handling of cutlass calls - currently, there are handle by `cudaGetLastError`, but this is not always correct - sometimes such calls can return cutlass errors like `cutlass::Status::kInternal` even without actual calls to cuda kernels.

This MR provides fix for that.

P.S. Thank you for your awesome work!


tridao:
Thanks!

Issue #2278, opened 2026-02-26
[Cute][Test] unittest `test_flash_attn_kvcache` hangs on B200


Alkaid-Benetnash:
To reproduce, for example:
```
pytest -x 'tests/cute/test_flash_attn.py::test_flash_attn_kvcache[1-128-64-False-False-False-1-0.0-True-False-False-False-False-False-False-mha-dtype0]'
```

Symptom: GPU sm_util 100%, power usage remain at idle level, cpu-side stuck at cudaStreamSynchronize.

PR #2188, opened 2026-01-16
[Cute][Testing] Protyping a fast test mode for Cute

drisspg:
[Cute][Testing] Protyping a fast test mode for Cute

Kind of exploring here;
Problem: Running full blackwell test suite takes ~6hours. Makes it hard to validate fully on your changes.

This adds a new test mode; `FLASH_ATTN_FAST_TESTS=1 pytest tests/cute` runs in about 15 mins and hopefully touches all major features.

## Usage

```bash
# Full suite
pytest tests/cute

# Fast mode
FLASH_ATTN_FAST_TESTS=1 pytest tests/cute

# Smoke mode (minimal)
FLASH_ATTN_SMOKE_TESTS=1 pytest tests/cute
```

## Collection counts

| Mode | Tests |
|------|------:|
| FULL | 163,457 |
| FAST | 4,640 |
| SMOKE | 583 |

## Coverage by file

| File | SMOKE | FAST |
|------|------:|-----:|
| `test_flash_attn.py` | 29 | 2,241 |
| `test_flash_attn_varlen.py` | 192 | 384 |
| `test_flash_attn_race_condition.py` | 0 | 1,168 |
| `test_score_mod_varlen.py` | 192 | 480 |
| `test_block_sparsity.py` | 113 | 295 |
| `test_mask_mod.py` | 25 | 25 |
| `test_score_mod.py` | 22 | 37 |
| `test_utils.py` | 10 | 10 |

### SMOKE mode (583 tests)

| Test | Count |
|------|------:|
| `test_varlen` | 192 |
| `test_varlen_score_mod_with_paged_kvcache_global` | 64 |
| `test_varlen_score_mod_kvcache` | 64 |
| `test_parameterized_masks` | 64 |
| `test_fast_sampling` | 40 |
| `test_varlen_with_score_mod` | 32 |
| `test_varlen_with_global_idx_score_mod` | 32 |
| `test_q_boundary_masking_block_sparse_bwd` | 14 |
| `test_flash_attn_output` | 12 |
| `test_score_mod_with_paged_kvcache` | 8 |
| `test_fixed_length_masks` | 8 |
| `test_cute_vs_flex_attention` | 8 |
| `test_flash_attn_kvcache` | 6 |
| `test_edge_cases` | 5 |
| `test_flash_attn_combine_fast` | 4 |
| `test_flash_attn_bwd_preallocated_outputs` | 4 |
| `test_flash_attn_varlen_output_fast` | 3 |
| Other (1-2 each) | 17 |

[...]

PR #2279, opened 2026-02-26
[Cute][Testing] Accelerate cutedsl fa4 tests with parallel compile and persistent compile cache

Alkaid-Benetnash:
### Split this PR to one sub-PR per commit:
- [x] #2283
- [x] #2304
- [x] #2311
- ~~Gave up: Once cutedsl upstream fixed their bugs, enable caching by default~~

### Overview of changes
**The proposed new two-pass test workflow (full sweep in ~25min):**

Step.1: pre-compile tests in parallel with pytest-xdist (e.g., 256 workers)

```
FLASH_ATTENTION_FAKE_TENSOR=1 FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1 pytest -n 256 -x tests/cute/test_flash_attn.py
```

Compiled binaries will be dumped to `/tmp/${USER}/flash_attention_cute_dsl_cache/` by default

On a B200 machine, I got ~43 compiles/sec:
> 30044 passed, 16128 skipped, 413456 warnings in 699.09s (0:11:39)

Step.2: run tests without `cute.compile`
```
FLASH_ATTENTION_FAKE_TENSOR=0 FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1 pytest -x tests/cute/test_flash_attn.py -k 'not test_flash_attn_kvcache'
```
Skipping test_flash_attn_kvcache because it currently has [GPU hangs](https://github.com/Dao-AILab/flash-attention/issues/2278). On the same B200 machine, with single-GPU-single-worker, I got ~36 passed/sec:
> 27548 passed, 15552 skipped, 3072 deselected, 553 warnings in 767.08s (0:12:47)

PS: with `pytest -n 8`, full sweep with all cached compilation would only take 2min
> 27548 passed, 15552 skipped, 1183 warnings in 126.37s (0:02:06)

The new introduced envvar flags are disabled by default. For comparison, the speed of the default/existing single-GPU test is ~0.5 passed/sec:
> 1488 passed, 891 skipped, 3072 deselected, 18113 warnings in 3125.21s (0:52:05) 

**How it works:**

- Use [torch fake tensor mode](https://docs.pytorch.org/docs/stable/user_guide/torch_compiler/torch.compiler_fake_tensor.html#api-the-important-bits) to enable cute.compile but don't allocate GPU memory or run kernels.
- Changes are made throughout `test_flash_attn.py` and `flash_attn/cute/interface.py` to bypass certain "data-dependent" operations when inside fake tensor mode.
- Use cutedsl [ahead-of-time compilation](https://docs.nvidia.com/cutlass/latest/media/docs/pythonDSL/cute_dsl_general/dsl_ahead_of_time_compilation.html#export-interface) support to export and load compiled modules to/from external storage.

> [!IMPORTANT]
> To make the above workflow work, I had to patch the following cutedsl bugs:
> - leaked MLIR threading pool: [cutlass#3062](https://github.com/NVIDIA/cutlass/pull/3062)
> - ahead-of-time compiling not loading shared library correctly: [cutlass#3075](https://github.com/NVIDIA/cutlass/pull/3075)

> [!NOTE]
>  As of cutedsl 4.4.0, cutedsl does not support exporting compiled tvm_ffi kernels as "shared library" (i.e., [only support object files](https://github.com/NVIDIA/cutlass/blob/057635de5c493a34e9913e2c5e836bdd98c58861/python/CuTeDSL/cutlass/cutlass_dsl/tvm_ffi_provider.py#L447)), but tvm_ffi only supports shared library. So my current workaround is an explicit `ld -shared -o foo.so foo.o`.

cutedsl patches attached bellow for your convenience (`cd site-packages && patch -p1 < xxx`):
```
--- a/nvidia_cutlass_dsl/python_packages/cutlass/base_dsl/dsl.py 2026-02-25 17:32:03.674296431 -0800
+++ b/nvidia_cutlass_dsl/python_packages/cutlass/base_dsl/dsl.py 2026-02-23 23:42:18.111766576 -0800
@@ -1345,7 +1345,8 @@
         location=None,
     ):
         """Generate MLIR module and compile iself.T_provider."""
-        with ir.Context(), self.get_ir_location(location):
+        with ir.Context() as ctx, self.get_ir_location(location):
+            ctx.enable_multithreading(False)
             try:
                 # Convert input arguments to MLIR arguments
                 exe_args, func_types, adapted_args = self.generate_mlir_function_types(
--- a/nvidia_cutlass_dsl/python_packages/cutlass/cute/runtime.py 2026-02-25 17:32:03.678703799 -0800
+++ b/nvidia_cutlass_dsl/python_packages/cutlass/cute/runtime.py 2026-02-25 17:28:11.608996616 -0800
@@ -951,7 +951,7 @@
         # no need to load tvm_ffi library here since it will be loaded by tvm_ffi package.
         for path in find_runtime_libraries(enable_tvm_ffi=False):
             if Path(path).exists():
-                _LOAD_MODULE_LIBS_CACHE.append(ctypes.CDLL(path))
+                _LOAD_MODULE_LIBS_CACHE.append(ctypes.CDLL(path, mode=ctypes.RTLD_GLOBAL))

     if enable_tvm_ffi:
         import tvm_ffi
```

**Other minor changes**

Added statistics printing to breakdown number of tests by test name:
```
pytest --co -qq tests/cute/test_flash_attn.py
```
Would print stats like:
```
cute/test_flash_attn.py: 46172
test_counts={'test_flash_attn_output': 2304, 'test_flash_attn_varlen_output': 39168, 'test_flash_attn_kvcache': 3072, 'test_flash_attn_bwd_preallocated_outputs': 8, 'test_flash_attn_combine': 1620}
```

General QoL w.r.t. `pytest -n N`
- Add per-worker logging at `/tmp/${USER}/flash_attention_tests/tests_gw${worker_id}.log`
- Only let one worker to query nvidia-smi and cache that results for other workers. When spawning many workers initializing pytorch, nvidia-smi becomes extremely slow and always timeout.

### DIFF
--- a/flash_attn/cute/cache_utils.py
+++ b/flash_attn/cute/cache_utils.py
@@ -1,0 +1,233 @@
+# Manage Ahead-of-Time (AOT) compiled kernels
+import fcntl
+import hashlib
+import os
+import pickle
+import tempfile
+import time
+from distutils.ccompiler import CCompiler, new_compiler
+from getpass import getuser
+from pathlib import Path
+from typing import Hashable, Optional, TypeAlias
+
+import cutlass.cute as cute
+import tvm_ffi
+from cutlass.cutlass_dsl import JitCompiledFunction
+
+CompileKeyType: TypeAlias = tuple[Hashable, ...]
+CallableFunction: TypeAlias = JitCompiledFunction | tvm_ffi.Function
+
+
+# Enable cache via `FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1`
+def is_cache_enabled() -> bool:
+    return int(os.getenv("FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED", 0)) == 1
+
+
+# Customize cache dir via `FLASH_ATTENTION_CUTE_DSL_CACHE_DIR`, default is
+# `/tmp/${USER}/flash_attention_cute_dsl_cache``
+def get_cache_path() -> Path:
+    default_cache_path = os.getenv("FLASH_ATTENTION_CUTE_DSL_CACHE_DIR", None)
+    if default_cache_path is None:
+        tmp_dir = Path(os.environ.get("TMPDIR", tempfile.gettempdir()))
+        cache_path: Path = tmp_dir / getuser() / "flash_attention_cute_dsl_cache"
+    else:
+        cache_path: Path = Path(default_cache_path)
+    cache_path.mkdir(parents=True, exist_ok=True)
+    return cache_path
+
+
+class FileLock:
+    """Context manager for advisory file locks using fcntl.flock.
+
+    Supports exclusive (write) and shared (read) locks.
+    Always blocks with polling until the lock is acquired or timeout is reached.
+
+    Usage:
+        with FileLock(lock_path, exclusive=True, timeout=15, label="abc"):
+            # do work under lock
+    """
+
+    def __init__(
+        self,
+        lock_path: Path,
+        exclusive: bool,
+        timeout: float = 15,
+        label: str = "",
+    ):
+        """
+        Args:
+            lock_path: Path to the lock file on disk.
+            exclusive: True for exclusive (write) lock, False for shared (read) lock.
+            timeout: Max seconds to wait for lock acquisition before raising RuntimeError.
+            label: Optional human-readable label for error messages.
+        """
+        self.lock_path: Path = lock_path
+        self.exclusive: bool = exclusive
+        self.timeout: float = timeout
+        self.label: str = label
+        self._fd: int = -1
+
+    @property
+    def _lock_label(self) -> str:
+        kind = "exclusive" if self.exclusive else "shared"
+        return f"{kind} {self.label}" if self.label else kind
+
+    def __enter__(self) -> "FileLock":
+        open_flags = (
+            os.O_WRONLY | os.O_CREAT if self.exclusive else os.O_RDONLY | os.O_CREAT
+        )
+        lock_type = fcntl.LOCK_EX if self.exclusive else fcntl.LOCK_SH
+
+        self._fd = os.open(str(self.lock_path), open_flags)
+
+        deadline = time.monotonic() + self.timeout
+        acquired = False
+        while time.monotonic() < deadline:
+            try:
+                fcntl.flock(self._fd, lock_type | fcntl.LOCK_NB)
+                acquired = True
+                break
+            except OSError:
+                time.sleep(0.1)
+        if not acquired:
+            os.close(self._fd)
+            self._fd = None
+            raise RuntimeError(
+                f"Timed out after {self.timeout}s waiting for "
+                f"{self._lock_label} lock: {self.lock_path}"
+            )
+
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb) -> None:
+        if self._fd is not None:
+            fcntl.flock(self._fd, fcntl.LOCK_UN)
+            os.close(self._fd)
+            self._fd = None
+
+
+class JITCache:
+    """
+    In-memory cache for compiled functions.
+    """
+
+    def __init__(self):
+        self.cache: dict[CompileKeyType, CallableFunction] = {}
+
+    def __setitem__(self, key: CompileKeyType, fn: JitCompiledFunction) -> None:
+        self.cache[key] = fn
+
+    def __getitem__(self, key: CompileKeyType) -> CallableFunction:
+        return self.cache[key]
+
+    def __contains__(self, key: CompileKeyType) -> bool:
+        return key in self.cache
+
+
+class JITPersistentCache(JITCache):
+    """
+    In-memory cache for compiled functions, which is also backed by persistent storage.
+    Use cutedsl ahead-of-time (AOT) compilation.
+    """
+
+    EXPORT_FUNCTION_PREFIX = "func"
+    LOCK_TIMEOUT_SECONDS = 15
+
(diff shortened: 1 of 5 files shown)


Alkaid-Benetnash:
cc @drisspg regarding #2188

drisspg (commented):
FWIW I really like this direction, https://docs.nvidia.com/cutlass/latest/media/docs/pythonDSL/cute_dsl_general/compile_with_tvm_ffi.html#fake-tensor-for-compilation

How would this look if we fully switched over to fake tensors for compilation and then real tensors for execution?
 
It would be great to decouple somethings where possible:
1. Can we switch over to fake tensors for compilation
2. Switch from Dict to a JitCache that supports optional disk caching
3. Update conftest to basically always use the new diskbased cache

Alkaid-Benetnash:
@drisspg 
> How would this look if we fully switched over to [fake tensors for compilation](https://docs.nvidia.com/cutlass/latest/media/docs/pythonDSL/cute_dsl_general/compile_with_tvm_ffi.html#fake-tensor-for-compilation) and then real tensors for execution?

To clarify/disambiguate, fake tensors now refer to two things: torch-fake tensors (`torch._subclasses.fake_tensor`); and cutedsl-fake tensors (`cute.runtime.make_fake_compact_tensor`).

What has been implemented in this PR uses **torch-fake tensors**; so it reuses as much existing logic as possible, in both test cases construction as well as `interface.py` plumbing before passing to `cute.compile`. This is possible because pytorch already has a lot of fake-tensor-aware wrappers for common operations (with data-dependent operators being the exception).

If switching to use cutedsl-fake tensors for compilation, my top concern would be the burden of maintaining two consistent code paths: one processing torch tensors with real data; the other construct cutedsl-fake tensors for compilation.

And I can't think of any benefit cutedsl-fake tensors would add on top of torch-fake tensors, TBH.

> It would be great to decouple somethings where possible:

You mean decouple these functionalities to be independent?
Or asking splitting commits to smaller ones?

Assuming the topic being "make functionalities independent", let me comment on the this PR's behavior:

> Can we switch over to fake tensors for compilation

An envvar flag `FLASH_ATTENTION_FAKE_TENSOR=0/1` controls whether we wrap each test case into a "always use torch-fake tensor" context manager.
The caching policies do not assume anything about how the JIT function is compiled.

> Switch from Dict to a JitCache that supports optional disk caching

envvar flag `FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=0/1` controls whether we use an in-memory cache (class name `JITCache`, mimicking existing behavior) or use an disk-backed cache (class name `JITPersistentCache`).
Both of them have the same dict-alike APIs, so `interface.py` makes no assumptions on the caching implementation.

> Update conftest to basically always use the new diskbased cache

Yes, this can be easily done via setting envvar in conftest.

drisspg:
I am saying that this PR should be split up and that I think there are 3 features here, fake tensor compilation, in-memory/disk multithreaded cache, and porting over tests 

And I think we should update in that order. My question about fake tensor is that I know we have fake tensor mode in Pytorch but I don't know how this interacts with dlpack/tvm and if this is expected to work and will stay working or it happens to work. We shud get clarity from cutlass folks on this

tridao:
I like this direction as well. I agree w @drisspg that we should get the fake tensor compilation in first.

Alkaid-Benetnash:
Hi @tridao @drisspg,

Understood. And hopefully I got your intention correctly.

I have split this PR to three smaller commits, as what was suggested above.
I will be creating separate PRs for each so you can review and merge step-by-step.

Check the updated PR description for the first step, fake tensor support.

flash_attn/cute/cache_utils.py
```diff
@@ -0,0 +1,233 @@
...
+class JITPersistentCache(JITCache):
+    """
+    In-memory cache for compiled functions, which is also backed by persistent storage.
+    Use cutedsl ahead-of-time (AOT) compilation.
+    """
+
+    EXPORT_FUNCTION_PREFIX = "func"
+    LOCK_TIMEOUT_SECONDS = 15
+
+    _compiler: CCompiler = new_compiler()
```
drisspg, line 136: is this supposed t be on class definition?

flash_attn/cute/cache_utils.py
```diff
@@ -0,0 +1,233 @@
...
+            )
+
+    def _key_to_hash(self, key: CompileKeyType) -> str:
+        return hashlib.sha256(pickle.dumps(key)).hexdigest()
+
+    def _lock_path(self, sha256_hex: str) -> Path:
+        return self.cache_path / f"{sha256_hex}.lock"
+
+
+def get_jit_cache(name: Optional[str] = None) -> JITCache:
```
drisspg, line 222: Super nit: str | None

flash_attn/cute/cache_utils.py
```diff
@@ -0,0 +1,233 @@
...
+                return
+            obj_path = self.cache_path / f"{sha256_hex}.o"
+            fn.export_to_c(
+                object_file_path=str(obj_path),
+                function_name=self.EXPORT_FUNCTION_PREFIX,
+            )
+            # TODO: as of cutedsl 4.4.0, `export_to_c` only supports exporting
+            # "relocatable" .o files. But tvm_ffi expects "shared library" .so
+            # files. Link ourselves to workaround.
+            JITPersistentCache._compiler.link_shared_object(
```
drisspg, line 211: should we then remove the object files?

Alkaid-Benetnash:
Closing this. All functionality PRs have landed a long time ago.

PR #2283, opened 2026-02-27
[Cute][Testing] Add fake tensor mode support for compile-only test passes

Alkaid-Benetnash:
Splitting main PR #2279 to smaller ones.

Use torch FakeTensorMode to enable cute.compile without allocating GPU memory or running kernels. This allows a fast pre-compilation pass (e.g., with pytest-xdist parallelism) to populate the compile cache.

Changes:
- Add `maybe_fake_tensor_mode` decorator and `is_fake_mode` helper to `testing.py`
- Guard kernel execution calls in `interface.py` with `is_fake_mode()` so compilation happens but kernels are skipped in fake mode
- Guard data-dependent operations in tests (torch.nonzero, .item(), torch.randint, reference computations, assertions) that are unsupported or unnecessary in FakeTensorMode

The new envvar flag `FLASH_ATTENTION_FAKE_TENSOR=1` is disabled by default.

### DIFF
--- a/tests/cute/test_flash_attn.py
+++ b/tests/cute/test_flash_attn.py
@@ -3,6 +3,7 @@
 import math
 import itertools
 import os
+import random
 
 import pytest
 import torch
@@ -113,7 +114,9 @@ def test_flash_attn_output(
         pytest.skip()
     device = "cuda"
     # set seed
-    torch.random.manual_seed(0)
+    seed = 0
+    random.seed(seed)
+    torch.random.manual_seed(seed)
     torch.cuda.empty_cache()
     torch.cuda.synchronize()
     batch_size = 9 if seqlen_k <= 2048 else 2
@@ -163,14 +166,9 @@ def test_flash_attn_output(
         else:
             qv_ref = None
         # Put window_size after QKV randn so that window_size changes from test to test
-        if not is_fake_mode():
-            window_size = (
-                (None, None) if not local else torch.randint(0, seqlen_k, (2,)).tolist()
-            )
-        else:
-            window_size = (
-                (None, None) if not local else (seqlen_k, seqlen_k)
-            )
+        window_size = (
+            (None, None) if not local else tuple(random.randrange(0, seqlen_k) for _ in range(2))
+        )
         if local_enum == 2:
             window_size = (None, -window_size[1])
         elif local_enum == 3:
@@ -483,7 +481,9 @@ def test_flash_attn_varlen_output(
         seqlen_k = seqlen_q
     device = "cuda"
     # set seed
-    torch.random.manual_seed(seqlen_q + seqlen_k + d + int(causal) * 2 + int(local))
+    seed = seqlen_q + seqlen_k + d + int(causal) * 2 + int(local)
+    random.seed(seed)
+    torch.random.manual_seed(seed)
     batch_size = 49 if seqlen_q <= 1024 else 7
     nheads = 6
     # nheads = 1
@@ -530,14 +530,9 @@ def test_flash_attn_varlen_output(
         else:
             qv_ref = None
         # Put window_size after QKV randn so that window_size changes from test to test
-        if not is_fake_mode():
-            window_size = (
-                (None, None) if not local else torch.randint(0, seqlen_k, (2,)).tolist()
-            )
-        else:
-            window_size = (
-                (None, None) if not local else (seqlen_k, seqlen_k)
-            )
+        window_size = (
+            (None, None) if not local else tuple(random.randrange(0, seqlen_k) for _ in range(2))
+        )
         if local_enum == 2:
             window_size = (None, window_size[1])
         elif local_enum == 3:
@@ -956,7 +951,9 @@ def test_flash_attn_kvcache(
         pytest.skip()
     device = "cuda"
     # set seed
-    torch.random.manual_seed(0)
+    seed = 0
+    random.seed(seed)
+    torch.random.manual_seed(seed)
     batch_size = 5
     # batch_size = 1
     batch_size_cache = batch_size if not has_batch_idx else batch_size * 2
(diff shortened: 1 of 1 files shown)


Alkaid-Benetnash:
To test this commit standalone:
```
CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 FLASH_ATTENTION_FAKE_TENSOR=1 TORCH_SHOW_CPP_STACKTRACES=1 pytest -n 256 -x tests/cute/test_flash_attn.py
```
Explicitly set `CUDA_VISIBLE_DEVICES` is a must because the default "query topology via nvidia-smi" will timeout due to massive number of pytorch initializing. This will be mitigated by a topology cache in follow-up PRs (when we do the full test sweep end-to-end).

Expected behavior:
- 32 python processes running on each of the 8 GPUs, each cost <1GB HBM due to pytorch initialization, etc. No extra HBM allocations and sm_util == 0
- All tests pass --> just meaning the fake tensor compilation passes, no kernels were actually executed
- All compilation results still cached in-memory and will be lost after the test is complete. Persistent caching will be implemented in follow-up PRs.
- The time to compile all tests would be slower w/o a disk cache. Because different workers might see the same kernel signature and each compiling themselves.

drisspg:
For provenance; spoke offline -> want to confirm fake_tensor mode compilation is expected to work for a while with cutlass folks, and that some of the `is_fake_mode` gates in the tests could be removed if we switched over to non pytorch apis for randomeness that end up calling .item()

flash_attn/cute/interface.py
```diff
@@ -517,6 +518,9 @@ def _flash_attn_fwd(
             options="--enable-tvm-ffi",
         )
 
+    if is_fake_mode():
```
drisspg: I think this is a good place to leave a comment on the symantics of fake mode

flash_attn/cute/interface.py
```diff
@@ -877,17 +881,18 @@ def _flash_attn_bwd(
...
-        dout,
-        dpsum,
-        lse,
-        lse_log2,
-        dq_accum,
-        cu_seqlens_q,
-        seqused_q,
-        current_stream,
-    )
+    if not is_fake_mode():
```
drisspg, line 886: maybe something like we dont early return so that the delta and post proc kernel completion proceeds , and instead just skip runtime execution

tests/cute/test_flash_attn.py
```diff
@@ -158,9 +162,14 @@ def test_flash_attn_output(
         else:
             qv_ref = None
         # Put window_size after QKV randn so that window_size changes from test to test
-        window_size = (
-            (None, None) if not local else torch.randint(0, seqlen_k, (2,)).tolist()
-        )
+        if not is_fake_mode():
+            window_size = (
+                (None, None) if not local else torch.randint(0, seqlen_k, (2,)).tolist()
```
drisspg: nit could keep this cleaner if we use random.randrange(seqlen_k)..

tests/cute/test_flash_attn.py
```diff
@@ -182,41 +191,42 @@ def test_flash_attn_output(
...
-            v_descale=v_descale,
-            window_size=window_size,
-            attention_chunk=attention_chunk,
-            learnable_sink=learnable_sink,
-            softcap=softcap,
-            upcast=False,
-            reorder_ops=True,
-            intermediate_dtype=dtype if dtype == torch.float8_e4m3fn else None,
-        )
+        if not is_fake_mode():
```
drisspg: im a lil surprised these fails, are they not using just pytorch ops?

tests/cute/test_flash_attn.py
```diff
@@ -258,6 +269,8 @@ def test_flash_attn_output(
                 num_splits=num_splits,
                 deterministic=deterministic,
             )
+            if is_fake_mode():
```
drisspg, line 270: maybe leave a note that this is the last invocation of cute kerenls

tests/cute/test_flash_attn.py
```diff
@@ -510,9 +527,14 @@ def test_flash_attn_varlen_output(
         else:
             qv_ref = None
         # Put window_size after QKV randn so that window_size changes from test to test
-        window_size = (
-            (None, None) if not local else torch.randint(0, seqlen_k, (2,)).tolist()
-        )
+        if not is_fake_mode():
+            window_size = (
+                (None, None) if not local else torch.randint(0, seqlen_k, (2,)).tolist()
```
drisspg: ditto

Alkaid-Benetnash:
> cc @tridao left mostly nits, but otherwise the cute folks said fake_tensor subclass mode is expected to work - will leave final sign off to you

And here is [an "official" example](https://github.com/NVIDIA/cutlass/blob/b9847690c5838ac3d909ebc163ed16c388802485/examples/python/CuTeDSL/cute/torch_fake_tensor.py) regarding torch fake tensor added in cutedsl v4.3

tests/cute/test_flash_attn.py
Alkaid-Benetnash: Some of the pytorch ops are data-dependent, not supporting fake tensor. Let me try take a look at what would it cost to support fake mode inside attention_ref instead of skipping here in the test file.

drisspg: yeah more just curious here

flash_attn/cute/interface.py
Alkaid-Benetnash: Added new comments here.

flash_attn/cute/interface.py
Alkaid-Benetnash: Agree. The exact code here does not early return; but there were indeed fake mode early return elsewhere.

Now updated all fake mode special handling in `interface.py` to only skip the kernel invocation.
In the case of multiple kernels compiling and running in a single interface API, all of them will be populated in cache.

tests/cute/test_flash_attn.py
Alkaid-Benetnash: Updated to random.randrange

tests/cute/test_flash_attn.py
Alkaid-Benetnash: Checked each fake mode handling in the `test_flash_attn.py`.
The attention reference impl already supports fake mode out of the box.
Only the postprocess accuracy analysis not working with fake mode. I commented the attention ref previously because it would be deadcode.

But it makes sense to me to optimize for minimum fake mode predicates. This attention_ref is no longer skipped in fake mode in the updated PR.

tests/cute/test_flash_attn.py
Alkaid-Benetnash: I added a note to every `continue`:
> no more flash_attn cutedsl calls for the rest of the loop
> skip data-dependent postprocessing

Note that it is NOT the last invocation of cute kernels. It just skip the outer for-loop (not the entire rest of the test). There could be more cutedsl kernels to run in later for-loop iterations.

tests/cute/test_flash_attn.py
Alkaid-Benetnash: Updated as well.

Alkaid-Benetnash:
Due to how github/git works (there is no stack of PRs), I will wait until this PR getting merged then create follow-up commits as separate PRs.
In the meantime, feel free to review the follow-up commits in the main PR.

tridao:
Thank you!

PR #2285, opened 2026-02-28
[Cute,Sm100,Bwd] Static Persistent Kernel 

tzadouri:
Static Persistent Kernel for backward. This gives a boost for seqlen < 8K, especially for non-causal. Will update the same PR with dynamic persistent using CLC. 

* The synchronization between mma & compute_loop for pipeline_dP and pipeline_dQ is slightly altered to address deadlocks.
* Currently, the tile scheduler is using symmetric load balancing. 
* Also in compute_loop, for d <= 128, when copying into sdS_xchg, made the change to avoid using the run time exchange_stage (cta_rank_in_cluster ^ 1) in the inner branch

### DIFF
--- a/flash_attn/cute/tile_scheduler.py
+++ b/flash_attn/cute/tile_scheduler.py
@@ -18,7 +18,6 @@ from quack.cute_dsl_utils import ParamsBase
 
 import flash_attn.cute.utils as utils
 from flash_attn.cute.fast_math import clz
-import quack.utils as quack_utils
 import math
 
 
@@ -465,7 +464,6 @@ class CLCDynamicPersistentBWDTileScheduler:
         params: Params,
         sched_smem: cute.Tensor,
         scheduler_pipeline,
-        clc_phase: Int32,
         sched_state,
         *,
         loc=None,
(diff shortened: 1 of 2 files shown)


flash_attn/cute/tile_scheduler.py
```diff
@@ -248,6 +250,414 @@ def __new_from_mlir_values__(self, values):
...
+            (Int32(n_block), Int32(head_idx), Int32(batch_idx), Int32(0)),
+            True,
+        )
+
+    def prefetch_next_work(self, *, loc=None, ip=None):
+        pass
+
+    # producer (scheduler warp)
+    @cute.jit
+    def fetch_next_work(self, *, loc=None, ip=None):
```
Edenzzzz, line 555: I wonder if this is actually slower for causal, as LPT scheduling cannot be implemented with CLC?

tridao: We'll eventually add LPT to CLC as well

Edenzzzz: We will need to resort to `atomicAdd` in that case because `clusterlaunchcontrol.try_cancel` does not specify a cancel order?

PR #2292, opened 2026-02-28
Fix sm100 fwd missing tSrQs init regression

drisspg:
Stacked PRs:
 * __->__#2292
 * #2291

--- --- ---

Fix sm100 fwd missing tSrQs init regression

Main is broken with me, looks this was removed on accident


drisspg:
@tridao stack-pr caused this to basically get merged into the wrong branch and not main: 
https://github.com/Dao-AILab/flash-attention/pull/2293

reopend here

PR #2304, opened 2026-03-03
[Cute][Testing] Add persistent compile cache for cutedsl AOT compilation

Alkaid-Benetnash:
Splitting main PR #2279 to smaller ones.
Introduce `cache_utils.py` with dict-alike cache classes that support ahead-of-time (AOT) compiled kernel persistence:

- `JITCache`: in-memory compile cache (drop-in replacement for `{}`)
- `JITPersistentCache`: disk-backed cache using cutedsl `export_to_c` with file locking for concurrent multi-process safety
- `get_jit_cache()`: factory that returns persistent cache when `FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1`, plain in-memory otherwise

Replace bare `{}` compile caches in `interface.py` with `get_jit_cache()` calls so compiled kernels can be exported/loaded across test runs.

The envvar `FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED` is disabled by default.

> [!IMPORTANT]
> To make the above workflow work, I had to patch the following cutedsl bugs:
> - leaked MLIR threading pool: [cutlass#3062](https://github.com/NVIDIA/cutlass/pull/3062)
> - ahead-of-time compiling not loading shared library correctly: [cutlass#3075](https://github.com/NVIDIA/cutlass/pull/3075)

> [!NOTE]
>  As of cutedsl 4.4.0, cutedsl does not support exporting compiled tvm_ffi kernels as "shared library" (i.e., [only support object files](https://github.com/NVIDIA/cutlass/blob/057635de5c493a34e9913e2c5e836bdd98c58861/python/CuTeDSL/cutlass/cutlass_dsl/tvm_ffi_provider.py#L447)), but tvm_ffi only supports shared library. So my current workaround is an explicit `ld -shared -o foo.so foo.o`.

cutedsl folks acknowledged the above issues and said they are working on them. Not sure about their timeline.

### DIFF
--- a/flash_attn/cute/cache_utils.py
+++ b/flash_attn/cute/cache_utils.py
@@ -169,7 +169,7 @@ class JITCache:
 class JITPersistentCache(JITCache):
     """
     In-memory cache for compiled functions, which is also backed by persistent storage.
-    Use cutedsl ahead-of-time (AOT) compilation.
+    Use cutedsl ahead-of-time (AOT) compilation, only supporting enable_tvm_ffi=True
     """
 
     EXPORT_FUNCTION_PREFIX = "func"
@@ -177,10 +177,9 @@ class JITPersistentCache(JITCache):
 
     _compiler: CCompiler | None = None
 
-    def __init__(self, cache_path: Path, enable_tvm_ffi: bool = True):
+    def __init__(self, cache_path: Path):
         super().__init__()
         cache_path.mkdir(parents=True, exist_ok=True)
-        self.enable_tvm_ffi: bool = enable_tvm_ffi
         self.cache_path: Path = cache_path
 
     def __setitem__(self, key: CompileKeyType, fn: JitCompiledFunction) -> None:
@@ -206,38 +205,33 @@ class JITPersistentCache(JITCache):
         Holds a shared lock during loading to prevent concurrent writes.
         """
         sha256_hex = self._key_to_hash(key)
-        if self.enable_tvm_ffi:
-            so_path = self.cache_path / f"{sha256_hex}.so"
-            with FileLock(
-                self._lock_path(sha256_hex),
-                exclusive=False,
-                timeout=self.LOCK_TIMEOUT_SECONDS,
-                label=sha256_hex,
-            ):
-                if so_path.exists():
-                    logger.debug(
-                        "Loading compiled function from disk: %s", so_path
-                    )
-                    m = cute.runtime.load_module(
-                        str(so_path), enable_tvm_ffi=self.enable_tvm_ffi
-                    )
-                    fn = getattr(m, self.EXPORT_FUNCTION_PREFIX)
-                    JITCache.__setitem__(self, key, fn)
-                    return True
-                else:
-                    logger.debug(
-                        "Cache miss on disk for key hash %s", sha256_hex
-                    )
-        else:
-            raise NotImplementedError("non tvm_ffi path is not supported")
+        so_path = self.cache_path / f"{sha256_hex}.so"
+        with FileLock(
+            self._lock_path(sha256_hex),
+            exclusive=False,
+            timeout=self.LOCK_TIMEOUT_SECONDS,
+            label=sha256_hex,
+        ):
+            if so_path.exists():
+                logger.debug(
+                    "Loading compiled function from disk: %s", so_path
+                )
+                m = cute.runtime.load_module(
+                    str(so_path), enable_tvm_ffi=True
+                )
+                fn = getattr(m, self.EXPORT_FUNCTION_PREFIX)
+                JITCache.__setitem__(self, key, fn)
+                return True
+            else:
+                logger.debug(
+                    "Cache miss on disk for key hash %s", sha256_hex
+                )
         return False
 
     def _try_export_to_storage(
         self, key: CompileKeyType, fn: JitCompiledFunction
     ) -> None:
         """Export a compiled function to persistent storage under exclusive lock."""
-        if not self.enable_tvm_ffi:
-            raise NotImplementedError("non tvm_ffi path is not supported")
         sha256_hex = self._key_to_hash(key)
         with FileLock(
             self._lock_path(sha256_hex),


Alkaid-Benetnash:
To test this PR, for example:
Run
```
CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 FLASH_ATTENTION_FAKE_TENSOR=1 FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1 pytest -n 256 -x tests/cute/test_flash_attn.py
```
Then confirm cutedsl compilation cache at `/tmp/${USER}/flash_attention_cute_dsl_cache`.
Then run the test with cache on single-GPU:
```
FLASH_ATTENTION_FAKE_TENSOR=0 FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1 pytest -x tests/cute/test_flash_attn.py -k 'not test_flash_attn_kvcache'
```

As of the compilation cache of all tests inside `test_flash_attn.py`, it's around 400MB for sm100.

tridao:
Looks fine to me.
@drisspg what do you think?

flash_attn/cute/cache_utils.py
```diff
@@ -0,0 +1,233 @@
...
+import cutlass.cute as cute
+import tvm_ffi
+from cutlass.cutlass_dsl import JitCompiledFunction
+
+CompileKeyType: TypeAlias = tuple[Hashable, ...]
+CallableFunction: TypeAlias = JitCompiledFunction | tvm_ffi.Function
+
+
+# Enable cache via `FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1`
+def is_cache_enabled() -> bool:
```
drisspg: can we do `CUTE_DSL_CACHE_ENABLED: bool = os.getenv("CUTE_DSL_CACHE_ENABLED", "0") == "1"`

Its nice because you can also be set programmatically

flash_attn/cute/cache_utils.py
```diff
@@ -0,0 +1,233 @@
...
+    return int(os.getenv("FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED", 0)) == 1
+
+
+# Customize cache dir via `FLASH_ATTENTION_CUTE_DSL_CACHE_DIR`, default is
+# `/tmp/${USER}/flash_attention_cute_dsl_cache``
+def get_cache_path() -> Path:
+    default_cache_path = os.getenv("FLASH_ATTENTION_CUTE_DSL_CACHE_DIR", None)
```
drisspg: Maybe:

```Py
CUTE_DSL_CACHE_DIR: str | None = os.getenv("CUTE_DSL_CACHE_DIR")

def get_cache_path() -> Path:
    if CUTE_DSL_CACHE_DIR is not None:
        return Path(CUTE_DSL_CACHE_DIR)
    tmp_dir = Path(os.environ.get("TMPDIR", tempfile.gettempdir()))
    return tmp_dir / getuser() / "flash_attention_cute_dsl_cache"

```

what do you think? same deal let users set it programmatically if needed and not just env vars. We dont need the mkdir since its in the acutal disk cache constructor right?

flash_attn/cute/cache_utils.py
```diff
@@ -0,0 +1,232 @@
...
+class JITPersistentCache(JITCache):
+    """
+    In-memory cache for compiled functions, which is also backed by persistent storage.
+    Use cutedsl ahead-of-time (AOT) compilation.
+    """
+
+    EXPORT_FUNCTION_PREFIX = "func"
+    LOCK_TIMEOUT_SECONDS = 15
+
+    _compiler: CCompiler = new_compiler()
```
drisspg: is this supposed t be on class definition?

flash_attn/cute/cache_utils.py
```diff
@@ -0,0 +1,232 @@
...
+            so_path = self.cache_path / f"{sha256_hex}.so"
+            if so_path.exists():
+                # Another process already exported.
+                return
+            obj_path = self.cache_path / f"{sha256_hex}.o"
+            fn.export_to_c(
+                object_file_path=str(obj_path),
+                function_name=self.EXPORT_FUNCTION_PREFIX,
+            )
+            # TODO: as of cutedsl 4.4.0, `export_to_c` only supports exporting
```
drisspg, line 257: should we then remove the object files?

flash_attn/cute/cache_utils.py
```diff
@@ -0,0 +1,232 @@
...
+            )
+
+    def _key_to_hash(self, key: CompileKeyType) -> str:
+        return hashlib.sha256(pickle.dumps(key)).hexdigest()
+
+    def _lock_path(self, sha256_hex: str) -> Path:
+        return self.cache_path / f"{sha256_hex}.lock"
+
+
+def get_jit_cache(name: Optional[str] = None) -> JITCache:
```
drisspg: Super nit: str | None

flash_attn/cute/cache_utils.py
```diff
@@ -0,0 +1,232 @@
...
+        return self._try_load_from_storage(key)
+
+    def _try_load_from_storage(self, key: CompileKeyType) -> bool:
+        """
+        Try to load a function from persistent storage into in-memory cache.
+        Returns True if loaded successfully, False if not found on disk.
+        Holds a shared lock during loading to prevent concurrent writes.
+        """
+        sha256_hex = self._key_to_hash(key)
+        if self.enable_tvm_ffi:
```
drisspg: we probably dont need the tvm-ffi helper right? I dont think the interface even supports non tvm compilation anymore?

flash_attn/cute/cache_utils.py
```diff
@@ -0,0 +1,232 @@
...
+                self._lock_path(sha256_hex),
+                exclusive=False,
+                timeout=self.LOCK_TIMEOUT_SECONDS,
+                label=sha256_hex,
+            ):
+                if so_path.exists():
+                    m = cute.runtime.load_module(
+                        str(so_path), enable_tvm_ffi=self.enable_tvm_ffi
+                    )
+                    fn = getattr(m, self.EXPORT_FUNCTION_PREFIX)
```
drisspg: nit: does this need to be getattr?

flash_attn/cute/cache_utils.py
Alkaid-Benetnash: Done

flash_attn/cute/cache_utils.py
Alkaid-Benetnash: Agree. They were kept around for debugging; now no longer needed.

flash_attn/cute/cache_utils.py
Alkaid-Benetnash: Every usage of that compiler instance here share the same compiler flags, include paths, etc.
So I thought it won't hurt to always reuse the same instance.

That API is documented here https://setuptools.pypa.io/en/stable/deprecated/distutils/apiref.html#module-distutils.ccompiler
I don't see any cautions regarding isolating multiple instances for different compilations.

flash_attn/cute/cache_utils.py
Alkaid-Benetnash: Sure, I was thinking the `os.putenv(xxx)` would allow programmatically setting these flags.
But yeah, operating on python module-level bool variable would be more intuitive.

Updated both places.

For this cache_dir, a bit more clarification on changes:
- Have to make sure the cache dir exists; so still need `.mkdir`
- I just realized the `tempfile.gettempdir` would check all necessary envvars to decide a proper tmp_dir, so removing the redundant `os.env`.

flash_attn/cute/cache_utils.py
Alkaid-Benetnash: Done

flash_attn/cute/cache_utils.py
Alkaid-Benetnash: I am a bit confused. What do you refer as "tvm-ffi helper"? Yes, I agree that fa4 only supports tvm compilation at this moment.

flash_attn/cute/cache_utils.py
Alkaid-Benetnash: The non-getattr alternative would just hardcode the function name as attr too:
```
m.export_to_c(function_name="func")
# then
load_module().func
```

I personally prefer the `getattr` than hardcoding `.func`.

flash_attn/cute/cache_utils.py
drisspg: I think it just that you will init a a compilier even if you are not going to use it, also I think we need to add setuptools to the pyproject toml 

AFAIK torch has this dep so no need to pin versions but we should add

flash_attn/cute/cache_utils.py
drisspg: ahh you are right, didnt see func name was dynamic

flash_attn/cute/cache_utils.py
drisspg: I mean `if self.enable_tvm_ffi:` this will always be true

Alkaid-Benetnash:
@drisspg 
> should we hash all sources so that we can more easily invalidate disk keys?

Inside cutedsl, their caching is indeed more robust, copy-pasting [their doc](https://docs.nvidia.com/cutlass/latest/media/docs/pythonDSL/cute_dsl_general/dsl_jit_caching.html#cache-in-dsl):
```
The cache key combines hashes of:
- MLIR bytecode of the MLIR program generated by CuTe DSL
- All CuTe DSL Python source files
- All CuTe DSL shared libraries
- All CuTe DSL environment variables
```

I thought of reusing their logic, but `cute.compile` is a monolithic process from flash-attn's perspective. The important information ("ast-rewritten py src", "MLIR bytecode", etc.) is not exposed.
It doesn't make sense to me to reimplement such caching logic outside of cutedsl or use cutedsl's unexposed internal helpers.

For the purpose of accelerating testing, manually invalidate/rm disk caches sounds good enough. For really exposing the cache for end-users, yes we shall be more careful on the cache key selection.

flash_attn/cute/cache_utils.py
Alkaid-Benetnash: Yes, it is expected to always be true.
I left `enable_tvm_ffi` configurable (but raise exception) here mostly due to seeing other internal helpers not hardcoding `enable_tvm_ffi=True`. For example `to_cute_tensor`.

I am happy to remove this if branch and hardcode it to always be true. What's your preference on hardcoding or not?

flash_attn/cute/cache_utils.py
Alkaid-Benetnash: I see; then let me make it lazily initialized.

I will also add the setuptools to pyproject. I saw its in the `[build-system]` but I just realized that is different from runtime dependencies.

Alkaid-Benetnash: All done.

drisspg:
Also: /home/dev/.venvs/nightly/bin/pytest tests/cute/test_mask_mod.py::test_sm100_block_sparse_q_stage1 -q -x

this test is failing since we invoke dict(cache) to test that we get a new entry, https://github.com/Dao-AILab/flash-attention/blob/944e4574c6d3d972fe7ec3b77951bd51ae10f2b5/tests/cute/test_mask_mod.py#L821-L822 needs to be updated

flash_attn/cute/cache_utils.py
drisspg: yeah thats a good point and mostly because it was implemented before the switch over. I think know that we are full tvvm-ffi no need to add branching to keep things simple

Alkaid-Benetnash:
> Also: /home/dev/.venvs/nightly/bin/pytest tests/cute/test_mask_mod.py::test_sm100_block_sparse_q_stage1 -q -x
> 
> this test is failing since we invoke dict(cache) to test that we get a new entry,

Ah, I have been only focusing on `test_flash_attn.py`. Let me check `test_mask_mod.py` too. Shall we do `fake_tensor` + cached compilation to any other tests? (I mean eventually all important long-running tests would benefit from this, but what should be a good target for this first adoption?)

drisspg:
@Alkaid-Benetnash 
image
This is kind of stale but I think test_mask_mod, test_score_od and the varlen tests probably have the most compile units. Forsure test_flash_attn is most important

Alkaid-Benetnash:
Summary of new updates to address issues discussed above:
- Added fingerprinting based on python and cutedsl version, as well as `flash_attn/cute/*.py` source directory content hash (credits to @drisspg & claude). The compilation cache would be stored at `${FLASH_ATTENTION_CUTE_DSL_CACHE_DIR}/${fingerprint}/{fwd,bwd,...}/${hash(compile_key)}.so`
- Small change to `test_mask_mod.py`
  - Its pre-existing testing logic manipulates compile caches.
  - Added a `clear` method in the new cache classes (`JITCache` and `JITPersistentCache`) to support this test (and better mimicing `dict` API).
- Added debug prints to instrument persistent cache hit/miss/export/load, etc.
- Removed `enable_tvm_ffi=False` code path, since forward-looking flash-attn is expected to only work with tvm_ffi.
- Added "setuptools" as runtime dependency (not just build dependency) in `pyproject.toml`

Also tested all other tests (excluding the main `test_flash_attn.py`) under `tests/cute`, all work fine:
```
CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7 TORCH_SHOW_CPP_STACKTRACES=1 pytest -n 8 -x tests/cute/ --ignore=tests/cute/test_flash_attn.py
```

Alkaid-Benetnash:
Expected behavior after merging this PR:
- For users, calling fa4 is not expected to have any differences.
- For developers, tests will NOT run faster by default
  - Require expliclt patching cutedsl (patch in #2279) until upstream PR got merged/fixed.
  - Require explicit opt-in caching via `FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1`.
  - Require explicit opt-in ahead-of-time compilation via `FLASH_ATTENTION_FAKE_TENSOR=1`. Which only compiles `test_flash_attn.py` much faster (we can adopt this workflow in other tests later).

PR #2306, opened 2026-03-06
Fix comment in cutedsl sm100 backward

Edenzzzz:


tridao:
Our notation was not consistent, I think it should actually be dV.T = P.T @ dO

PR #2308, opened 2026-03-06
feat: ring attention for sequence-parallel long-context training

ghost:
Adds `ring_flash_attn_func`, a sequence-parallel attention primitive that shards Q/K/V along the sequence dimension across a process group and rotates KV chunks in a peer-to-peer ring. Partial `(out, lse)` results are merged via the same numerically-stable online-softmax identity that FlashAttention uses internally, so outputs are bit-exact with single-GPU `flash_attn_func` (verified in the included `torchrun` test).

**Why this matters:** enables training at sequence lengths that exceed single-GPU HBM (e.g., 128K+ tokens on 8×H100) with near-zero communication overhead on NVLink, since NCCL P2P transfers are issued asynchronously and overlap with the local CUDA kernel.

All building blocks (`softmax_lse`, `all_gather_raw`, `reduce_scatter_raw`) were already present; this PR adds the ring loop and merge kernel only. Backward pass raises `NotImplementedError` and is scoped to a follow-up; `torch.utils.checkpoint` is the recommended workaround for causal training today.

### DIFF
--- a/tests/test_ring_flash_attn.py
+++ b/tests/test_ring_flash_attn.py
@@ -1,0 +1,57 @@
+"""
+Correctness test for ring_flash_attn_func.
+
+Run with torchrun:
+    torchrun --nproc_per_node=2 tests/test_ring_flash_attn.py
+"""
+import os
+import torch
+import torch.distributed as dist
+
+from flash_attn import flash_attn_func, ring_flash_attn_func
+from flash_attn.utils.distributed import split_sequence_for_ring
+
+
+def main():
+    dist.init_process_group("nccl")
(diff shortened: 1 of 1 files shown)


ghost:
Hi @tridao and @drisspg,

Apologies for the direct ping! This is my first PR here, and I just wanted to make sure it didn't get missed in the shuffle.

Is there a specific procedure or a typical timeline for reviews and merging I should be aware of? I’m happy to follow whatever workflow you have in place.

Thanks for your time!

tridao:
hi, do you have benchmark numbers for the implementation?

PR #2312, opened 2026-03-06
[Cute,Bwd,Sm100] Fix tmem use-after-free: add within-CTA dealloc barrier

Alkaid-Benetnash:
The TmemAllocator commit (a79ee34) removed the mbarrier synchronization that prevented the MMA warp from deallocating tmem while compute warps were still using it.

TmemAllocator only provides cross-CTA sync (two_cta_tmem_dealloc_mbar_ptr, used between peer CTAs in 2-CTA mode) but has no within-CTA sync to ensure local consumer warps are done before deallocation.

Add a separate within_cta_tmem_dealloc_mbar_ptr for this purpose:
- Init barrier with arrival count = all compute warp threads
- Compute warps arrive after finishing with tmem
- MMA warp waits on the barrier before calling tmem.free()

Its logic is basically the same as what existed before the TmemAllocator commit.

Tested with:
```
export CUDA_VISIBLE_DEVICES=0,1,2,3,4,5,6,7
export FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1
# test_flash_attn.py excluding the 'test_flash_attn_kvcache' known to hang (#2278)
FLASH_ATTENTION_FAKE_TENSOR=1 pytest -n 256 -x tests/cute/test_flash_attn.py
FLASH_ATTENTION_FAKE_TENSOR=0 pytest -n 8 -x tests/cute/test_flash_attn.py -k 'not test_flash_attn_kvcache'
# 27548 passed, 15552 skipped, 1260 warnings in 182.04s (0:03:02)
# All other tests
pytest -n 8 -x tests/cute/ --ignore=tests/cute/test_flash_attn.py
# 22030 passed, 18971 skipped, 78 xfailed, 143188 warnings in 1222.08s (0:20:22)
```

Before: error "torch.AcceleratorError: CUDA error: unspecified launch failure" when running tests. The error was also non-deterministic (run the failing single test would pass; run the whole test suite would fail).
After: passing all tests.

### DIFF
--- a/flash_attn/cute/flash_bwd_sm100.py
+++ b/flash_attn/cute/flash_bwd_sm100.py
@@ -782,7 +782,8 @@ class FlashAttentionBackwardSm100:
                     cutlass.Int64, self.dQaccum_reduce_stage // 2
                 ]
                 tmem_holding_buf: Int32
-                tmem_dealloc_mbar_ptr: cutlass.Int64
+                two_cta_tmem_dealloc_mbar_ptr: cutlass.Int64
+                within_cta_tmem_dealloc_mbar_ptr: cutlass.Int64
 
                 # 2-CTA
                 Qt_mbar_ptr: cute.struct.MemRange[cutlass.Int64, 2 * self.Q_stage]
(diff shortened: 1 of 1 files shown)


Alkaid-Benetnash:
cc. @tridao
Discussed with @drisspg offline. Would require your expertise to review this change.

tridao:
Thanks! There was indeed a bug. I think we can just use a NamedBarrier to sync before dealloc, it's a bit simpler than mbarrier. I'll give that a try

tridao:
Can you try https://github.com/Dao-AILab/flash-attention/commit/dd8a2729a228b2292411da34a52784aa0fff06ef

Alkaid-Benetnash:
Hi, I can confirm that https://github.com/Dao-AILab/flash-attention/commit/dd8a2729a228b2292411da34a52784aa0fff06ef fixed this bug and passed all tests on my end.

Also agree that reusing the tmem allocator's existing NamedBarrier is better than yet another mbar.
Closing this PR.

PR #2320, opened 2026-03-09
Change 2cta opt in to have min seqlen > 2*m_block_size

drisspg:
Stacked PRs:
 * #2218
 * __->__#2320

--- --- ---

Change 2cta opt in to have min seqlen > 2*m_block_size

### DIFF
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -333,6 +333,20 @@ def _flash_attn_fwd(
         out_partial = torch.empty(num_splits, *q_batch_seqlen_shape, num_head, head_dim_v, dtype=torch.float32, device=device)
         lse_partial = torch.empty(num_splits, *lse_shape, dtype=torch.float32, device=device)
 
+    use_2cta_instrs = (
+        arch // 10 in [10, 11]
+        and not causal
+        and not local
+        and not is_split_kv
+        and cu_seqlens_q is None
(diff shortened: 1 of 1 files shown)


flash_attn/cute/interface.py
```diff
@@ -333,6 +333,20 @@ def _flash_attn_fwd(
...
+        and not causal
+        and not local
+        and not is_split_kv
+        and cu_seqlens_q is None
+        and seqused_q is None
+        and not use_block_sparsity
+        and page_size in [None, 128]
+        and int(math.ceil(head_dim / 16) * 16) == 128
+        and int(math.ceil(head_dim_v / 16) * 16) == 128
+        and seqlen_q_packgqa >= 2 * m_block_size
```
drisspg: This one is new, I dont think we want to enable < 256

drisspg: But correct me if im wrong

jayhshah: Nit: for 2cta mode and qstage=2 we cover 512 of M over the two ctas, so I would expect inequality should be `seqlen_q_packgqa > 2 * m_block_size`. Put another way, don't want to double grid size by using cluster size 2 if m_blocks = 1.

drisspg: Ahh yeah totally agree

PR #2324, opened 2026-03-10
[varlen] add autograd function to zero out nan padding

liangel-02:
**Summary**

in FA2 and FA3, gradients are initialized using torch.empty() and it uses the length of the `q, k, v`. but when using varlen attention where `q, k, v` are initialized with extra padding (ie the length > `cu_seqlens[-1]`), this results in invalid entries in the gradients remaining uninitialized which can lead to NaNs getting returned and propagating through model layers. 

as a fix, i've added an autograd function `no_grad_forward_slice_backward` and you can call as follows
 
```python
q = no_grad_forward_slice_backward(q, cu_seqlens_q[-1])
k = no_grad_forward_slice_backward(k, cu_seqlens_k[-1])
v = no_grad_forward_slice_backward(v, cu_seqlens_k[-1])
out = flash_attn_varlen_func(q, k, v, cu_seqlens_q, cu_seqlens_k, ...)
```
before you run `flash_attn_varlen_func` which zeroes out the extra padding. and since this is a user opt-in function, it doesn't add any additional unwanted overhead. 

**Testing**

i added `test_no_grad_forward_slice_backward_varlen` to both FA2 and FA3 which runs a reference forward, "poisons" the GPU with NaNs (so we guarantee dq, dk, dv are initialized with NaNs) and then i test functionality and correctness with `no_grad_forward_slice_backward`. i also add a negative test to first verify that we see NaNs in the gradients without the new autgrad function.

### DIFF
--- a/flash_attn/flash_attn_interface.py
+++ b/flash_attn/flash_attn_interface.py
@@ -219,7 +219,7 @@ def _flash_attn_varlen_forward_fake(
     paged_kv = block_table is not None
     batch_size = cu_seqlens_q.numel() - 1
     total_q, num_heads, _ = q.shape
-    
+
     out = torch.empty_like(q)
     softmax_lse = torch.empty((num_heads, total_q), dtype=torch.float32, device=q.device, layout=q.layout)
     p = torch.empty((0,), dtype=q.dtype, device=q.device, layout=q.layout)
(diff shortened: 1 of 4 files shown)


liangel-02:
cc @drisspg

liangel-02:
closing after speaking with @drisspg offline, to workaround this the user should zero out padding before the forward call

PR #2297, opened 2026-03-02
[CuTeDSL][Sm80] basic fix for new api

zhuochenKIDD:
fix CuTeDSL FA foward & backward basic API errors. Now FlashAttentionForwardSm80 & FlashAttentionBackwardSm80 can be enabled for sm80. 

Also there are many missing features under sm80, mentioned by https://github.com/Dao-AILab/flash-attention/pull/2103.

For Backward, in my local test, dim 128 works, but other dim like 32 & 64 need careful plan of AtomLayoutMSdP NdKV MdQ,  will provide more fix if needed

PR #2328, opened 2026-03-11
fix: duplicate softmax_scale param

NanoCode012:
Small regression from PR #2297

Caught this while trying FA4 on H100. Keeping the Optional one following other interface APIs.


NanoCode012:
@zhuochenKIDD tagging as related

tridao:
Thanks!

Issue #2310, opened 2026-03-06
FA4 Installation and Usage instructions both fail; ModuleNotFoundError: No module named 'flash_attn.cute'


Birch-san:
Hi, thanks for releasing FA4.

I tried following the readme. Installation fails:

```
pip install flash-attn-4
Defaulting to user installation because normal site-packages is not writeable
ERROR: Ignored the following yanked versions: 0.0.1
ERROR: Could not find a version that satisfies the requirement flash-attn-4 (from versions: 4.0.0b3, 4.0.0b4)

[notice] A new release of pip is available: 25.2 -> 26.0.1
[notice] To update, run: python3 -m pip install --upgrade pip
ERROR: No matching distribution found for flash-attn-4
```

So I specified an explicit version; this is fine:

```
pip install flash-attn-4==4.0.0b4
Installing collected packages: cuda-pathfinder, apache-tvm-ffi, cuda-bindings, torch-c-dlpack-ext, cuda-python, nvidia-cutlass-dsl-libs-base, nvidia-cutlass-dsl, quack-kernels, flash-attn-4
Successfully installed apache-tvm-ffi-0.1.9 cuda-bindings-13.1.1 cuda-pathfinder-1.4.0 cuda-python-13.1.1 flash-attn-4-4.0.0b4 nvidia-cutlass-dsl-4.4.1 nvidia-cutlass-dsl-libs-base-4.4.1 quack-kernels-0.2.10 torch-c-dlpack-ext-0.1.5
```

Then I tried importing FA4:

```
python -c 'from flash_attn.cute import flash_attn_func'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'flash_attn.cute'
```

FA4 is indeed installed in this env:

```
pip show flash-attn-4
Name: flash-attn-4
Version: 4.0.0b4
Summary: Flash Attention CUTE (CUDA Template Engine) implementation
Home-page: https://github.com/Dao-AILab/flash-attention
Author: Tri Dao
Author-email: 
License: BSD 3-Clause License
Location: /home/alex/.local/lib/python3.10/site-packages
Requires: apache-tvm-ffi, einops, nvidia-cutlass-dsl, quack-kernels, setuptools, torch, torch-c-dlpack-ext, typing_extensions
Required-by: 


pip show flash-attn-3
Name: flash_attn_3
Version: 3.0.0b1
Summary: FlashAttention-3
Home-page: 
Author: 
[...]

PR #2331, opened 2026-03-11
Fix FA2 + FA4 co-existence

drisspg:
# Summary

This extends the path search to also optionally include fa4 if installed in path (site packages)

SO is helpful here https://stackoverflow.com/questions/2699287/what-is-path-useful-for

For:#2310


drisspg:
@tridao locally this fixed it, will need to cute a new fa2 release though

tridao (approved):

PR #2330, opened 2026-03-11
[Bwd,Sm120] Add SM120 backward pass support

blake-snc:
## Summary

Add backward pass support for SM120 (Blackwell GeForce / DGX Spark). SM120 uses SM80-era MMA instructions (`mma.sync.aligned.m16n8k16`) with 99 KB shared memory (vs 163 KB on SM80).

- `FlashAttentionBackwardSm120` subclass overrides `can_implement()` with SM120 SMEM capacity (99 KB)
- SM120 backward tile config: m=n=64, stages=1 (D>64) or 2 (D≤64), 128 threads (4 warps, all in M-direction)
- Pre/postprocess kernels use SM80 code paths (`post_arch=80`)
- Skips head_dim validation (SM120 not in the validation table) and preserves 128 threads (no override to 384)

Depends on PR #2329 (SM120 forward pass).

## Validation

Tested on SM121a (DGX Spark, NVIDIA GB10) with BF16 across 14 configurations:

| Head dim | Seq len | Causal | dq_max | dk_max | dv_max |
|----------|---------|--------|--------|--------|--------|
| 64 | 256 | No | 0.0039 | 0.0039 | 0.0039 |
| 64 | 256 | Yes | 0.0156 | 0.0078 | 0.0156 |
| 64 | 512 | No | 0.0039 | 0.0039 | 0.0039 |
| 64 | 512 | Yes | 0.0078 | 0.0098 | 0.0156 |
| 96 | 256 | No | 0.0039 | 0.0039 | 0.0039 |
| 96 | 256 | Yes | 0.0156 | 0.0156 | 0.0156 |
| 96 | 512 | No | 0.0039 | 0.0039 | 0.0020 |
| 96 | 512 | Yes | 0.0156 | 0.0156 | 0.0313 |
| 128 | 256 | No | 0.0039 | 0.0078 | 0.0039 |
| 128 | 256 | Yes | 0.0156 | 0.0156 | 0.0313 |
| 128 | 512 | No | 0.0020 | 0.0020 | 0.0020 |
| 128 | 512 | Yes | 0.0156 | 0.0078 | 0.0313 |
| 128 | 1024 | No | 0.0020 | 0.0020 | 0.0020 |
| 128 | 1024 | Yes | 0.0156 | 0.0156 | 0.0156 |

All 14/14 tests pass. Gradient max diffs < 0.032 (BF16 expected range).

## Test plan

- [ ] Run backward pass tests on SM121a hardware
[...]

PR #2329, opened 2026-03-11
SM120 forward pass (Blackwell GeForce / DGX Spark)

blake-snc:
## Summary

- Add `FlashAttentionForwardSm120`, a thin subclass of `FlashAttentionForwardSm80` that overrides `can_implement()` with SM120's 99 KB SMEM capacity check
- Wire SM120 dispatch into `interface.py` with SM120-tuned tile sizes and SM80-compatible argument ordering
- Fix an SM80 causal softmax bug in `flash_fwd.py` that was exposed by SM120 testing (see below)

SM120 (Blackwell GeForce / DGX Spark) uses the same SM80-era MMA instructions (`mma.sync.aligned.m16n8k16`) but has 99 KB shared memory vs 163 KB on SM80 and 228 KB on datacenter Blackwell. No tcgen05, no WGMMA, no TMEM — but does have TMA (not used in this PR).

### SM80 causal bug fix (first commit, can be cherry-picked independently)

The causal masking loop in `FlashAttentionForwardSm80.kernel()` passed `is_first_n_block=True` to `compute_one_n_block` for all causal iterations. This caused `online_softmax` to reset the running max/sum on each iteration, producing incorrect output when there are 2+ causal N-blocks (D>64 with tile_n=64). The SM90 kernel's causal section correctly omits `is_first_n_block` (defaults to `False`). This was a regression from the SM80 refactor that split SM90 into its own file.

### Tile sizes

| head_dim | tile_m x tile_n | SMEM usage | Notes |
|----------|----------------|------------|-------|
| ≤64 | 128×128 | 48 KB | Good occupancy |
| >64 | 128×64 | 64 KB | 128×128 would use 96 KB |

### Argument ordering

SM80 `__call__` takes `(stream, softmax_scale)` while SM90/SM100/SM110 take `(softmax_scale, stream)`. SM80 also requires `Optional[Float32]` for softmax_scale. The dispatch handles this via conditional argument swapping.

## Limitations (future PRs)

- No SplitKV (flash decoding)
- No paged KV cache
- No varlen sequences
- No backward pass
- No TMA optimization

## Test plan

Validated on NVIDIA GB10 (DGX Spark, SM121a) — 17/17 configs pass:

```
GPU: NVIDIA GB10
Compute capability: (12, 1)

PASS B=2 Sq=256 Sk=256 H=8 D=64  causal=False: max_diff=0.003906
PASS B=2 Sq=256 Sk=256 H=8 D=64  causal=True:  max_diff=0.007812
[...]

PR #2333, opened 2026-03-12
Add SM120 varlen attention support

blake-snc:
## Summary

- Add variable-length (`flash_attn_varlen_func`) support for SM120 (Blackwell GeForce / DGX Spark)
- Fix backward tile size mismatch in `SeqlenInfoQK.create` that caused incorrect dQ gradients
- Fix forward crash from OOB `cu_seqlens` reads by wasted varlen scheduler grid tiles

Depends on #2329 (forward) and #2330 (backward).

## Changes

**Forward varlen (flash_fwd.py):**
- Accept 3D varlen tensors `(total_seqlen, nheads, headdim)` alongside existing 4D layout
- Select `SingleTileVarlenScheduler` when `cu_seqlens` are provided
- Use `cute.domain_offset` for per-batch Q/K/V indexing from cu_seqlens offsets
- Reshape LSE output to `(nheads, total_seqlen)` for varlen

**Forward OOB fix (flash_fwd.py, interface.py):**
The varlen tile scheduler can produce "wasted" grid tiles where `batch_idx >= num_batch`.
The SM80 forward kernel calls `SeqlenInfoQK.create(batch_idx)` before checking tile
validity, causing `cu_seqlens[num_batch+1]` OOB reads and `cudaErrorIllegalAddress`.
Fixed by:
- Padding `cu_seqlens` with one extra element (repeating last value) so the OOB read
  returns `seqlen=0` instead of crashing
- Clamping `n_block = max(n_block_max - 1, 0)` so wasted tiles with `seqlen=0` don't
  produce negative K/V block indices (the existing load/store predicates already guard
  all memory accesses when seqlen is 0)

**Backward tile fix (flash_bwd.py, flash_bwd_preprocess.py):**
SM120 backward uses `m_block_size=64, n_block_size=64` but `SeqlenInfoQK.create` defaulted
to `tile_m=128, tile_n=128`, causing incorrect `padded_offset` computation and wrong dQ.
Fixed by passing actual tile sizes.

## Validation

Tested on SM121a (DGX Spark) with 12 varlen configs:
- Head dims: D=64, D=96, D=128
- Causal and non-causal
- Equal seqlens: `[128,128]`, `[256,256]`, `[512,256]`
- Unequal seqlens: `[64,128,256]`, `[128,64,192]`, `[128,256]`

All 12 forward+backward pass (fwd max_diff ≤ 0.016, bwd max_diff ≤ 0.031).
Non-varlen regression: 6/6 pass.

### DIFF
--- a/flash_attn/cute/flash_fwd.py
+++ b/flash_attn/cute/flash_fwd.py
@@ -666,12 +666,15 @@ class FlashAttentionForwardSm80(FlashAttentionForwardBase):
             LSE_layout_transpose = [2, 1, 0] if const_expr(mCuSeqlensQ is None) else [1, 0]
             mLSE = cute.make_tensor(mLSE.iterator, cute.select(mLSE.layout, mode=LSE_layout_transpose))
         # TileScheduler for varlen, simple grid for non-varlen
-        if const_expr(mCuSeqlensQ is not None):
+        if const_expr(mCuSeqlensQ is not None or mSeqUsedQ is not None):
             TileScheduler = SingleTileVarlenScheduler
-            num_batch = mCuSeqlensQ.shape[0] - 1
         else:
             TileScheduler = SingleTileScheduler
-            num_batch = mQ.shape[3]
+        num_batch = (
+            mCuSeqlensQ.shape[0] - 1
+            if const_expr(mCuSeqlensQ is not None)
+            else mQ.shape[3]
+        )
         tile_sched_args = TileSchedulerArguments(
             num_block=cute.ceil_div(mQ.shape[0], self.tile_m),
             num_head=cute.size(mQ.shape[2]),
@@ -680,7 +683,9 @@ class FlashAttentionForwardSm80(FlashAttentionForwardBase):
             seqlen_k=0,
             headdim=mQ.shape[1],
             headdim_v=mV.shape[1],
-            total_q=cute.size(mQ.shape[0]),
+            total_q=cute.size(mQ.shape[0])
+            if const_expr(mCuSeqlensQ is not None)
+            else cute.size(mQ.shape[0]) * cute.size(mQ.shape[3]),
             tile_shape_mn=(self.tile_m, self.tile_n),
             qhead_per_kvhead_packgqa=self.qhead_per_kvhead if const_expr(self.pack_gqa) else 1,
             mCuSeqlensQ=mCuSeqlensQ,


flash_attn/cute/interface.py
```diff
@@ -299,6 +299,13 @@ def _flash_attn_fwd(
             learnable_sink,
         )
     ), "inputs must be on CUDA device"
+    # Pad cu_seqlens with one extra element (repeating the last value) so that
```
tridao: This is very expensive because it causes a gpu-cpu sync. We should avoid changing cu_seqlens_q

blake-snc:
Removed the cu_seqlens padding in 8093403. It was unnecessary — `SingleTileVarlenScheduler` already guards OOB reads: `_get_num_m_blocks()` checks `batch_idx <= num_batch` before reading cu_seqlens (valid for num_batch+1 element arrays), and wasted tiles get `is_valid=False` so the kernel never enters `SeqlenInfoQK` with an invalid batch_idx. Validated on SM121a with non-causal, causal, and unaligned-seqlen varlen configs.

flash_attn/cute/flash_fwd.py
```diff
@@ -650,18 +651,43 @@ def __call__(
...
-            mLSE = cute.make_tensor(mLSE.iterator, cute.select(mLSE.layout, mode=[2, 1, 0]))
-        # grid_dim: (m_block, num_head, batch_size)
-        grid_dim = (
-            cute.ceil_div(mQ.shape[0], self.tile_m),
-            cute.size(mQ.shape[2]),
-            cute.size(mQ.shape[3]),
+            LSE_layout_transpose = [2, 1, 0] if const_expr(mCuSeqlensQ is None) else [1, 0]
+            mLSE = cute.make_tensor(mLSE.iterator, cute.select(mLSE.layout, mode=LSE_layout_transpose))
+        # TileScheduler for varlen, simple grid for non-varlen
+        if const_expr(mCuSeqlensQ is not None):
```
tridao: mCuSeqlensQ is not None or mSequsedQ is not None

flash_attn/cute/flash_fwd.py
```diff
@@ -650,18 +651,43 @@ def __call__(
...
+            TileScheduler = SingleTileVarlenScheduler
+            num_batch = mCuSeqlensQ.shape[0] - 1
```
tridao: need to deal w the case where there's no cu_seqlens_q but there's seqused_q

blake-snc:
Fixed in 74e13f8. Updated the scheduler selection condition to `mCuSeqlensQ is not None or mSeqUsedQ is not None` (matching SM90 forward on main), and fixed `total_q` to be `seqlen * batch` for the 4D layout case. Validated on SM121a: seqused-only output is bit-exact with the equivalent short-sequence reference.

PR #2338, opened 2026-03-12
[Fwd,SM100,CuTe] Fix split KV OOM with diff headdim + fix SM100 kwarg mismatch

MatthewBonanni:
Using split KV with diff headdim (e.g. 192/128 for DeepSeek MLA prefill) exceeds SMEM due to float32 partial output doubling the O buffer size. This PR reduces tile size or disables splitting in that case, with a heuristic to ensure optimal performance.

Also fixes a bug introduced in 99d0148f where the SM100 constructor is called with `tile_m`/`tile_n` instead of `m_block_size`/`n_block_size`, which would cause a `TypeError`.

This change has been applied to [vLLM's FA fork](https://github.com/vllm-project/flash-attention) via https://github.com/vllm-project/flash-attention/pull/123 and https://github.com/vllm-project/flash-attention/pull/126 to enable using FA4 for MLA prefill in vLLM (https://github.com/vllm-project/vllm/pull/34732). This PR applies the fix upstream.

Benchmarking performed using vLLM attention benchmark tool:
559494626-b96cdd34-9c9a-4a27-95ee-73f437e5d80d

### DIFF
--- a/flash_attn/cute/flash_fwd_sm100.py
+++ b/flash_attn/cute/flash_fwd_sm100.py
@@ -274,7 +274,7 @@ class FlashAttentionForwardSm100:
             self.head_dim_padded == 192 and self.head_dim_v_padded == 128 and self.kv_stage == 3
         )
         self.uneven_kv_smem_offset = (
-            self.m_block_size * (self.head_dim_padded - self.head_dim_v_padded) // 2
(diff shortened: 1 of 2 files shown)


tridao:
I feel the right approach is to subtile the O inside the kernel so we can keep the same size of smem. But that's more annoying to impl so for now we can just decrease tile_n.

PR #2343, opened 2026-03-13
[ROCm] Auto-detect Triton backend if C++ extension is missing

Soddentrough:
On AMD/ROCm systems, especially with RDNA-based GPUs (where only the Triton backend is supported), it's currently necessary to manually set `FLASH_ATTENTION_TRITON_AMD_ENABLE=TRUE`. This change implements a fallback that automatically switches to the Triton backend if it's a HIP system and the main C++ extension (`flash_attn_2_cuda`) is not available. Tested on gfx1201.

### DIFF
--- a/flash_attn/flash_attn_interface.py
+++ b/flash_attn/flash_attn_interface.py
@@ -5,6 +5,7 @@ from typing import Optional, Sequence, Tuple, Union
 import torch
 import torch.nn as nn
 import os
+import warnings
 
 # isort: off
 # We need to import the CUDA kernels after importing torch
@@ -13,6 +14,7 @@ if not USE_TRITON_ROCM and getattr(torch.version, 'hip', None) is not None:
     try:
         import flash_attn_2_cuda
     except ImportError:
+        warnings.warn("flash_attn_2_cuda (which has ROCm/HIP kernels) not found, falling back to Triton implementation")
         USE_TRITON_ROCM = True
 
 if USE_TRITON_ROCM:
--- a/hopper/flash_attn_interface.py
+++ b/hopper/flash_attn_interface.py
@@ -5,6 +5,7 @@ from typing import Optional, Union, List, Tuple
 import os
 import torch
 import torch.nn as nn
+import warnings
 
 
 USE_TRITON_ROCM = os.getenv("FLASH_ATTENTION_TRITON_AMD_ENABLE", "FALSE") == "TRUE"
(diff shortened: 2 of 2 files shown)


tridao:
cc @micmelesse does this look fine to you?

hopper/flash_attn_interface.py
```diff
@@ -17,7 +23,7 @@
 
     # isort: on
 
-    flash_attn_3_gpu = torch.ops.flash_attn_3
+    flash_attn_3_gpu = torch.ops.flash_attn_3_gpu
```
micmelesse: Is this intentional? It's on the non-Triton path and looks unrelated.

flash_attn/flash_attn_interface.py
```diff
@@ -9,6 +9,12 @@
 # isort: off
 # We need to import the CUDA kernels after importing torch
 USE_TRITON_ROCM = os.getenv("FLASH_ATTENTION_TRITON_AMD_ENABLE", "FALSE") == "TRUE"
+if not USE_TRITON_ROCM and getattr(torch.version, 'hip', None) is not None:
+    try:
+        import flash_attn_2_cuda
+    except ImportError:
```
micmelesse, line 16: We should at least log a warning when falling back. So that it is clear what is happening to users.

hopper/flash_attn_interface.py
```diff
@@ -8,6 +8,12 @@
 
 
 USE_TRITON_ROCM = os.getenv("FLASH_ATTENTION_TRITON_AMD_ENABLE", "FALSE") == "TRUE"
+if not USE_TRITON_ROCM and getattr(torch.version, 'hip', None) is not None:
+    try:
+        import flash_attn_3._C
+    except ImportError:
```
micmelesse, line 15: A similar warning log for users here would be good.

hopper/flash_attn_interface.py
rocking5566: The TORCH_LIBRARY registration in hopper/flash_api.cpp uses flash_attn_3 (not flash_attn_3_gpu), I thought this might be incoorect

Soddentrough: Sorry, that is indeed _not_ correct. The intent was to match variable name in the USE_TRITON_ROCM path ("import flash_attn_3 as flash_attn_3_gpu") but not required or desired.

flash_attn/flash_attn_interface.py
Soddentrough: Is it best to import 'warnings' and use that for logging in this instance?

Issue #2344, opened 2026-03-13
Need bump quark-kernels to at least 0.3.0


rainj-me:
Current main throw no 'create_ragged_tensor_for_tma' error. I double checked the quark-kernels, the 'create_ragged_tensor_for_tma' is introduced at quark-kernels>=0.3.0, please bump the quark-kernels.

```bash
    return then_block(*block_args) if pred else else_block(*block_args)
           ^^^^^^^^^^^^^^^^^^^^^^^
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

    if const_expr(self.qhead_per_kvhead == 1):
        mdK_tma = (
>           copy_utils.create_ragged_tensor_for_tma(mdK, ragged_dim=0, ptr_shift=True)
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            if self.varlen_k
            else mdK
        )
E       AttributeError: module 'quack.copy_utils' has no attribute 'create_ragged_tensor_for_tma'
```

tridao:
Thanks for the fix

PR #2345, opened 2026-03-13
fix the create_ragged_tensor_for_tma issue

rainj-me:
#2344 

The quack-kernels 0.2.10 doesn't have the create_ragged_tensor_for_tma issue function in quack.utils.


tridao:
Thanks!

PR #2336, opened 2026-03-12
Add SM120 split-KV (FlashDecoding) with FP32 partial outputs

blake-snc:
## Summary

- Add split-KV (FlashDecoding) support for SM120 (Blackwell GeForce / DGX Spark) — enables `num_splits > 1` for decode-like workloads with short Q and long K
- FP32 partial outputs: split-KV kernel writes FP32 accumulators directly to GMEM, bypassing SMEM to preserve full precision for the combine kernel
- Paged KV support removed from this PR per review feedback — will be a separate follow-up

Builds on #2329, #2330, #2333.

## Key changes

**flash_fwd.py (FlashAttentionForwardBase + FlashAttentionForwardSm80)**:
- `is_split_kv` parameter wired through `__init__`, `_check_type` (allow FP32 mO), epilogue (`split_idx` O/LSE indexing), `__call__` (O/LSE layout transposes with split dimension, num_splits through tile scheduler), and kernel body
- Kernel body wrapped in `n_block_max > n_block_min` guard for empty-split correctness
- Remaining-iterations loop uses `n_block - n_block_min` to stop at split boundary
- **FP32 direct write epilogue**: for `o_dtype != dtype`, bypasses SMEM entirely and writes MMA accumulators directly to GMEM via `thr_mma.partition_C(gO)` with per-element bounds checking. This avoids the SMEM store/load layout mismatch between `make_tiled_copy_C` (MMA accumulator layout) and `gmem_tiled_copy_O` (GMEM coalescing layout) that corrupts FP32 data through SMEM.

**interface.py**:
- `out_partial` allocated as FP32 `torch.zeros`, `lse_partial` as `torch.full(-inf)`
- `is_split_kv` passed to SM120 constructor

**flash_fwd_sm120.py**:
- `can_implement` unchanged (FP32 direct write doesn't use SMEM for output)

## Validation on SM121a (DGX Spark)

```
Split-KV: 14/14 configs pass
  - num_splits: 2, 4
  - causal and non-causal
  - Sq=16/32/64/128/256, Sk=256/512/1024
  - H=1/2/4/8, D=64/128
  - split_diff matches nosplit precision (0.001-0.002 vs reference)

Non-split regression: 5/5 configs pass
  - B=1/2, Sq=64/128/256, Sk=64/128/256
  - causal and non-causal, D=64/128
```

Contributed by Second Nature Computing (https://joinsecondnature.com)

PR #2348, opened 2026-03-13
Add SM120 kernel-level paged KV cache support

blake-snc:
## Summary

Implement paged KV cache for SM120 forward kernel using PagedKVManager for in-kernel page table resolution. All page table lookups, pointer arithmetic, and cp.async loads happen inside the CUDA kernel — no PyTorch-level page gathering.

Builds on #2336 (split-KV), #2329, #2330, #2333.

## Key changes

**flash_fwd.py (FlashAttentionForwardBase)**:
- \`_paged_load_page_table_and_ptrs\`: helper that computes per-row GMEM pointers from page table using plain Int32 division. Avoids FastDivmodDivisor which causes MLIR SSA domination errors when used as a loop-carried variable inside the SM80 mainloop.
- \`paged_load_K\` / \`paged_load_V\`: per-row paged loads using shuffle_sync pointer broadcast + chunk-by-chunk cp.async to SMEM. Uses PagedKVManager's copy atom (identical thread layout to SM80's gmem_tiled_copy_K/V).
- Kernel body branches on \`mPageTable is not None\` to create PagedKVManager and use paged load functions instead of standard contiguous loads.

**interface.py**:
- Removed SM120 paged KV assertion
- Added seqused_k requirement for paged KV (since mK.shape[0] = page_size, not seqlen_k)
- For D>64: bumps tile_n from 64 to 128 for paged KV (PagedKVManager needs tile_n >= num_threads). Fits in 99KB SMEM (96KB data + 3KB alignment).

## Validation on SM121a (DGX Spark)

\`\`\`
Paged KV: 8/8 configs pass (D=64+128, page_size=128)

D=64:
  B=1 Sq=64  Sk=128 causal=False: diff=0.003906
  B=1 Sq=64  Sk=256 causal=True:  diff=0.001953
  B=2 Sq=128 Sk=256 causal=False: diff=0.003906

D=128 (tile_n bumped to 128 for paged KV):
  B=1 Sq=128 Sk=128 causal=False: diff=0.003906
  B=1 Sq=128 Sk=256 causal=False: diff=0.001953
  B=1 Sq=128 Sk=256 causal=True:  diff=0.001953
  B=2 Sq=64  Sk=256 causal=False: diff=0.003906
  B=1 Sq=64  Sk=512 causal=True:  diff=0.001953

Non-paged regression: passes
\`\`\`

### DIFF
--- a/flash_attn/cute/flash_fwd.py
+++ b/flash_attn/cute/flash_fwd.py
@@ -34,6 +34,8 @@ from flash_attn.cute.pack_gqa import PackGQA
 from flash_attn.cute.named_barrier import NamedBarrierFwd
 from flash_attn.cute.block_sparsity import BlockSparseTensors
 from flash_attn.cute.tile_scheduler import SingleTileScheduler, SingleTileVarlenScheduler, TileSchedulerArguments
+from flash_attn.cute.paged_kv import PagedKVManager
+from cutlass.cute import FastDivmodDivisor
 
 
 class FlashAttentionForwardBase:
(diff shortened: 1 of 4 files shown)


johnnynunez:
+viz @tridao @drisspg

mmdbhs:
The inference time of this PR's paged KV cache has increased by 10 times compared to the dense KV cache's kernel time. Why is the performance gap so large? Are there any optimizations planned for the future?

blake-snc:
~~The 10x gap is expected — it's a direct consequence of scatter-gather memory access vs. contiguous loads, not a bug or missing optimization in this PR.~~

**Edit: this explanation was wrong.** As @tridao correctly pointed out, head_dim is the contiguous dimension — the loads within each row are still coalesced regardless of paging. The 10x gap was a real performance problem, not an architectural inevitability. See the follow-up comment for the fix.

tridao:
I don't think perf diff should be 10x. In the past we've see 0-10% perf diff between dense and paged KV. Note that the load should still be coalesced (head dim the contiguous dimension)

blake-snc:
Implemented double-buffered (num_stages=2) K/V loading for SM120 in the latest commit. This pipelines the page-table-lookup + CpAsync scatter loads with MMA execution, which is the core optimization for reducing the paged-KV latency gap.

Two changes:

**`paged_kv.py`**: Fixed a floor-division bug in `page_entry_per_thread` (was `n_block_size // num_threads = 64 // 128 = 0` for D=128, causing a crash). Changed to ceiling division so all threads get a page entry slot even when tile_n < num_threads, with out-of-range entries guarded by `is_valid=False`.

**`interface.py`**: Bumped SM120 to `num_stages=2`. The SMEM budget works out:
- D=128, tile_n=64: sQ(32KB) + sK(32KB) + sV(32KB) = 96KB ≤ 99KB SM120 capacity
- D≤64, tile_n=128: 16+32+32 = 80KB ✓

Validated on SM121a (DGX Spark): dense causal/non-causal and paged KV causal/non-causal all pass across D=64/128 and various sequence lengths. @tridao you're right that with coalesced head-dim loads the gap should be much smaller — the remaining overhead is just the per-row pointer indirection (page table lookup) vs. a single strided base address.

blake-snc:
Benchmarked paged vs dense on SM121a (DGX Spark, BF16, page_size=64, causal=True, decode scenario Sq=1):

| Config | Dense (ms) | Paged (ms) | Ratio |
|--------|-----------|-----------|-------|
| B=1 Sk=2048 H=16 D=128 | 0.080 | 0.095 | 1.18x |
| B=1 Sk=4096 H=16 D=128 | 0.159 | 0.189 | 1.19x |
| B=4 Sk=2048 H=16 D=128 | 0.289 | 0.282 | 0.98x |
| B=4 Sk=4096 H=16 D=128 | 0.577 | 0.652 | 1.13x |
| B=1 Sk=2048 H=16 D=64  | 0.046 | 0.054 | 1.19x |
| B=4 Sk=2048 H=16 D=64  | 0.149 | 0.163 | 1.09x |

0–19% overhead, consistent with the 0–10% baseline @tridao mentioned. The fix was `num_stages=2` (double-buffered K/V loads) which pipelines the per-row page table lookups with MMA, plus a ceiling-division fix in `page_entry_per_thread` that was required for D=128 with tile_n=64.

**Update**: The B=1 entries showing ~19% overhead are within measurement noise. When re-run with CUDA events across B=1–32, the per-kernel σ is ~54µs while the delta between paged and dense is ~19µs (SNR < 1). At B=8–32 paged actually runs a few µs *faster* than dense — also noise. There is no statistically significant batch-size-dependent overhead; paged KV runs within measurement precision of dense KV across all tested configs.

blake-snc:
Rebased onto current main (6362bd3) and validated on SM121a (DGX Spark).

**Validation results (all pass, max_diff=0.000000):**

| Config | max_diff_O | max_diff_LSE |
|--------|-----------|-------------|
| non-causal bs=2 sq=128 sk=512 d=128 page=64 | 0.000000 | 0.000000 |
| causal bs=2 sq=128 sk=512 d=128 page=64 | 0.000000 | 0.000000 |
| non-causal bs=1 sq=256 sk=256 d=128 page=64 | 0.000000 | 0.000000 |
| causal bs=1 sq=256 sk=256 d=128 page=64 | 0.000000 | 0.000000 |
| non-causal bs=4 sq=64 sk=512 d=64 page=64 | 0.000000 | 0.000000 |
| non-causal bs=2 sq=128 sk=256 d=128 page=128 | 0.000000 | 0.000000 |
| decode bs=1 sq=1 sk=512 d=128 page=64 | 0.000000 | 0.000000 |
| decode-causal bs=1 sq=1 sk=512 d=128 page=64 | 0.000000 | 0.000000 |

This PR depends on #2336 (split-KV) — the paged KV commits are layered on top of that branch.

johnnynunez:
ping @tridao @drisspg

geraldstanje1:
hi @blake-snc i would like to test it on nvidia rtx 6000 pro - how can i run the same benchmark you run above?
also will this also reduce end-to-end latency with prefix caching enabled and concurrency 1-5? any settings i need for vllm to run your pr for vllm bench?

blake-snc:
@geraldstanje1 The benchmarks in the PR description are correctness tests (output matches non-paged reference), not performance benchmarks. To run the correctness tests on your RTX PRO 6000:

```bash
git clone https://github.com/blake-snc/flash-attention.git -b feat/sm120-paged-kv
cd flash-attention
pip install -e "flash_attn/cute[dev]"
pytest tests/cute/test_flash_attn.py -k "paged" -x -v
```

For end-to-end vLLM latency with prefix caching: we're actively building a flash-attn-4 attention backend for vLLM to wire these SM120 kernels (paged KV, split-KV, etc.) into the serving path. Will update here once it's ready for testing.

In the meantime, the working SM120 inference path in vLLM is `--attention-backend flashinfer`.

blake-snc:
@geraldstanje1 here's the actual perf bench script I used for the dense-vs-paged table above. Drop it somewhere and run it against a checkout of `blake-snc/flash-attention:feat/sm120-paged-kv` with `flash_attn/cute[dev]` installed:

<details>
<summary><code>bench_paged_kv_sm120.py</code></summary>

```python
"""Benchmark paged KV vs dense KV for FlashAttentionForwardSm120 on Blackwell consumer GPUs."""
from __future__ import annotations
import gc, statistics as stats, time
import torch


def _bench(fn, warmup=5, iters=50):
    for _ in range(warmup): fn()
    torch.cuda.synchronize()
    samples = []
    for _ in range(iters):
        torch.cuda.synchronize()
        t0 = time.perf_counter()
        fn()
        torch.cuda.synchronize()
        samples.append((time.perf_counter() - t0) * 1000.0)
    return stats.median(samples)


def _build_paged_kv(B, Sk, H_kv, D, page_size, device, dtype):
    assert Sk % page_size == 0
    pages_per_batch = Sk // page_size
    num_pages = B * pages_per_batch
    k_paged = torch.randn(num_pages, page_size, H_kv, D, device=device, dtype=dtype)
    v_paged = torch.randn(num_pages, page_size, H_kv, D, device=device, dtype=dtype)
    page_table = torch.arange(num_pages, dtype=torch.int32, device=device).view(B, pages_per_batch)
    return k_paged, v_paged, page_table


def bench_dense(B, Sq, Sk, H, D, dtype, causal):
    q = torch.randn(B, Sq, H, D, device="cuda", dtype=dtype)
    k = torch.randn(B, Sk, H, D, device="cuda", dtype=dtype)
    v = torch.randn(B, Sk, H, D, device="cuda", dtype=dtype)
    scale = 1.0 / (D ** 0.5)
    return lambda: _flash_attn_fwd(q, k, v, softmax_scale=scale, causal=causal)


def bench_paged(B, Sq, Sk, H, D, page_size, dtype, causal):
    q = torch.randn(B, Sq, H, D, device="cuda", dtype=dtype)
    k_paged, v_paged, page_table = _build_paged_kv(B, Sk, H, D, page_size, "cuda", dtype)
    seqused_k = torch.full((B,), Sk, dtype=torch.int32, device="cuda")
    scale = 1.0 / (D ** 0.5)
    return lambda: _flash_attn_fwd(
        q, k_paged, v_paged,
        page_table=page_table, seqused_k=seqused_k,
        max_seqlen_k=Sk, softmax_scale=scale, causal=causal,
    )


def main():
    configs = [
        # (B, Sq, Sk, H, D) — decode-style (Sq=1)
        (1, 1, 2048, 16, 128), (1, 1, 4096, 16, 128),
        (4, 1, 2048, 16, 128), (4, 1, 4096, 16, 128),
        (1, 1, 2048, 16,  64), (4, 1, 2048, 16,  64),
    ]
    cap = torch.cuda.get_device_capability(0)
    print(f"GPU: {torch.cuda.get_device_name(0)} (SM{cap[0]}{cap[1]}) | BF16 | causal=True | page_size=64\n")
    header = f"{'Config':<34} {'Dense (ms)':>12} {'Paged (ms)':>12} {'Ratio':>8}"
    print(header); print("-" * len(header))
    for B, Sq, Sk, H, D in configs:
        torch.manual_seed(0); dense_ms = _bench(bench_dense(B, Sq, Sk, H, D, torch.bfloat16, True))
        gc.collect(); torch.cuda.empty_cache()
        torch.manual_seed(0); paged_ms = _bench(bench_paged(B, Sq, Sk, H, D, 64, torch.bfloat16, True))
        gc.collect(); torch.cuda.empty_cache()
        print(f"B={B} Sk={Sk} H={H} D={D:<20} {dense_ms:>12.3f} {paged_ms:>12.3f} {paged_ms/dense_ms:>7.2f}x")


if __name__ == "__main__":
    main()
```

</details>

Output on SM121a (DGX Spark GB10) reproduces the table in the PR description within noise. On your RTX PRO 6000 Blackwell I would expect absolute ms to be lower, ratios in the same 1.0x-1.2x band.

On the e2e / prefix-caching question: at the kernel level, prefix caching and paged KV are orthogonal. Paged KV is what enables vLLM's paged attention abstraction, which is what prefix caching then reuses, but the per-call kernel latency you see in the table above is independent of whether the KV cache is fresh or reused. End-to-end latency wins from prefix caching come from skipping prefill compute for the shared prefix, not from the attention kernel itself running faster.

For vLLM e2e testing on RTX 6000 Pro: FA4 is actually already integrated in vLLM via the vendored flash-attn tree, it is gated out for SM12 in `_is_fa4_supported()` due to three runtime bugs in the upstream `FlashAttentionForwardSm120` class that I recently fixed at https://github.com/Dao-AILab/flash-attention/pull/2484. Once that fix plus this PR (#2348) plus #2336 land upstream and vLLM does a vendored flash-attn sync, the SM12 gate can be opened and paged KV will be accessible through vLLM's standard FA4 path.

Worth knowing before you plan around it: FA4 on SM120 is slower than FA2 on SM120 today. SM120 has no tcgen05 / TMEM, so FA4 compiles down to the same SM80 era `mma.sync` path as FA2 with extra dispatch overhead on top. On realistic Qwen 3 prefill shapes I measured patched FA4 within roughly 15 percent of FA2 at longer sequences and slower at short ones. So even after the gate opens, FA4 is not a perf win for SM120, the motivation for enabling it is feature reach (score_mod, paged KV, split KV, block sparse) rather than throughput.

For `gpt-oss-20b` on your RTX PRO 6000 specifically, the sliding-window issue you hit is a separate gap in fmha_v2 on SM120, tracked at https://github.com/vllm-project/vllm/pull/40085 (adds the SM120 sinks path via fmha_v2) and flashinfer #3016 (wires SM120 standard shapes into fmha_v2). Neither of those is in this PR's scope but I will follow up over on #2336 where you were asking about it.

blake-snc:
Rebased and restructured as a stack on #2336 (first commit is the split-KV change; this PR's own commit is the paged work). Fills the second #2329 gate ("Paged KV not supported on SM 12.0"): non-TMA paged KV with page-table lookups interleaved into the cp.async pipeline (num_stages=2; SMEM budget at D=128, tile_n=64 is 96 KB <= 99 KB).

Validation on sm_121a: paged output is bit-identical to the dense equivalent (max diff 0.00000) across D=64/128, page sizes 16/32/64, causal and GQA, via `flash_attn_varlen_func` + `page_table` + `seqused_k`. Paged + split-KV compose (verified at num_splits 2 and 4). Requires seqused_k, matching the existing paged contract.

blake-snc:
Status update: our SM120 hardware is currently allocated to other work, so I am unable to rebase and re-validate this PR for the time being. The branch was validated on sm_121a as of the 2026-07-15 restructure (paged output bit-identical to dense; details above); it is stacked on #2336 and will need a rebase before merge. The paged KV gate it fills is still present on current main ("Paged KV not supported on SM 12.0 in this PR"), so I would be glad to see it adopted by anyone with the hardware. For design questions I am also reachable at github.com/bledden. Leaving this open for adoption; maintainers should feel free to close if preferred.

PR #2351, opened 2026-03-14
Fix RuntimeError when out_ is specified in _flash_attn_forward (fix #2073)

### LINKED ISSUE #2073: RuntimeError when `out_` is specified in `_flash_attn_forward`
In `_flash_attn_forward` function, if `out_` argument is specified, an error will raise as an output tensor is present in argument (returned `out` is the same as argument `out_`).

I resolved this error by adding the following code to  `_flash_attn_forward` function.

```python
    if out_ is not None:
        out = None

```

Detailed error:

```python
 RuntimeError: _flash_attn_forward (with implementation in <module 'torch._library.custom_ops' from '/usr/local/lib/python3.13/site-packages/torch/_library/custom_ops.py'>): The output of this custom operator (1) must not also be an input to this custom operator and (2) may not alias any inputs to this custom operator or other returns. The most common way to trigger this error is if we have y = custom_op(x) and y and x are the same Tensor. Please instead return a clone of the offending output tensor(s) (e.g. return x.clone()) or refactor the custom operator to not return y.
```

cc @guilhermeleobas

NJX-njx:
## Summary
This fix addresses issue #2073 by adding a check to handle the case when the out_ parameter is provided.

## Problem
When the out_ argument is specified in _flash_attn_forward, an error is raised because PyTorch custom operators cannot return an alias of an input tensor.

## Solution
Added a check: when out_ is specified, set out = None to avoid returning an alias of the input, which violates PyTorch custom operator rules.

## Error Before Fix
RuntimeError: The output of this custom operator (1) must not also be an input to this custom operator and (2) may not alias any inputs to this custom operator or other returns.

Closes #2073

### DIFF
--- a/hopper/flash_attn_interface.py
+++ b/hopper/flash_attn_interface.py
@@ -147,6 +147,9 @@ def _flash_attn_forward(
     if softmax_lse_accum is None:
         softmax_lse_accum = torch.tensor([], device=out.device)
 
+    if out_ is not None:
+        out = None
+
     return out, softmax_lse, out_accum, softmax_lse_accum
 
 


NJX-njx:
Rebased onto main. Reduced the diff to only the 3-line bug fix (setting `out = None` when `out_` is provided). Removed unrelated formatting changes, Chinese documentation files, and hopper/flash.h comments that were accidentally included.

PR #2357, opened 2026-03-16
[Cute, Fwd] Add return_row_max to expose per-head max attention logits

### LINKED ISSUE #1876: Per Head Max-Attention-Logits
Muon optimizers are train and scale well for transformers. Recent kimi K2 paper introduced Muon with QK clipping which helps in stabilizing large models. 

To implement this, we need per head max attention logits. What’s the reliable way to get this information without sacrificing performance.

NJX-njx:
## Summary
- Add `return_row_max` parameter to `flash_attn_func` and `flash_attn_varlen_func` that returns per-head attention logit maximums (`softmax.row_max`) alongside the existing LSE output
- The row_max values are written to global memory using the same pattern as LSE in the forward epilogue
- Useful for QK clipping in optimizers like Muon (Kimi K2 paper)

## Details
- `row_max` values are the raw maximums **before** multiplying by `softmax_scale` (typically `1/sqrt(d)`)
- Same tensor shape as `lse`: `(batch, num_heads, seqlen_q)` or `(num_heads, total_q)` for varlen
- Supported on SM80, SM90, SM120. SM100/SM110 raises `NotImplementedError` (different epilogue system, will be addressed in a follow-up)

## Files changed
- `flash_attn/cute/interface.py` — parameter threading, tensor allocation, compile key, autograd functions
- `flash_attn/cute/flash_fwd.py` — base epilogue row_max write, SM80 `__call__`/`kernel` parameter threading
- `flash_attn/cute/flash_fwd_sm90.py` — SM90 `__call__`/`kernel`/`mma` parameter threading
- `flash_attn/cute/flash_fwd_sm100.py` — accept parameter (no-op, not yet implemented)

## Test plan
- [ ] Verify `return_row_max=True` returns correct per-head max logits on SM80/SM90/SM120
- [ ] Verify `return_row_max=False` (default) has no impact on existing behavior
- [ ] Verify SM100/SM110 raises `NotImplementedError`
- [ ] Compare row_max values against reference PyTorch computation: `(q @ k.T).max(dim=-1).values`

Closes #1876

### DIFF
--- a/flash_attn/cute/flash_fwd.py
+++ b/flash_attn/cute/flash_fwd.py
@@ -328,6 +328,8 @@ class FlashAttentionForwardBase:
         m_block: Int32,
         head_idx: Int32,
         batch_idx: Int32,
+        row_max: Optional[cute.Tensor] = None,
+        mRowMax: Optional[cute.Tensor] = None,
     ):
         # store acc_O
         rO = cute.make_fragment_like(acc_O, self.dtype)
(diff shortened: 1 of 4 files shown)


tridao:
This feature is not quite general enough yet so I don't think we'll be supporting this in the main repo.

NJX-njx:
Thanks for the feedback. Makes sense — just exposing row_max alone is a bit narrow.

Would a more general approach work better? e.g. a unified `return_attn_stats` that bundles row_max, entropy, and possibly other per-row statistics into a single optional output dict? That way we avoid adding one-off parameters for each stat.

Happy to rework this if there is a direction you would prefer, or close it if this is out of scope for now.

tridao:
This is out of scope at the moment. We support many attention variants but I don't think adding these justifies the complexity and maintenance burden.

PR #2360, opened 2026-03-17
[Fwd,Sm90] Add paged KV attention support (tma and cp.async)

henrylhtsang:
## Summary
Add paged KV cache support for SM90 forward via two paths:
- TMA path (page_size == n_block_size): fast, uses TMA with page index indirection
- cp.async path (page_size != n_block_size): general, uses PagedKVManager with PipelineAsync

Also fixes hardcoded `128` → `tile_n` in compile key for paged KV non-TMA boundary (affects SM100 too).

The hard part was `pipeline_k` was also used to count transaction for Q. 

## Test plan
```bash
pytest -n 64 -x tests/cute/test_flash_attn.py -k "test_flash_attn_kvcache"

# 2496 passed, 576 skipped
```

### DIFF
--- a/flash_attn/cute/flash_fwd_sm90.py
+++ b/flash_attn/cute/flash_fwd_sm90.py
@@ -668,8 +668,12 @@ class FlashAttentionForwardSm90(FlashAttentionForwardBase):
                         gV = cute.local_tile(mV_cur, (self.tile_n, self.tile_hdimv), (0, 0, None))
                     else:
                         # Non-paged TMA
-                        mK_cur = seqlen.offset_batch_K(mK, batch_idx, dim=3)[None, None, head_idx_kv]
-                        mV_cur = seqlen.offset_batch_K(mV, batch_idx, dim=3)[None, None, head_idx_kv]
+                        mK_cur = seqlen.offset_batch_K(mK, batch_idx, dim=3)[
+                            None, None, head_idx_kv
+                        ]
+                        mV_cur = seqlen.offset_batch_K(mV, batch_idx, dim=3)[
+                            None, None, head_idx_kv
+                        ]
                         gK = cute.local_tile(mK_cur, (self.tile_n, self.tile_hdim), (None, 0))
                         gV = cute.local_tile(mV_cur, (self.tile_n, self.tile_hdimv), (None, 0))
                     # TODO: mcast
@@ -702,12 +706,20 @@ class FlashAttentionForwardSm90(FlashAttentionForwardBase):
                     )
 
                 load_K = partial(
-                    self.load_KV, tma_load_K_fn, paged_kv_manager, sK,
-                    pipeline_kv=pipeline_k, K_or_V="K",
+                    self.load_KV,
+                    tma_load_K_fn,
+                    paged_kv_manager,
+                    sK,
+                    pipeline_kv=pipeline_k,
+                    K_or_V="K",
                 )
                 load_V = partial(
-                    self.load_KV, tma_load_V_fn, paged_kv_manager, sV,
-                    pipeline_kv=pipeline_v, K_or_V="V",
+                    self.load_KV,
+                    tma_load_V_fn,
+                    paged_kv_manager,
+                    sV,
+                    pipeline_kv=pipeline_v,
+                    K_or_V="V",
                 )
 
                 if const_expr(not self.use_block_sparsity):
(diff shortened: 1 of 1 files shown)


tridao:
Should we switch so that pipeline_k no longer counts transaction for Q? Would that simplify things?

henrylhtsang:
> Should we switch so that pipeline_k no longer counts transaction for Q? Would that simplify things?

yeah that would likely make things more clear. Let me give it a try

henrylhtsang:
Ready for another look. It simplifies the logic quite a bit. 

I checked perf roughly, separating Q from pipelike_k is roughly perf neutral.

flash_attn/cute/flash_fwd_sm90.py
```diff
@@ -430,30 +443,58 @@ def kernel(
             #     cute.arch.mbarrier_init(mbar_ptr_Q + tidx, 1 if tidx == 0 else self.num_mma_threads)
             if const_expr(not self.use_tma_Q):
                 cute.arch.mbarrier_init(mbar_ptr_Q, self.num_Q_load_threads)
```
tridao: can use `1 if const_expr(self.use_tma_Q) else self.num_Q_load_threads`

flash_attn/cute/flash_fwd_sm90.py
```diff
@@ -430,30 +443,58 @@ def kernel(
...
+            pipeline_v = pipeline.PipelineTmaAsync.create(
+                barrier_storage=storage.mbar_ptr_V.data_ptr(),
+                num_stages=self.num_stages,
+                producer_group=pipeline_kv_producer_group,
+                consumer_group=pipeline_kv_consumer_group,
+                tx_count=self.tma_copy_bytes["V"],
+                defer_sync=False,
+            )
+        else:
+            # PipelineAsync: no thread gating in producer_commit/consumer_release,
```
tridao, line 476: We should eventually implement a variant of PipelineAsync that does thread gating (i.e. only 1 thread out of 32 will signal). But that's for later

flash_attn/cute/flash_fwd_sm90.py
```diff
@@ -571,19 +615,29 @@ def load(
...
         SeqlenInfoCls: Callable,
         TileSchedulerCls: Callable,
     ):
         warp_idx_in_wg = cute.arch.make_warp_uniform(cute.arch.warp_idx()) % 4
-        if warp_idx_in_wg == 0:
+        tidx, _, _ = cute.arch.thread_idx()
+
+        if const_expr(self.use_tma_KV):
+            # TMA path: only warp 0 does loads
+            is_load_thread = warp_idx_in_wg == 0
```
tridao: this should be called `is_load_warp` then

tridao: `TMA: only warp 0 loads. cp_async: all warps load`
`is_load_warp = warp_idx_in_wg == 0 or not self.use_tma_KV`

flash_attn/cute/flash_fwd_sm90.py
```diff
@@ -598,82 +652,145 @@ def load(
...
+                        gK = cute.local_tile(mK_cur, (self.tile_n, self.tile_hdim), (0, 0, None))
+                        gV = cute.local_tile(mV_cur, (self.tile_n, self.tile_hdimv), (0, 0, None))
+                    else:
+                        # Non-paged TMA
+                        mK_cur = seqlen.offset_batch_K(mK, batch_idx, dim=3)[None, None, head_idx_kv]
+                        mV_cur = seqlen.offset_batch_K(mV, batch_idx, dim=3)[None, None, head_idx_kv]
+                        gK = cute.local_tile(mK_cur, (self.tile_n, self.tile_hdim), (None, 0))
+                        gV = cute.local_tile(mV_cur, (self.tile_n, self.tile_hdimv), (None, 0))
+                    # TODO: mcast
+                    # TODO check warp_idx if we have 128 producer threads
```
tridao: can remove this TODO now

tridao (approved):

tridao:
Some nits but other than that LGTM. We can merge when it's ready.

flash_attn/cute/flash_fwd_sm90.py
henrylhtsang: noob question: how will that work for cp.async? something like a syncwarp then one thread to signal arrival?

henrylhtsang:
addressed comments. should be good to go.

flash_attn/cute/flash_fwd_sm90.py
tridao: nvm producer commit would need all threads to commit (not just 1), but consumer release we can do sync_warp then elect one thread to signal.

tridao:
Thank you!

PR #2365, opened 2026-03-17
[Sm90] Fix test_mask_mod and bwd block-sparse kwarg mismatch

henrylhtsang:
## Summary

- **test_mask_mod.py:** Update `_flash_attn_fwd` calls to use `tile_mn=(m, n)` instead of removed `m_block_size`/`n_block_size` params (broken since [`99d0148`](https://github.com/Dao-AILab/flash-attention/commit/99d0148)). Fixes 1,714 failures.
- **flash_bwd_sm90.py:** Fix kwarg `num_wg_mma` → `num_mma_warp_groups` in `dQaccum_store_block_sparse_bwd_sm90` call to match function signature in `block_sparse_utils.py`. (introduced in https://github.com/Dao-AILab/flash-attention/commit/c325b5b)

### Not fixed

Vectorized score_mod tests in `test_score_mod.py` / `test_score_mod_varlen.py` (917 failures) produce substantially different results on SM90 (max diff ~3.0). Never worked since introduction ([`c4d8b06`](https://github.com/Dao-AILab/flash-attention/commit/c4d8b06)).

## Test plan

```bash
cd tests/cute
pytest test_mask_mod.py::test_static_masks -k "dtype0-128" -q             # 60 passed
pytest test_mask_mod.py::test_parameterized_masks -k "dtype0" -q          # 360 passed
pytest test_mask_mod.py::test_q_boundary_masking_block_sparse_bwd -q      # 14 passed
```


tridao (approved):

drisspg:
cc @reubenconducts on the vectorized score mods

reubenconducts:
The problem is the thread ownership pattern of the WGMMA accumulator. The easiest fix would be to limit `vec_size` to 2 on Sm90, but theoretically one could be quite smart about swizzling `aux_tensors` so that one thread can issue a vectorized load. @tridao @drisspg what do you think?

tridao:
vec_size=2 might be good enough for perf? I don't have good intuition here about how important vectorized load is.

PR #2371, opened 2026-03-18
[CuTe,Flex] limit vec_size to 2 for score mod when not on Sm100

reubenconducts:
This addresses the discussion in #2365: `vec_size > 2` is generically only supported on Sm100 due to accumulator thread ownership pattern. Tests modified to skip unsupported  `vec_size` while still checking supported ones.

Theoretically we could write architecture-specific vectorized score mods and lift this restriction, but I don't see that being worth the hassle at the moment. 

cc @drisspg

### DIFF
--- a/tests/cute/test_score_mod.py
+++ b/tests/cute/test_score_mod.py
@@ -107,7 +107,7 @@ SEQLEN_CONFIGS = [
     (4224, 4224),
 ]
 
-VEC_SIZES_TO_CHECK_EQUALITY = [1, 2, 4]
+VEC_SIZES_TO_CHECK_EQUALITY = [1, 2, 4] if COMPUTE_CAPABILITY == 10 else [1, 2]
 
 
 def create_tensors(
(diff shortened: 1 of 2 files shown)


tests/cute/test_score_mod.py
```diff
@@ -224,9 +224,13 @@ def test_cute_score_mod_vectorized(
 
     for vec_size in VEC_SIZES_TO_CHECK_EQUALITY:
         cute_vectorized_score_mod.__vec_size__ = vec_size
-        out = run_cute_flash(q, k, v, cute_vectorized_score_mod, pack_gqa=pack_gqa)
-
-        assert torch.equal(out, out_ref)
+        try:
+            out = run_cute_flash(q, k, v, cute_vectorized_score_mod, pack_gqa=pack_gqa)
+            assert torch.equal(out, out_ref)
+        except ValueError as e:
```
tridao: Would this hide errors?

reubenconducts: fair point, handled properly now.

Issue #1533, opened 2025-03-12
FlashAttention forward support for Turing


ssiu:
Hi, I tried my hands on implementing flash attention forward pass for the Turing architecture. This is the repo:

https://github.com/ssiu/flash-attention-turing

As this is still an early implementation, it only supports:
- ```head_dim = 128```
- vanilla attention i.e. no masking
- ```seq_len``` must be divisible by 128

For ```batch_size = 4, num_heads = 32, head_dim = 128```, our implementation is currently around 2x faster than Pytorch's ```F.scaled_dot_product_attention```, which calls ```Memory-Efficient Attention``` in the backend. This was tested for T4.

![Image](https://github.com/user-attachments/assets/0cb3b053-30bc-46cb-9c2f-7063d7df11a1)

For long sequences it reaches around 63% compute throughput. 

![Image](https://github.com/user-attachments/assets/72519dcf-a969-4c6b-a974-d4b5a1a327b9)

Thanks!

tridao:
Wow this is great work!

OMS24:
Thanks for this, I has been trying to find a way to get flash attention working on T4 GPU. 

You help me so much.

lwllvyb:
This is great work! I am looking for this. Is there any new progress?

ssiu:
Hi @lwllvyb, yes I will be continuing to work on this.

wangsen-zy:
Great, praise your work!

michaelshekasta:
@ssiu This is awesome! I can’t wait for this to be official. You’re doing incredible work - keep going, you’re amazing! 🙌💪

PR #2379, opened 2026-03-21
[Sm75] Add README link for initial Turing support

ssiu:
# FlashAttention Turing

This PR adds a link to the [flash-attention-turing](https://github.com/ssiu/flash-attention-turing) repo that provides support for Turing (SM75) architecture in FlashAttention, following #1533.

## Features

Supports:

 - fwd and bwd
 - head dim 64, 128
 - causal mask
 - grouped-query attention (GQA)
 - variable sequence lengths (varlen)

Does not support:

 - dropout
 - local mask
 - kv cache

## Performance

Benchmarks are reported on Nvidia T4 GPUs.

### Forward pass 

Up to 2.19x and 1.95x faster than PyTorch's [Attention](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html) for non-causal and causal workloads.

On Turing GPUs, PyTorch's Attention uses Memory-Efficient Attention from [xformers](https://github.com/facebookresearch/xformers), since FlashAttention does not provide optimized kernels for SM75.

For long sequences, the forward kernel reaches up to 66% compute throughput.

Forward pass benchmark for head dimension 128
Forward pass benchmark for head dimension 64

### Backward pass 

The backward pass is split into two kernels: one for `dQ` and one for `dK` and `dV`.

Up to 1.35x and 1.51x faster than PyTorch's Attention for non-causal and causal workloads.

For long sequences, the backward kernels reach up to 49% compute throughput for `dK` and `dV`, and 45% for `dQ`.

Backward pass benchmark for head dimension 128
Backward pass benchmark for head dimension 64

## Correctness and numerical differences

From our tests in `test_flash_attn.py`, we consistently observe maximum and mean absolute differences of ~1e-3 and ~1e-5 respectively relative to PyTorch's attention kernels.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -122,9 +122,7 @@ We recommend the
 container from Nvidia, which has all the required tools to install FlashAttention.
 
 FlashAttention-2 with CUDA currently supports:
-1. Ampere, Ada, or Hopper GPUs (e.g., A100, RTX 3090, RTX 4090, H100). Support for Turing
-   GPUs (T4, RTX 2080) is coming soon, please use FlashAttention 1.x for Turing
-   GPUs for now.
+1. Ampere, Ada, or Hopper GPUs (e.g., A100, RTX 3090, RTX 4090, H100). For Turing GPUs (T4, RTX 2080), see the separate [flash-attention-turing](https://github.com/ssiu/flash-attention-turing) repo, which supports a core subset of FlashAttention features on Turing.
 2. Datatype fp16 and bf16 (bf16 requires Ampere, Ada, or Hopper GPUs).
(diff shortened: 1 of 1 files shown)


tridao:
Thanks!
Given that we're switching to Cute-DSL for current and future development, we won't support Turing in this repo (cute-dsl requires Ampere+). You're welcome to have the Turing impl in another repo and we could link to it.

ssiu:
Thanks for the comment!

Sure, I’ll keep the Turing implementation in a separate repo and share it here once it’s cleaned up.

ssiu:
Hi @tridao ,

I just cleaned up the Turing repo. I think it's good to go now. Thanks again!

PR #2388, opened 2026-03-24
fix: use LSE accum strides from params instead of hardcoded ones

ZeronSix:
In the Split-KV path, the forward kernel computes LSE accumulator addresses using hardcoded strides instead of the stride values provided in the params structure. The combine kernel already uses the explicit strides from params, so this creates an inconsistency between the two kernels.

As a result, when the caller supplies an LSE accumulator layout that differs from the layout assumed by the forward kernel, the forward pass writes to incorrect locations and produces wrong output.

This change updates the forward kernel to use the LSE accumulator strides from params, matching the behavior of the combine kernel and ensuring correct results for arbitrary accumulator layouts.


tridao:
Thanks!

PR #2389, opened 2026-03-25
Add SM80/SM120 block-sparse forward attention support

blake-snc:
## Summary

- Implements block-sparse forward attention in the SM80 base-class kernel (`flash_fwd.py`), which is shared by SM80 and SM120
- Two block types: `mask_blocks` (partially masked, apply `mask_mod` per element) and `full_blocks` (fully unmasked, skip masking entirely)
- SM120 (`FlashAttentionForwardSm120`) inherits the SM80 base class and gets block sparsity for free; removes the `assert not use_block_sparsity` guard in `interface.py`
- Follows the same `mma_one_n_block` callback pattern as SM90/SM100

### mma_one_n_block callback pattern (mirrors SM90)

A new `mma_one_n_block_bs` method handles one KV block: load K, load V, GEMM QK, optional score_mod, mask, online softmax, GEMM PV. This mirrors SM90's `mma_one_n_block`.

A new `run_block_sparse_mainloop_sm80` utility in `block_sparse_utils.py` takes the callback and iterates mask blocks (applying `mask_mod`) then full blocks (seqlen masking only). This mirrors the non-intra-wg-overlap path of `consume_block_sparse_loads`.

SM80 can't use the exact TMA producer/consumer split (no warpgroup specialization), but the `mma_one_n_block` callback abstraction is the same.

### First full block seqlen masking

The first full block always receives `mask_seqlen=True` in `run_block_sparse_mainloop_sm80`, since full blocks may sit at a higher n position than any mask block (and thus need seqlen-boundary masking).

### SM120 arch fix

`FlashAttentionForwardSm120.__init__` now forces `self.arch = Arch.sm_80`. Without this, `FlashAttentionForwardBase.__init__` sets `self.arch` from the real GPU arch (sm_121a on DGX Spark), which causes the SM80 epilogue to incorrectly attempt TMA-O — crashing since `tma_atom_O=None` in this kernel variant.

### Dense mainloop unchanged

The original dense mainloop is guarded by `if const_expr(blocksparse_tensors is None):` and is completely unmodified.

## Validation

Validated on SM121a (DGX Spark GB10):
- `test_block_sparsity.py`: **4621 passed, 40 skipped** (causal, sliding window, block diagonal; bf16/fp16; D=64/128/256; various seqlens)

### DIFF
--- a/flash_attn/cute/flash_fwd.py
+++ b/flash_attn/cute/flash_fwd.py
@@ -33,7 +33,7 @@ from flash_attn.cute.block_info import BlockInfo
 from flash_attn.cute.pack_gqa import PackGQA
 from flash_attn.cute.named_barrier import NamedBarrierFwd
 from flash_attn.cute.block_sparsity import BlockSparseTensors
-from flash_attn.cute.block_sparse_utils import sparse_tensor_m_block, run_block_sparse_mainloop_sm80
+from flash_attn.cute.block_sparse_utils import run_block_sparse_mainloop_sm80, get_total_block_count
 from flash_attn.cute.tile_scheduler import SingleTileScheduler, SingleTileVarlenScheduler, TileSchedulerArguments
 
 
@@ -954,11 +954,8 @@ class FlashAttentionForwardSm80(FlashAttentionForwardBase):
             # ///////////////////////////////////////////////////////////////////////////////
             qkv_factor = self.qhead_per_kvhead if const_expr(self.pack_gqa) else 1
             subtile = self.q_subtile_factor if self.q_subtile_factor is not None else 1
-            bs_m = sparse_tensor_m_block(m_block, qkv_factor, subtile)
-            total_block_cnt = blocksparse_tensors[0][batch_size, num_head, bs_m] + (
-                blocksparse_tensors[2][batch_size, num_head, bs_m]
-                if const_expr(blocksparse_tensors[2] is not None)
-                else Int32(0)
+            total_block_cnt = get_total_block_count(
+                blocksparse_tensors, batch_size, num_head, m_block, qkv_factor, subtile
             )
 
             bs_mask = AttentionMask(
--- a/flash_attn/cute/flash_fwd.py
+++ b/flash_attn/cute/flash_fwd.py
@@ -33,6 +33,7 @@ from flash_attn.cute.block_info import BlockInfo
 from flash_attn.cute.pack_gqa import PackGQA
 from flash_attn.cute.named_barrier import NamedBarrierFwd
 from flash_attn.cute.block_sparsity import BlockSparseTensors
+from flash_attn.cute.block_sparse_utils import sparse_tensor_m_block, run_block_sparse_mainloop_sm80
 from flash_attn.cute.tile_scheduler import SingleTileScheduler, SingleTileVarlenScheduler, TileSchedulerArguments
 
 
@@ -728,6 +729,7 @@ class FlashAttentionForwardSm80(FlashAttentionForwardBase):
             TileScheduler,
             aux_tensors,
             fastdiv_mods,
+            blocksparse_tensors,
         ).launch(
             grid=grid_dim,
             block=[self.num_threads, 1, 1],
@@ -767,6 +769,7 @@ class FlashAttentionForwardSm80(FlashAttentionForwardBase):
         TileScheduler: cutlass.Constexpr[Callable],
         aux_tensors=None,
         fastdiv_mods=None,
+        blocksparse_tensors: Optional[BlockSparseTensors] = None,
     ):
         # Thread index, block index
         tidx, _, _ = cute.arch.thread_idx()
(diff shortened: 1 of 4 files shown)


flash_attn/cute/flash_fwd.py
reubenconducts, line 803: Can you rebase on main? Varlen was already merged in #2333, as you know.

flash_attn/cute/flash_fwd.py
```diff
@@ -945,134 +948,425 @@ def kernel(
...
-        self.load_Q(gmem_thr_copy_Q, gQ, sQ, m_block, seqlen=seqlen.seqlen_q, headdim=mQ.shape[1])
-        cute.arch.cp_async_commit_group()
+        if const_expr(blocksparse_tensors is not None):
+            # ///////////////////////////////////////////////////////////////////////////////
+            # Block-sparse mainloop (SM80/SM120)
+            # ///////////////////////////////////////////////////////////////////////////////
+            # Unpack sparse block lists for this (batch, head, m_block) tile.
+            # mask_blocks = partially-masked KV blocks (need mask applied).
+            # full_blocks = fully-unmasked KV blocks (no masking needed).
+            _bs_mask_cnt, _bs_mask_idx, _bs_full_cnt, _bs_full_idx = blocksparse_tensors
```
reubenconducts: Please keep variable names consistent with the other kernels. Is there a reason not to use the same design pattern for handling blocksparsity as in the others, e.g. https://github.com/Dao-AILab/flash-attention/blob/b8eda39e6ea9b2264aa30d7c7c2e227eec62da73/flash_attn/cute/flash_fwd_sm90.py#L880?

flash_attn/cute/flash_fwd.py
```diff
@@ -945,134 +948,425 @@ def kernel(
...
+                    cute.arch.barrier()
+                else:
+                    cute.arch.cp_async_wait_group(0)
+                    cute.arch.barrier()
+
+                # === Process mask blocks (partial masking) ===
+                # Indices stored in decreasing order: highest n_block first.
+                # First mask block: is_first=True, mask_seqlen=True, mask_mod=self.mask_mod.
+                # Remaining mask blocks: is_first=False, mask_seqlen=False, mask_mod=self.mask_mod.
+                # CuTe DSL forbids closures in dynamic CF, so block processing is inlined at
```
reubenconducts: Unnecessary comment on CuTe DSL

flash_attn/cute/flash_fwd.py
```diff
@@ -945,134 +948,425 @@ def kernel(
...
-            thr_mma=thr_mma_qk,
-            mask_causal=self.is_causal,
-            mask_local=self.is_local,
-            aux_tensors=aux_tensors,
-            fastdiv_mods=fastdiv_mods if const_expr(self.mask_mod is not None) else None,
-        )
+                # === Process full blocks (no masking) ===
+                # When no mask blocks preceded: first full block is the first block overall
+                # (is_first=True) and gets mask_seqlen=True for seqlen_k boundary check.
+                # When mask blocks preceded: all full blocks have is_first=False.
```
reubenconducts: First full block must always get seqlen masking, since it is possible that all mask blocks are further left than some full block.

reubenconducts (changes requested):

blake-snc:
Thanks for the review! Here's a summary of what changed in the v2 rewrite:

**1. Rebased on main** ✅

**2. Design pattern matching SM90** ✅ The block-sparse mainloop now follows the same `mma_one_n_block` callback pattern as SM90/SM100:
- New `mma_one_n_block_bs` method handles one KV block (load K, load V, GEMM QK, score_mod, mask, softmax, GEMM PV) — mirrors SM90's `mma_one_n_block`
- New `run_block_sparse_mainloop_sm80` utility in `block_sparse_utils.py` takes a `mma_one_n_block` callable and iterates mask blocks then full blocks — mirrors SM90's `consume_block_sparse_loads` (non-intra-wg-overlap path)

SM80 can't use the exact TMA producer/consumer split since it has no warpgroup specialization, but the `mma_one_n_block` callback abstraction is the same.

**3. CuTe DSL comment removed** ✅ (was in the old version)

**4. First full block seqlen masking** ✅ In `run_block_sparse_mainloop_sm80`, the first full block always gets `mask_seqlen=True` regardless of whether mask blocks were processed first.

Also fixed two bugs found during validation on SM121a:
- `blocksparse_tensors` was missing from the `kernel()` method signature
- `FlashAttentionForwardSm120.__init__` now forces `self.arch = Arch.sm_80` to prevent the SM80 epilogue from incorrectly using TMA-O (which crashes on SM121a since `tma_atom_O=None` in this variant)

Validated: `test_block_sparsity.py` — 4621 passed, 40 skipped on SM121a.

johnnynunez:
ping @drisspg @tridao

blake-snc:
Thanks for the review @reubenconducts, and apologies for the long gap. Rebased onto current `main` and reworked the implementation against where the block-sparse code has landed since. Summary of how each point was addressed:

**Rebased on main (varlen #2333, etc.)** — done; the PR is now a single forward-only addition on top of current `main`.

**Design pattern / consistency with `flash_fwd_sm90.py`** — this needed a caveat I should flag explicitly. `flash_fwd_sm90.py` (and sm100) drive block-sparse through `produce_block_sparse_loads` / `consume_block_sparse_loads`, which are warp-specialized (producer/consumer warp-groups + the mbarrier handshake described in the `block_sparse_utils` SM100 note). The Sm80/Sm120 forward is the Ampere `cp.async` pipeline — there's no producer warp to run those helpers, so they don't port 1:1. What *is* shared is the block-list model, so the implementation:
- reuses the same `get_curr_blocksparse_tensors` accessor and the mask/full block split;
- adds `run_block_sparse_mainloop_sm80` in `block_sparse_utils.py` that mirrors the masking contract of `consume_block_sparse_loads` exactly (mask blocks first with `mask_mod` + seqlen masking, then full blocks with seqlen masking only);
- uses a per-block `mma_one_n_block_bs` (load K/V → QK → mask → softmax → PV) instead of the dense `compute_one_n_block`, since sparse blocks aren't contiguous so the dense path's cross-block prefetch doesn't apply.

Naming now matches the other kernels (`use_block_sparsity`, `curr_mask_block_idx`, etc.). Open to restructuring further if you'd prefer a different split.

**First full block must always get seqlen masking** — fixed, and it matches `consume_block_sparse_loads`: the first full block gets `mask_seqlen=True` whether or not mask blocks preceded it (a full block can sit at the seqlen_kv boundary regardless of mask-block positions); it only gets `is_first_n_block=True` when there were no mask blocks.

**Unnecessary CuTe DSL comment** — gone in the reworked version.

**Validation** — on DGX Spark (GB10, `sm_121a`), ~493 cases from `tests/cute/test_block_sparsity.py` pass (fixed-length block_diagonal/mini_causal across tiles/batch/heads/non-square shapes, larger seqlens, edge cases, and the parameterized sliding-window masks). The dense path is untouched by this change (block-sparse runs as a separate branch behind `use_block_sparsity`). Happy to add CI coverage notes or run anything else on hardware.

blake-snc:
Status update: our SM120 hardware is currently allocated to other work, so I am unable to re-validate this for the time being. The branch currently applies cleanly against main, and the gap it fills is still present (block sparsity is asserted off for SM 12.0 on current main). Glad to see it adopted by anyone with SM120-class hardware; for design questions I am also reachable at github.com/bledden.

PR #2390, opened 2026-03-25
[Cute,Sm100,Bwd] refine bwd swizzle for deterministic

jayhshah:
This PR fixes the varlen swizzle num_n_blocks from num_m_blocks calculation for 2cta and includes dqaccum in the head size for not spilling l2. This noticeably improves backward deterministic FLOPs (for non-varlen, limited to non-causal but doesn't appear to cause regression for causal; for varlen, improve perf across the board thanks to the fix).

Example benchmarks:

non-varlen deterministic, MHA nheads = 16
```
hdim causal batch seqlen     PR          		MAIN
----------------------------------------------------------------
128  False     4   8192      4.58/1200/53.3% 	4.75/1157/51.4%
128  False     2  16384      9.32/1180/52.4%	10.08/1091/48.5%
128  False     1  32768     19.69/1117/49.6%	20.48/1074/47.7%

128   True     4   8192      2.50/1100/48.9%	2.50/1100/48.9%
128   True     2  16384      4.60/1195/53.1%	4.60/1195/53.1%
128   True     1  32768      9.74/1129/50.2%	9.86/1115/49.6%
```

varlen deterministic, MHA nheads = 16
```
hdim causal batch seqlen     PR          		MAIN
----------------------------------------------------------------
128  False     4   8192      4.66/1180/52.5%	5.13/1071/47.6%
128  False     2  16384      8.97/1225/54.5%	10.15/1083/48.1%
128  False     1  32768     19.45/1131/50.3%	20.85/1054/46.9%

128   True     4   8192      2.59/1061/47.1%	2.83/973/43.2%
128   True     2  16384      4.72/1166/51.8%	5.06/1086/48.3%
128   True     1  32768      9.72/1131/50.3%	10.09/1089/48.4%
```

### DIFF
--- a/flash_attn/cute/tile_scheduler.py
+++ b/flash_attn/cute/tile_scheduler.py
@@ -395,8 +395,8 @@ class SingleTileLPTBwdScheduler:
         ) -> "SingleTileLPTBwdScheduler.Params":
             size_l2 = 50 * 1024 * 1024
             size_one_qdo_head = args.seqlen_k * (args.headdim + args.headdim_v) * args.element_size
-            # size_one_dqaccum_head = args.seqlen_k * (args.headdim) * 4
-            size_one_dqaccum_head = 0
+            size_one_dqaccum_head = args.seqlen_k * (args.headdim) * 4
+            # size_one_dqaccum_head = 0
             size_one_head = size_one_qdo_head + size_one_dqaccum_head
             log2_floor = lambda n: 31 - clz(n)
             swizzle = 1 if size_l2 < size_one_head else (1 << log2_floor(size_l2 // size_one_head))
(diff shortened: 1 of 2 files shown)


flash_attn/cute/tile_scheduler.py
```diff
@@ -654,6 +657,7 @@ def get_current_work(self, *, loc=None, ip=None) -> WorkTileInfo:
                 num_n_blocks = (
                     num_m_blocks
                     * params.tile_shape_mn[0]
+                    * params.cluster_shape_m
```
tridao, line 660: does this affect any of the 2cta bwd code?

jayhshah: This change is meant to get the right head swizzle heuristic for 2cta bwd, by accounting for the num_m_blocks being defined with respect to tiler divided by the cluster shape.

num_n_blocks here is only used to derive nheads_in_l2, so it doesn't affect correctness.

tridao: For the `tile_shape_mn[0]` do we pass the CTA's tile shape or the cluster tile shape? I think we had this discussion and realized we have not been consistent.
In any case if it doesn't affect correctness it's fine w me.

jayhshah: Currently for the tile scheduler args we pass `cta_tiler[:2]` as `tile_shape_min` and `cluster_shape_mn` as a separate parameter; this makes the most sense to me since cluster shape is in principle separate from use of 2cta mma.

PR #2391, opened 2026-03-25
[Cute,Sm80,Fwd] Guard kernel body with tile validity check to fix varlen illegal memory access

zhuochenKIDD:
On sm80 varlen fwd, when exec pytext tests/cute/test_flash_attn_varlen.py, FA crashes because SingleTileVarlenScheduler's get_grid_shape provides some non-valid block, leads to illegal memory access.

also change to persistent tile loop style same as sm90/sm100 prehaps enables future optimization on sm80


zhuochenKIDD:
@tridao hi would you take a look at this fix?

PR #2393, opened 2026-03-25
Add FA4 CI: GitHub Actions workflow with Apptainer on B200 runner

Johnsonms:
- GitHub Actions workflow (ci.yml) triggering on push to main
- Two-pass test strategy: kernel compilation (FakeTensorMode) + GPU execution
- Pulls Docker image from Docker Hub (togethercomputer/training-performance)
  and converts to Apptainer SIF, cached by image tag on the runner
- CI_WORK_DIR repo variable for configurable large-disk path on the runner
- Shared Python driver (tools/ci/run_fa4_ci.py) used by both CI and local runs
- Docker image build scripts and Apptainer SIF definition (tools/ci/)
- CI_SETUP.md setup guide covering runner registration, secrets, and migration


AI/CI_SETUP.md
```diff
@@ -0,0 +1,242 @@
...
+cd ~/scratch/flash-attention
+uv venv .venv --python 3.12
+source .venv/bin/activate
+
+# Install PyTorch nightly matching your CUDA driver
+# Check your CUDA version first: cat /usr/local/cuda/version.json
+uv pip install --pre torch --index-url https://download.pytorch.org/whl/nightly/cu129
+
+# Install flash-attn-4 + dev dependencies
+uv pip install -e "flash_attn/cute[dev]"
```
drisspg: There is unfortunately a large delta in perf (depending on cases) between the 12.9. I dont think this effects correctness in any way fwiw but curious if there are plans for newer driver version (beggars cant be choosers though )

Johnsonms: Thanks for the note! The target CI machine is running driver 575.57 with CUDA 13.0 (B200) after upgrading last Friday, so we can upgrade to cu130. Will rebuild the container image and update the references — will push an update to this PR.

tools/ci/docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+
+# Install torch from the published cu130 nightly package page and let pip resolve
+# the matching nightly Triton wheel from PyTorch's nightly triton page.
+RUN uv pip install --system --break-system-packages --no-cache --pre \
+        "torch>=2.12.0.dev0,<2.13.0" \
+        --find-links https://download.pytorch.org/whl/nightly/cu130/torch \
+        --find-links https://download.pytorch.org/whl/nightly/triton
+
+# FA4 dependencies — torch stays pinned from the direct wheel install above.
+RUN uv pip install --system --break-system-packages --no-cache \
```
drisspg: can we just rely on the install for fa-4 for this

tools/ci/docker/Dockerfile
```diff
@@ -0,0 +1,35 @@
...
+    rm -rf /var/lib/apt/lists/*
+
+# Install uv
+RUN curl -LsSf https://astral.sh/uv/install.sh | sh
+
+# Install torch from the published cu130 nightly package page and let pip resolve
+# the matching nightly Triton wheel from PyTorch's nightly triton page.
+RUN uv pip install --system --break-system-packages --no-cache --pre \
+        "torch>=2.12.0.dev0,<2.13.0" \
+        --find-links https://download.pytorch.org/whl/nightly/cu130/torch \
```
drisspg, line 20: should we be installing nighlty

tools/ci/run_fa4_ci.py
```diff
@@ -0,0 +1,211 @@
...
+from __future__ import annotations
+
+import argparse
+import os
+import shutil
+import subprocess
+from dataclasses import dataclass
+from pathlib import Path
+
+DEFAULT_TEST_FILTER = "1-1-128-True-0-0.0-False-False-False-mha-dtype0"
```
drisspg: maybe we dont hardocde this in the file but set this perhaps in the acutal ci.yml

tools/ci/run_fa4_ci.py
```diff
@@ -0,0 +1,211 @@
...
+    mode is 'apptainer' or 'venv'.
+    python_bin is None for apptainer (container provides python3).
+    """
+    if sif and shutil.which("apptainer"):
+        return "apptainer", None
+    if sif and not shutil.which("apptainer"):
+        print(f"WARNING: FA4_SIF is set ({sif}) but apptainer is not on PATH — falling back to venv")
+
+    # venv mode
+    venv_path = venv if (venv and venv.is_absolute()) else repo_root / (venv or Path(".venv"))
```
drisspg: do we even need two modes? this file is only expected to run in apptainer right? maybe this was for testing

tools/ci/run_fa4_ci.py
```diff
@@ -0,0 +1,211 @@
...
+        Step(
+            name="Pass 2: run tests on GPU",
+            command=[*pytest_base, "-n", str(run_workers), "-x"],
+            extra_env={
+                "FLASH_ATTENTION_FAKE_TENSOR": "0",
+                "CUDA_VISIBLE_DEVICES": test_visible_devices,
+            },
+        ),
+    ]
+    if not skip_benchmark:
```
drisspg, line 88: is it easy to get/parse these numbers back from the run

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
+name: CI
+
+on:
+  push:
+    branches: [main]
+
+env:
```
drisspg, line 10: Can we explicitly set workflow permissions to read-only here, e.g. permissions: { contents: read }? Since this runs on a self-hosted runner and doesn’t need to write back to GitHub, I’d rather not inherit the default GITHUB_TOKEN scopes or accidentally grant write access later.\n\nWe can suggest .

.github/workflows/ci.yml
```diff
@@ -0,0 +1,38 @@
...
+  FA4_IMAGE: togethercomputer/training-performance:flash-attn-cu13.0-26.03.31
```
drisspg: can  pin FA4_IMAGE by full image digest, e.g. togethercomputer/training-performance@sha256:...

tools/ci/run_fa4_ci.py
```diff
@@ -0,0 +1,211 @@
...
+        install_cmd = f"uv pip install --system --break-system-packages --no-deps -q -e {repo_root}/flash_attn/cute"
+
+        # Convert relative test/benchmark paths to absolute so we can run from /tmp.
+        # Running from /tmp ensures Python does not insert repo_root into sys.path[0]
+        # (which would cause flash_attn/__init__.py to trigger FA2 imports unavailable in the SIF).
+        command = [
+            str(repo_root / arg) if (arg.startswith("tests/") or arg.startswith("benchmarks/")) else arg
+            for arg in step.command
+        ]
+        env_exports = " && ".join(f"export {k}={v}" for k, v in step.extra_env.items())
```
drisspg: If we keep bash -c, should we at least use shlex.join / shlex.quote more here instead of plain string concatenation?

.github/workflows/ci.yml
Johnsonms: Added at the workflow level in ci.yml. Good catch

.github/workflows/ci.yml
Johnsonms: Pinned FA4_IMAGE to @sha256:07016df... in ci.yml.

tools/ci/run_fa4_ci.py
Johnsonms: Fixed in run_fa4_ci.py. Was a real correctness issue if paths ever contain spaces.

tools/ci/run_fa4_ci.py
Johnsonms: Moved to an explicit --test-filter arg in action.yml. The Python file now defaults to "" (run all), keeping local usage flexible.

tools/ci/docker/Dockerfile
Johnsonms: Good point — the duplication is worth fixing. The tricky part is that --writable-tmpfs gives each step a fresh overlay, so deps installed at step time don't persist to the next step. That's why we pre-bake them into the image and use --no-deps at runtime.

tools/ci/docker/Dockerfile
Johnsonms: This is intentional — CUDA 13.0 (cu130) isn't available in a stable PyTorch release yet, so nightly is currently the only option for B200/SM100 support. We'll switch to stable as soon as a cu130 stable wheel ships. The Docker image can be built monthly or serval monthly or consider maintaining a shared dev Docker image (similar to SGLang) to simplify environment consistency and upgrades.

tools/ci/run_fa4_ci.py
Johnsonms: Good call — dropped venv mode entirely. The script is now apptainer-only; it exits immediately with a clear error if FA4_SIF is not set. Local testing can just use pytest directly, which is a more faithful local environment anyway since venv mode didn't replicate the --writable-tmpfs overlay or the per-step FA4 editable install.

tools/ci/run_fa4_ci.py
Johnsonms: Not easily right now — the benchmark script only outputs a human-readable ASCII table. I'd like to add --json output to benchmark_attn.py and wire it up to a GitHub Actions job summary, but that's a meaningful standalone addition. Will track it as a follow-up PR to keep this one focused.

tools/ci/run_fa4_ci.py
drisspg: Yeah, running all tests is ideal. We should reassess after this lands the actual time to complete a run

.github/workflows/ci.yml
```diff
@@ -0,0 +1,44 @@
...
+  push:
+    branches: [main]
+
+permissions:
+  contents: read
+
+env:
+  FA4_IMAGE: togethercomputer/training-performance:flash-attn-cu13.0-26.04.01@sha256:56e50b056eb4d671410846c3483e843ee7bd0f5b13cb45b6f0d7eb8bd27694a5
+  CI_WORK_DIR: ${{ vars.CI_WORK_DIR || format('/scratch/user/{0}', github.actor) }}
+  FA4_TEST_FILTER: "1-1-128-True-0-0.0-False-False-False-mha-dtype0"
```
drisspg, line 13: we will expand this later on right?

AI/CI_SETUP.md
```diff
@@ -0,0 +1,242 @@
+# FA4 CI Setup Guide
+
```
drisspg: nit I think this has deviated a little bit from source of ruth;

Also do we think its helpful to have, amybe just trim it down and add to the CI readme.md and then if/when we want to tweak setup, more tests, changing images etc it will be right there

flash_attn/cute/cache_utils.py
drisspg: can oyu rebase I just changed this

tools/ci/build_sif.sh
```diff
@@ -0,0 +1,45 @@
...
+
+echo "=== FA4 SIF Build ==="
+echo "  def file : $DEF_FILE"
+echo "  output   : $OUTPUT"
+echo "  tmp dir  : $TMP_DIR"
+echo "  cache dir: $CACHE_DIR"
+echo
+
+sudo \
+  APPTAINER_TMPDIR="$TMP_DIR" \
```
drisspg, line 36: I dont know too much about apptainer, but 1 other thing is that since we are using hosted runner I imagine that the cache is persistent across runs, right now we hash all sources so it will be invalidated every time we probably wnat to make sure we dont just grow forever

drisspg:
Some last minute comments and then we can land

.github/workflows/ci.yml
Johnsonms: Yes—this is just a smoke test for the initial CI bring-up: a single case(for faster test) to verify that the kernel compiles and runs correctly on the self-hosted B200 runner. 
This is just needed you or Tri to give me a guidance for tests scope.

tools/ci/build_sif.sh
Johnsonms: Good catch! Fixed in the Pull step — after the current SIF is confirmed present, we now delete any other *.sif files in $CI_WORK_DIR that don't match the current image tag, so stale images from old digests are cleaned up automatically.

AI/CI_SETUP.md
Johnsonms: Yes, That's my notes for the setup.
Moved to tools/ci/README.md and trimmed it down to just the maintainer-relevant bits (credentials, updating the image, expanding tests, FA2 isolation). Dropped all the one-time runner setup steps.

flash_attn/cute/cache_utils.py
Johnsonms: Done. Thanks @drisspg

Johnsonms:
Thanks @drisspg  — I’ve updated the code and addressed all the comments. please take another look

PR #2404, opened 2026-03-28
fix(cute): SM120 forward/bwd and atomic add compatibility

### LINKED ISSUE #2386: [BUG] CuTe on sm_120: forward TMA epilogue mis-gating, backward non-sm90 config bug, and NVVM atomic API incompatibility
### Summary

While bringing the CuTe path up on `sm_120`, I ran into 3 separate issues in the CuTe forward/backward stack.

These were all hit in a minimal smoke test that only does:

1. import `flash_attn.cute`
2. run a small causal forward
3. run a simple backward from a scalar loss

After fixing these issues locally in a vendored copy, the same forward+backward smoke test passes on RTX 5090.

### Environment

- GPU: NVIDIA GeForce RTX 5090
- Compute capability: `sm_120`
- torch: `2.10.0a0+b4e4ee81d3.nv25.12`
- nvidia-cutlass-dsl: `4.4.2`
- quack-kernels: `0.3.5`

### Minimal repro

```python
import torch

device = "cuda"
dtype = torch.float16

q = torch.randn(1, 128, 2, 64, device=device, dtype=dtype, requires_grad=True)
k = torch.randn(1, 128, 2, 64, device=device, dtype=dtype, requires_grad=True)
v = torch.randn(1, 128, 2, 64, device=device, dtype=dtype, requires_grad=True)

out, lse = flash_attn_func(q, k, v, causal=True)
loss = out.float().square().mean() + 1e-3 * lse.float().square().mean()
loss.backward()
```

---

### Error

```text
```

### Traceback location

The failure comes from the forward epilogue calling:

- `copy_utils.tma_get_copy_fn(...)`
- which eventually reaches `cpasync.tma_partition(...)`
- with `tma_atom_O=None`

### Root cause

In the shared forward base path, `epilogue()` is called with `tma_atom_O=None`, but `self.use_tma_O` can still be enabled by architecture gating.

So this path can enter the TMA output-store branch even though it has no valid TMA output atom.

---

### Error

```text
UnboundLocalError: cannot access local variable 'dQ_single_wg' where it is not associated with a value
```

### Root cause

In `_flash_attn_bwd`, `dQ_single_wg` is assigned in the `arch // 10 == 9` path, but the compile-key construction for `arch // 10 in [8, 9, 12]` still references it.

So on `sm_120`, backward compile-key generation fails before kernel compilation.

---

### Error

```text
TypeError: atomicrmw() got an unexpected keyword argument 'res'
```

### Root cause

The helper used by backward atomic accumulation directly calls:

```python
nvvm.atomicrmw(res=..., op=..., ptr=..., a=...)
```

That is not compatible with the NVVM binding exposed in this environment (`nvidia-cutlass-dsl 4.4.2` on this stack).

---

## Result after local fixes

After applying the following local fixes:

1. restrict / disable the invalid TMA output-store branch in the shared forward path
2. initialize `dQ_single_wg` for non-sm90 backward paths
3. replace the direct `nvvm.atomicrmw(...)` float atomic-add call with `cute.arch.atomic_add(...)`

the same forward + backward smoke test succeeds on `sm_120`.

NJX-njx:
## Summary

Addresses #2386 (sm_120 smoke test: forward + backward).

1. **Forward (Sm80 path on Blackwell):** FlashAttentionForwardSm80 set use_tma_O from runtime arch (\\>= sm_90\\), but the Sm80/CpAsync epilogue does not provide a valid TMA O atom. On RTX 5090 the runtime arch is sm_120, so the TMA-O branch was taken with 	ma_atom_O=None. The Sm80 forward path now always disables TMA for the output store.

2. **Backward (SM120):** Initialize dQ_single_wg and 
um_stages_PdS in the SM120 tile branch so the shared compile_key tuple is well-defined.

3. **Atomic add:** Replace 
vvm.atomicrmw(..., res=...) with inline PTX 
ed.global.add.f32 for compatibility with newer NVVM Python bindings.

## Testing

- Not run on sm_120 hardware in this environment; please verify with the minimal repro in #2386 if possible.

Fixes #2386

### DIFF
--- a/flash_attn/cute/flash_fwd.py
+++ b/flash_attn/cute/flash_fwd.py
@@ -17,7 +17,6 @@ import cutlass.cute as cute
 from cutlass import Constexpr, Float32, Int32, const_expr, Boolean
 from cutlass.cute.nvgpu import cpasync, warp
 import cutlass.utils as utils_basic
-from cutlass.base_dsl.arch import Arch
 from cutlass.cutlass_dsl import BaseDSL
 
 from quack import copy_utils
@@ -648,8 +647,10 @@ class FlashAttentionForwardSm80(FlashAttentionForwardBase):
         self.num_producer_threads = self.num_threads
         self.num_Q_load_threads = self.num_threads
         self.num_epilogue_threads = self.num_threads
-        # self.use_tma_O = self.arch >= 90 and mCuSeqlensQ is None
-        self.use_tma_O = self.arch >= Arch.sm_90
+        # Sm80 path always uses the cpasync/universal epilogue for O. `self.arch` is the runtime
+        # GPU (e.g. sm_120 on RTX 5090), not the kernel family; comparing to sm_90 incorrectly
+        # enables TMA-O and hits tma_atom_O=None (Dao-AILab/flash-attention#2386).
+        self.use_tma_O = False
         self._setup_attributes()
         SharedStorage = self._get_shared_storage_cls()
         mQ, mK, mV, mO = [assume_tensor_aligned(t) for t in (mQ, mK, mV, mO)]
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -1027,6 +1027,10 @@ def _flash_attn_bwd(
         cluster_size = 1
         use_2cta_instrs = False
         num_threads = 128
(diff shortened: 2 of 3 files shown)


NJX-njx:
Rebased onto main and removed unrelated files (Chinese documentation and hopper/flash.h comments) that were accidentally included. The diff now contains only the SM120 fix changes.

PR #2411, opened 2026-03-30
Various bug fixes / enable subtile > 2

drisspg:
Various bug fixes / enable subtile > 2

1. For small headdim we were expanding up to block_m 192. For the default block-q of 128. Update to take the block_sparse q size for both fwd and bwd. 
2. added `get_sparse_q_block_size` this ideally just returns None if no sparsity or the set size from input. I am like 99% all things have been swapped to using the explicit setting of blocksparse data. We have switch over already in PT but want to give it 1 more release so we still infer just in case. 
3. When enabling a viable 192, 128  -> noticed a kernel hang in the bwd. This was the first time we were hitting `dQ_single_wg` so I fixed to use the correct wg count
4. Also enabled subtiling > 2 which is the case for 192 block Q and m_tile 64 in thw bwd

### DIFF
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -64,6 +64,7 @@ from flash_attn.cute.flash_fwd_combine import FlashAttentionForwardCombine
 
 from flash_attn.cute.block_sparsity import (
     BlockSparseTensorsTorch,
+    get_sparse_q_block_size,
     to_cute_block_sparse_tensors,
     normalize_block_sparse_config,
     normalize_block_sparse_config_bwd,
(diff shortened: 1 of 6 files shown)


flash_attn/cute/interface.py
```diff
@@ -1094,16 +1113,7 @@ def _flash_attn_bwd(
     num_head_kv = k.shape[-2]
 
     use_block_sparsity = block_sparse_tensors is not None
-
-    # SM90 block-sparse backward: tile_m=64 is the GCD between a m_block_size that fits,
-    # the base block_m of 128 from forward, and block-sparse size for subtiling.
-    if arch // 10 == 9 and use_block_sparsity:
-        m_block_size = 64
-        # dQ_swapAB tuning: use False when m_block_size=64 (same as causal case)
-        dQ_swapAB = False
```
v0i0, line 1103: read like we're loosing `dQ_swapAB = False`?

drisspg: yup intentional, now being set in; BwdConfig

also added a test that verifies correctness

PR #2417, opened 2026-03-31
Allow compact block sparse index tensors

jduprat:
Relax validation in block_sparsity.py to allow idx.shape[3] <= expected_n_blocks instead of requiring exact equality.

FA4 only accesses indices 0..cnt-1 per query tile, so the index tensor's last dimension does not need to be as large as ceil(seqlen_k / block_size_n). This enables memory-efficient compact index tensors that avoid O(N^2) memory at long sequence lengths (e.g., sparse attention).

Changes:
- _check_and_expand_block: accept compact n-block dimension and expand only the batch/head/m-block dimensions
- infer_block_sparse_expected_shapes: change strict equality check to upper-bound check (error only when n-blocks exceeds expected, not when smaller)
See: [Note] Allow Compact block sparse indices

Backward compatible: existing code that passes full-sized tensors is unaffected.
This is a followup on PR #2085

### DIFF
--- a/tests/cute/test_mask_mod.py
+++ b/tests/cute/test_mask_mod.py
@@ -1712,5 +1712,91 @@ def test_persistent_blocksparse_empty_tiles():
 
 
 
+def test_compact_block_sparse_indices():
+    """Test that compact block sparse index tensors (idx.shape[3] < n_blocks) work correctly.
+
+    FA4 only accesses indices 0..cnt-1 per query tile, so the index tensor's last
+    dimension does not need to be as large as ceil(seqlen_k / block_size_n). This
+    test verifies that truncated (compact) index tensors produce identical output
+    to full-sized ones.
+    """
+    torch.manual_seed(42)
+    batch_size = 1
+    nheads = 4
+    seqlen_q = 1024
+    seqlen_k = 1024
+    headdim = 128
+    tile_m = 128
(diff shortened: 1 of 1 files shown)


flash_attn/cute/block_sparsity.py
```diff
@@ -98,9 +98,18 @@ def _check_and_expand_block(
...
         cnt, expected_count_shape, f"{name}_block_cnt", context, hint
     )
-    expanded_idx = _expand_sparsity_tensor(
-        idx, expected_index_shape, f"{name}_block_idx", context, hint
-    )
+    # Allow the last dimension (n_blocks) of idx to be <= expected, since
+    # FA4 only accesses indices 0..cnt-1 per query tile. This enables compact
+    # index tensors that avoid O(N^2) memory at long sequence lengths.
+    if idx.ndim == 4 and idx.shape[3] <= expected_index_shape[3]:
+        compact_shape = (*expected_index_shape[:3], idx.shape[3])
```
drisspg: nit: 
can we do 

if:
shape..
else:
shape..
idx = _expand_sparsity_tensor(..)
return 

Also a pattern I like


is you do:
[Note] Allowable kv count shapes: <note>

And then alter on

See: [Note] Allowable kv count shapes

PR #2236, opened 2026-02-05
[Cute,Flex,Fwd] Allow vectorized score_mod definitions

reubenconducts:
[Reupload of #2215 with Sm90] This PR is a `score_mod` "power user" update that allows the user to specify vectorization for a given `score_mod`. It does so in two ways:
- One can set `score_mod.__vec_size__` and have the kernel read that, instead of using the current logic (vec_size = 2 if no `aux_tensors` are present, otherwise 1)
- One can set `buf.__assumed_align__` and `buf.__leading_dim__` for any `aux_tensors`, allowing vectorized loads in the `score_mod` when set. 
These options are not exposed in the API; they must be set specific to the given `score_mod` and `aux_tensors`, and are thus a "power user" feature. 

For a kv bias load `score_mod`, we see up to 2.9x speedup:
```
bias.__assumed_align__ == None
### headdim = 128, causal = True, seqlen_q = 8192, seqlen = 8192, batch_size = 1, nheads = 16, varlen = True ###
FA Python fwd with vec_size 1: 0.905ms, 303.7 TFLOPS
FA Python fwd with vec_size 2: 0.789ms, 348.3 TFLOPS
FA Python fwd with vec_size 4: 0.515ms, 533.9 TFLOPS
FA Python fwd with vec_size 8: 0.422ms, 651.0 TFLOPS
FA Python fwd with vec_size 16: 0.414ms, 663.4 TFLOPS
FA Python fwd with vec_size 32: 0.403ms, 682.3 TFLOPS
FA Python fwd with vec_size 64: 0.415ms, 661.8 TFLOPS
FA Python fwd with vec_size 128: 0.387ms, 709.6 TFLOPS

bias.__assumed_align__ == 4
### headdim = 128, causal = True, seqlen_q = 8192, seqlen = 8192, batch_size = 1, nheads = 16, varlen = True ###
FA Python fwd with vec_size 1: 0.903ms, 304.5 TFLOPS
FA Python fwd with vec_size 2: 0.752ms, 365.6 TFLOPS
FA Python fwd with vec_size 4: 0.458ms, 599.5 TFLOPS
FA Python fwd with vec_size 8: 0.366ms, 751.7 TFLOPS
FA Python fwd with vec_size 16: 0.338ms, 814.4 TFLOPS
FA Python fwd with vec_size 32: 0.330ms, 832.4 TFLOPS
FA Python fwd with vec_size 64: 0.321ms, 855.4 TFLOPS
[...]

PR #2418, opened 2026-03-31
Add compress_factor for compressed causal attention

jduprat:
Add a `compress_factor: int = 1` parameter to `_flash_attn_fwd` and
 `_flash_attn_bwd` that adjusts the causal mask relationship from
 `kv_idx <= q_idx` to `kv_idx <= q_idx // compress_factor`.

 This enables native causal masking for compressed KV sequences where
 each KV token represents `compress_factor` original tokens (e.g., in
 sparse attention). Without this, users must supply a custom `mask_mod`
 function, which is slower because it prevents blocks from being
 classified as "full" — all blocks become "mask" blocks requiring
 per-element evaluation.
    
seqlen_q | cf | compress_factor | mask_mod | speedup
-- | -- | -- | -- | --
2048 | 2 | 0.070 ms | 0.094 ms | 1.35x
4096 | 4 | 0.091 ms | 0.118 ms | 1.29x
8192 | 2 | 0.298 ms | 0.370 ms | 1.24x
16384 | 4 | 0.510 ms | 0.616 ms | 1.21x

### DIFF
--- a/flash_attn/cute/block_info.py
+++ b/flash_attn/cute/block_info.py
@@ -19,6 +19,7 @@ class BlockInfo:
     window_size_left: Optional[Int32] = None
     window_size_right: Optional[Int32] = None
     qhead_per_kvhead_packgqa: cutlass.Constexpr[int] = 1
+    compress_factor: cutlass.Constexpr[int] = 1
 
     @cute.jit
     def get_n_block_min_max(
@@ -33,7 +34,12 @@ class BlockInfo:
             m_idx_max = (m_block + 1) * self.tile_m
             if const_expr(self.qhead_per_kvhead_packgqa > 1):
                 m_idx_max = cute.ceil_div(m_idx_max, self.qhead_per_kvhead_packgqa)
-            n_idx = m_idx_max + seqlen_info.seqlen_k - seqlen_info.seqlen_q
+            if const_expr(self.compress_factor > 1):
+                # Compressed causal: kv_idx * compress_factor <= q_idx
+                # Max KV index for Q tile ending at m_idx_max is m_idx_max // compress_factor
+                n_idx = m_idx_max // self.compress_factor + 1
+            else:
+                n_idx = m_idx_max + seqlen_info.seqlen_k - seqlen_info.seqlen_q
             n_idx_right = n_idx if const_expr(self.is_causal) else n_idx + self.window_size_right
             n_block_max = min(n_block_max, cute.ceil_div(n_idx_right, self.tile_n))
         n_block_min = 0
@@ -60,7 +66,11 @@ class BlockInfo:
         m_block_min = 0
         if const_expr(self.is_causal or (self.is_local and self.window_size_right is not None)):
             n_idx_min = n_block * self.tile_n
-            m_idx = n_idx_min + seqlen_info.seqlen_q - seqlen_info.seqlen_k
+            if const_expr(self.compress_factor > 1):
+                # Compressed causal: q >= kv * compress_factor
+                m_idx = n_idx_min * self.compress_factor
+            else:
+                m_idx = n_idx_min + seqlen_info.seqlen_q - seqlen_info.seqlen_k
             m_idx_right = m_idx if const_expr(self.is_causal) else m_idx - self.window_size_right
             m_block_min = max(m_block_min, m_idx_right // self.tile_m)
         if const_expr(self.is_local and self.window_size_left is not None):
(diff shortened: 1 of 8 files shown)


tridao:
Are there models / algorithms using this? 
Adding this to the interface means we must continue to support this in the future to avoid breaking BC

jduprat:
Yes, I have an implementation of Sparse Attention (NSA) built of top of FA4 that benefits from this.  I do realize the cost of carrying this forward.

drisspg:
So this is an example of a static pattern that :might: benefit from vectorization, I have seem some interesting perf patterns with score_mod https://github.com/pytorch/pytorch/pull/176055

let me explore: https://github.com/Dao-AILab/flash-attention/pull/2236 for masking as well

jduprat:
How can we best evaluate these options?

jduprat:
There are currently two different ways to express compressed-causal sparsity pattern (kv_idx * 64 <= q_idx).  This PR makes the case that we need a 3rd.

KV is the compressed sequence (one entry per 64 original positions). A query at position q attends to compressed-KV index kv iff kv * 64 <= q.  Constants used below: Q_TILE=256, KV_BLOCK=128, cf=64. Each query-tile is 256 consecutive q-positions; each kv-block is 128 consecutive compressed kv-positions (8192 original positions). So one kv-block ≈ 32 q-tiles wide along the diagonal.

1. mask_mod
User passes a cute.jit callable: lambda b, h, q, kv, ...: kv * 64 <= q.
Kernel visits every kv-block for every q-tile (it doesn't know the structure → no tile skipping).
On every visited element, evaluates the predicate (per-element mask op).
Disables pack_gqa (the GQA-packing fast path can't reason about an opaque mask).
Disables the causal R2P (register-to-predicate) fast path — that path is hardcoded for kv <= q, not arbitrary callables.
Cost: O(N²/cf) compute and per-element predicate work on every visited element.

2. AOT block-sparse
User precomputes (block-granular) which kv-blocks each q-tile needs, and labels each kv-block as either full (predicate is true everywhere in the block) or partial (predicate is sometimes true → kernel still needs mask_mod on it).
Kernel Skips kv-blocks not in either list → tile-skipping.
For full blocks: no mask op (free).
For partial blocks: calls mask_mod per element. Still disables pack_gqa (because mask_mod is in the loop for at least one block per tile).

3. compress_factor (PR#2418)
User passes one int: compress_factor=64. Kernel modifies the native causal predicate from kv <= q to kv * cf <= q.
This is "causal with a stride" — it slots into the existing causal R2P fast path (boundary handled analytically in registers, no per-element mask call).
Block classification (full / boundary / skip) is computed inside the kernel from (q_tile_id, kv_block_id, cf) — no input tensors.
Per q-tile: identifies the diagonal kv-block analytically; full blocks are handled with no mask predicate at all; the diagonal is handled by R2P (a register-level mask, not a per-element callable invocation).
pack_gqa works (no mask_mod in the loop).
Backward goes through the same causal Q-direction-only path with no index tensors.

Cost on the user side: literally one int. No tensors to allocate, no transposed bwd layout, nothing to rebuild on shape changes.

|                                  | mask_mod          | AOT block-sparse        | compress_factor   |
|----------------------------------|-------------------|-------------------------|-------------------|
| User input                       | callable          | 4–8 tensors (fwd + bwd) | 1 int             |
| Tile skipping                    | no                | yes (data-driven)       | yes (analytical)  |
| Full-block fast path             | no                | yes                     | yes               |
| Diagonal/boundary block          | per-elem mask_mod | per-elem mask_mod       | causal R2P (free) |
| `pack_gqa`                       | disabled          | disabled                | enabled           |
| Index-tensor memory              | 0                 | O(B·H·M·N)              | 0                 |
| Bwd needs separate user input    | n/a               | yes (Q-direction)       | no                |
| Fwd vs `mask_mod` (N=1M)         | 1×                | 3.21×                   | 3.50×             |
| Fwd vs AOT bs (N=1M)             | —                 | 1×                      | **1.09×**         |
| Bwd vs `mask_mod` (N=131K)       | 1×                | **0.94× (slower)**      | 1.04×             |
| Bwd vs AOT bs (N=131K)           | —                 | 1×                      | **1.11×**         |

image

Between the removal of the AOT fwd/bwd tensors, the preservation of pack_gqa, and the speedups, this adds real value.  Besides AOT is slower in the bwd pass than mask_mod.

The benchmark used to produce this is in https://gist.github.com/jduprat/610cc5992559573a7f7557600f800c45

drisspg:
now that we updated the original benchmark to correctly use blocksparse tensors not just mask_mod, the delta in perf IMO does not warrant the addition of this extra arguement,

I do want to find a nice mechanism for using  user defined r2p + the blocksparse

jduprat:
The graph above was very low dpi, which made specifics hard to see.  It was also log-scale which didn't help.  Reattached so the perf gains are more clear —
bench_compress_factor
bench_compress_factor_linear

To summarize the performance gains —
– At least 9% faster than AOT in the forward pass
– At least 9% faster than mask_mod in the backward pass (AOT slower than mask_mod in backward pass)
– Enables pack_gqa which brings another level of perf itself
– Avoids the extra AOT tensors

Any suggestions for user defined r2p/blocksparse mechanisms to explore?

PR #2422, opened 2026-04-01
[CuTe,Fwd,SM90] Enable head dim 512 for SM90

IwakuraRein:
Relax the checks for head dim for SM90; Fix the N size of MMA and the tensor shape of `PagedKVManager`.

### DIFF
--- a/flash_attn/cute/flash_fwd_sm90.py
+++ b/flash_attn/cute/flash_fwd_sm90.py
@@ -93,13 +93,14 @@ class FlashAttentionForwardSm90(FlashAttentionForwardBase):
         return sQ_layout_atom, sK_layout_atom, sV_layout_atom, sO_layout_atom, sP_layout_atom
 
     def _get_tiled_mma(self):
+        atom_layout_n = 2 if self.tile_hdim > 256 or self.tile_hdimv > 256 else 1
         tiled_mma_qk = sm90_utils_basic.make_trivial_tiled_mma(
             self.dtype,
             self.dtype,
             warpgroup.OperandMajorMode.K,
             warpgroup.OperandMajorMode.K,
             Float32,
-            atom_layout_mnk=(self.tile_m // 64, 1, 1),
+            atom_layout_mnk=(self.tile_m // 64, atom_layout_n, 1),
             tiler_mn=(64, self.tile_n),
         )
         tiled_mma_pv = sm90_utils_basic.make_trivial_tiled_mma(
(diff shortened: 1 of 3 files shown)


tridao:
For hdim 512 we probably want 2 warp groups where WG0 computes the Q @ K, softmax, then write to smem. Then both warp groups compute P @ V. That's how the FA3 implementation did it.
https://github.com/Dao-AILab/flash-attention/blob/main/hopper/mainloop_fwd_sm90_tma_gmma_ws.hpp

As written this PR would have 2 WGs both computing the same Q @ K and softmax so there's redundancy here?

IwakuraRein:
@tridao Thanks for the suggestion. Currently it's still using 1 WG, and the performance in decode is bad. Will optimize based on the `LargeHeadDimV` path in the FA3.

IwakuraRein:
Hi @tridao, thanks again for your reply.

Could you clarify why using 1 warp group for QK is preferred? Is it mainly to reduce register pressure? From my understanding, using 1 warp group for QK and 2 for KV introduces an extra shared memory round trip.  It also requires synchronization after QK, which seems to be conflict with ping-pong scheduling. In my tests, using 2 warp groups for both QK and KV didn’t cause register spilling.

I'd really appreciate any insight you can share. Thanks!

tridao:
How do you split the work when using 2 WGs to compute QK?
this is a 64x 512 @ 64 x 512 MMA right? 
Are you spliting the work along the N (64) dimension? Or do both WGs compute the same 64 x 64 output?

IwakuraRein:
> How do you split the work when using 2 WGs to compute QK? this is a 64x 512 @ 64 x 512 MMA right? Are you spliting the work along the N (64) dimension? Or do both WGs compute the same 64 x 64 output?

@tridao I am splitting along the N dimension. I think QK is (64, 128, 16), and VP is (64, 512, 16). Therefore, the MNK of the MMA for each warpgroup is (64, 64, 16) for QK, and (64, 256, 16) for KV. The generated SASS confirmed this.

tridao:
What's the M, N, K dimension for the QK MMA? 
Shoudn't K be 512 if Q, K have headdim 512?

joanvelja:
Hi @IwakuraRein, @tridao I’ve implemented and tested the forward ownership pattern discussed here on GH200: one consumer warpgroup computes QK and softmax, publishes P to shared memory, and both consumer warpgroups compute PV.

the implementation is in my fork (https://github.com/joanvelja/flash-attention/blob/edd1e3e191a336c575a2e96d959fa57717bf85e4/flash_attn/cute/flash_fwd_sm90.py). It currently sits within a larger D512 forward/backward branch. My tests cover BF16/FP16, causal varlen attention at scale 1, and noncompact head slices used by ring attention.

Before opening an overlapping forward PR, I wanted to coordinate: are you still working on this, and would contributing here or opening a separate PR be more useful? I can extract the forward implementation, tests, and a reproducible benchmark from my branch.

PR #2423, opened 2026-04-02
Fix some bugs of CI

Johnsonms:
1. Fix double blank line in cache_utils.py (ruff format)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,7 +2,7 @@ name: CI
 
 on:
   push:
-    branches: [main, johnson/ci-fix]
+    branches: [main, ci-fix]
 
 permissions:
   contents: read


.github/workflows/ci.yml
```diff
@@ -2,13 +2,12 @@ name: CI
 
 on:
   push:
-    branches: [main]
+    branches: [main, johnson/ci-fix]
```
drisspg: will this be the perma ci fix branch? lol

Johnsonms: Just for testing purposes. Otherwise, I would need to test with a different CI runner or wait to discover CI issues after the merge. This way, we can trigger the CI runner and fix any issues before merging and reviewing the review. might be `ci-fix` looks better

PR #2428, opened 2026-04-02
CI: extend FA4 test matrix with causal/non-causal correctness and fwd+bwd benchmark seqlen 1K-32K

Johnsonms:
### Summary

The existing CI only runs a single degenerate smoke test (`seqlen=1`), which does not provide meaningful signal for real workloads.

This PR extends the FA4 CI to include:
- correctness checks (causal and non-causal)
- forward and backward benchmarks across sequence lengths 1K–32K

These updates provide the minimal meaningful coverage needed to catch regressions on B200 / Blackwell (SM100), the primary target hardware for FA4.

---

### Changes

- Added correctness tests for both causal and non-causal modes at `seqlen=1024`, with reference comparison
- Added benchmark sweeps (fwd + bwd) across `seqlen=1K–32K` for both causal and non-causal
- Moved `test_ci_local.sh` to `.github/scripts/` and fixed related paths
- Renamed CI job from `test` to `fa4-correctness-and-benchmark`

### DIFF
--- a/.github/scripts/test_ci_local.sh
+++ b/.github/scripts/test_ci_local.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/env bash
 set -euo pipefail
 
-SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")" && pwd)
+SCRIPT_DIR=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")/../../" && pwd)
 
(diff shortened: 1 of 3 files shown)


Johnsonms:
> Seems fine, OOC for tests let say we wanted to run a much more meaningful set, what is the runtime budget we have?

We don’t have much budget at the moment, so this would be a temporary short-term deployment on this node. However, we could run a weekly large-scale test, and I can handle that manually if you think that would be a good idea.

PR #2431, opened 2026-04-03
Draft: exploratory SM100 support for larger attention head dims

m0at:
## Summary
This draft captures exploratory SM100 work toward larger attention head dimensions in FA3, including an initial `V/O` chunking approach and the follow-up B200 findings that showed why that approach is not sufficient for dense `512`-dim attention.

## What This Draft Contains
- relax SM100/SM110 validation for an experimental native larger-head path
- clamp `q_stage` when the predicted SM100 TMEM allocation would overflow
- add experimental native `V/O` chunk dispatch in the interface layer
- add tests covering the experimental dispatch and validation logic

## Files
- `flash_attn/cute/interface.py`
- `flash_attn/cute/flash_fwd_sm100.py`
- `tests/cute/test_flash_attn.py`

## What We Learned On B200
- Gemma 4 full-attention is dense `q/k/v = 512`, not just a `V > 256` case
- the initial `V/O` chunking approach is therefore insufficient for this model
- current SM100 forward still breaks on the dense `512` path before a valid native training run
- current SM100 backward still only supports `<= 128` or `(192, 128)`-style shapes, so it also blocks Gemma 4 full-attention
- CUTLASS's own Blackwell `512` example uses the separate MLA path and a true `2Sm` design; the dense FA3 path is not there yet

## Validation
- `python3 -m py_compile flash_attn/cute/interface.py flash_attn/cute/flash_fwd_sm100.py tests/cute/test_flash_attn.py`
- direct B200 probing showed the old experimental path is not enough for dense `512`

## Status
Keep this PR as draft.

It should not be merged as-is for Gemma 4 / dense `512` support. The current code here is best viewed as exploratory prior work that helped identify the actual SM100 blocker.

## Likely Next Steps
- add true dense `512` support in `flash_attn/cute/flash_fwd_sm100.py`
- add matching dense `512` support in `flash_attn/cute/flash_bwd_sm100.py`
- likely mirror the CUTLASS Blackwell `2Sm` / staged-decomposition strategy rather than relying on the old `V/O`-only chunking
- runtime-validate on B200 before marking ready

### DIFF
--- a/tests/cute/test_flash_attn.py
+++ b/tests/cute/test_flash_attn.py
@@ -10,6 +10,7 @@ import pytest
 import torch
 
 from einops import rearrange, repeat
+import flash_attn.cute.interface as cute_interface
 
 try:
     from flash_attn.layers.rotary import apply_rotary_emb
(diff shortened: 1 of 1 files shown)


m0at:
Runtime update from direct B200 probing:

- Gemma 4 full-attention is dense `q/k/v = 512` because `attention_k_eq_v` is enabled in the full-attention layers
- the experimental `V/O` chunking in this draft is not sufficient for that case
- current SM100 forward still fails on dense `512`
- current SM100 backward still hard-limits to `<=128` or `(192,128)`-style shapes
- CUTLASS Blackwell `512` support currently lives in the separate MLA / `2Sm` path, which is the more likely direction for a real fix here

Keeping this PR draft and updating the body so it reflects the current state accurately.

m0at:
Closing this FA3-framed exploratory draft. Superseding it with a new FA4-native draft so SM100 larger-head work is tracked under the right backend and expectations.

PR #2441, opened 2026-04-06
[Cute,Sm100,Fwd] add MLA 64/512 with topk sparsity for MQA 128 heads

jayhshah:
Initial saturating decode perf numbers, deepseek 64/512 shape and topk length = 2048.
```
Vanilla MLA
batch: 512, seqlen_q: 1, seqlen_k: 16384, nheads: 128, -> 1.98 ms, 1180.70 TFLOPS
DSA, no bitmask (topk selected assumed in bounds)
batch: 512, seqlen_q: 1, seqlen_k: 16384, nheads: 128, -> 0.31 ms, 955.47 TFLOPS
DSA, bitmask
batch: 512, seqlen_q: 1, seqlen_k: 16384, nheads: 128, -> 0.33 ms, 898.08 TFLOPS
```

Now added:
1. persistent kernel for perf ✅ 
2. expose through interface ✅
3. handle seqlen_q > seqlen_k when causal (kernel currently hangs) ✅

Tests added in `test_flash_attn_mla_absorbed` and `test_flash_attn_mla_absorbed_varlen`.

tcgen05 mma codegen is not very good for this kernel, will need to be revised for more perf.

Not planned in this PR: page table (though high priority), sliding window.

PR authored by myself and @jack-carlisle-2025


tridao:
I'm mostly concerned w the interface. We're adding `qv` (makes sense), `topk_indices`, `topk_indices_maybe_oob` and `min_seqlenk`. 
Do we need `topk_indices_maybe_oob` and `min_seqlenk`?

tridao:
And we should call it `gather_kv_indices` or sth more generic instead of `topk`.

tridao:
Can you add the benchmark script as well? Will be useful later when we refactor (while making sure perf stays the same). 
Can you add the shapes that are important to benchmark?

jayhshah:
> I'm mostly concerned w the interface. We're adding `qv` (makes sense), `topk_indices`, `topk_indices_maybe_oob` and `min_seqlenk`. Do we need `topk_indices_maybe_oob` and `min_seqlenk`?

For `topk_indices_maybe_oob` I can remove this, theoretically the user could disable causal masking for rectangular shapes with seqlen_k > seqlen_q + topk_length if topk indices are guaranteed to be drawn to lie under the causal diagonal, but this case is somewhat niche.

For `min_seqlen_k`, I would like to retain this so the user can guarantee all sequences in a batch exceed the topk length to disable oob checking and obtain more perf in the varlen case (which is typical), which can yield up to a ~6% gain. We already pass in metadata like `max_seqlen_q` for informing heuristics so I think this is ok?

tridao:
Sure, let's keep `min_seqlen_k`. Other than that LGTM

tridao (approved):

tridao:
we can merge once the benchmark script is added

jayhshah:
On B300 and CUDA 13.1:

compute-bound s_q=s_k MQA 16
```
==========================================================
  FWD (ms / TFLOPS / MFU% / TBs / BW%)
==========================================================
     hdim causal batch seqlen                          FA4
----------------------------------------------------------
   64-512  False     4   8192    5.52/1692/75.2%/0.21/2.7%
   64-512   True     4   8192    2.72/1716/76.2%/0.43/5.4%
```

decode s_q=1 MQA 128 -- we see how MLA shape nicely balances flops and bandwidth for decode
```
   64-512  False   512   8192    0.87/1337/59.4%/5.69/71.1%
```

decode s_q=1 MQA 128 with topk=2048
```
   64-512  False   512   16384   0.29/996/44.3%/4.61/57.6%
```

jayhshah:
I've made the interface changes (removing `topk_indices_maybe_oob` and changing variable names to be more generic) and added qv and gather kv options to the benchmark_attn.py script. Merging in now.

PR #2448, opened 2026-04-08
expose num_splits for FA2 and add option for kernel blocksize alignment

liangel-02:
expose num_splits as a parameter in FA2. if user sets num_splits=1 we can assume they want a numerical match between paged KV vs non-paged. so when num_splits=1, align the split KV kernel (used for paging) block sizes to align with the standard kernel block sizes.

### DIFF
--- a/csrc/flash_attn/flash_api.cpp
+++ b/csrc/flash_attn/flash_api.cpp
@@ -532,7 +532,8 @@ mha_varlen_fwd(at::Tensor &q,  // total_q x num_heads x head_size, total_q := \s
                int window_size_right,
                const float softcap,
                const bool return_softmax,
-               std::optional<at::Generator> gen_) {
+               std::optional<at::Generator> gen_,
+               int num_splits = 0) {
 
     // Otherwise the kernel will be launched from cuda:0 device
     at::cuda::CUDAGuard device_guard{q.device()};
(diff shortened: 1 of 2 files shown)


liangel-02:
cc @drisspg

csrc/flash_attn/flash_api.cpp
```diff
@@ -532,7 +532,8 @@ mha_varlen_fwd(at::Tensor &q,  // total_q x num_heads x head_size, total_q := \s
                int window_size_right,
                const float softcap,
                const bool return_softmax,
-               std::optional<at::Generator> gen_) {
+               std::optional<at::Generator> gen_,
+               int num_splits) {
```
drisspg: are all call sites properly passing in this new arg?

drisspg: maybe optional?

csrc/flash_attn/src/flash_fwd_launch_template.h
```diff
@@ -162,7 +162,15 @@ void run_flash_splitkv_fwd(Flash_fwd_params &params, cudaStream_t stream) {
 
 template<typename T, int Headdim, bool Is_causal>
 void run_mha_fwd_splitkv_dispatch(Flash_fwd_params &params, cudaStream_t stream) {
-    constexpr static int kBlockM = 64;  // Fixed for all head dimensions
+    if (params.num_splits == 1) {
+        // match standard kernel: kBlockM=128, kBlockN=128 for hdim<=64,
+        // kBlockM=128, kBlockN=64 for hdim 96/128.
+        constexpr static int kBlockM = 128;
```
drisspg: do we actually need to change block_m ? did you do any perf measurements?

tridao (approved):

tridao:
just confirming that the default behavior is unchanged. Only when user passes in num_splits does it change from default?

liangel-02:
@tridao yeah, default behaviour is unchanged, also confirmed all FA2 tests pass

csrc/flash_attn/src/flash_fwd_launch_template.h
```diff
@@ -162,11 +162,16 @@ void run_flash_splitkv_fwd(Flash_fwd_params &params, cudaStream_t stream) {
 
 template<typename T, int Headdim, bool Is_causal>
 void run_mha_fwd_splitkv_dispatch(Flash_fwd_params &params, cudaStream_t stream) {
-    constexpr static int kBlockM = 64;  // Fixed for all head dimensions
+    constexpr static int kBlockM = 64;
     // TD [2023-08-28]: nvcc segfaults for headdim 96 with block size 64 x 256,
     // and for headdim 192 with block size 64 x 128.
     constexpr static int kBlockN = Headdim <= 64 ? 256 : (Headdim <= 128 ? 128 : 64);
-    run_flash_splitkv_fwd<Flash_fwd_kernel_traits<Headdim, kBlockM, kBlockN, 4, false, false, T>, Is_causal>(params, stream);
+    if (params.num_splits == 1) {
```
drisspg, line 172: nit leave a small comment about this behavior change

PR #2458, opened 2026-04-14
[FA3] uv installation support

bbuschkaemper:
## Motivation

FA3, even though in beta, is supported in many downstream frameworks, e.g. [megatron-core](https://github.com/NVIDIA/Megatron-LM). To streamline installation using `uv` and `pyproject.toml`, the building of the FA3 beta needs to be adapted: Currently, building using `uv` results in all package files being dropped directly into `site-packages/`, instead of the corresponding `site-packages/flash_attn_3` subfolder.

Existing downstream frameworks like `megatron-core` need a `flash_attn_3` package to work (see [here](https://github.com/NVIDIA/Megatron-LM/blob/ebfa13852f3d6bdb61c8571d68ccab13cb239089/megatron/core/transformer/attention.py#L64)), preventing a clean building solution using a single `pyproject.toml` and requiring custom build steps.

## Solution

To support projects to integrate FA3 correctly, I propose a backwards-compatible change to the package setup, allowing FA3 to be properly built into a package using a `pyproject.toml` while maintaining existing behaviour.

This change outputs the `flash_attn_interface.py` and `flash_attn_config.py` files into both the `site-packages/` (for backwards compatibility) and `site-packages/flash_attn_3` folders.


bbuschkaemper:
@tridao This PR is supposed to be an easy, backwards-compatible change to allow clean fa3 builds with uv so that packages that require a "flash_attn_3" package instead of a single "flash_attn_interface" module will work with it. If you have a different plan for fa3 building let me know; I can adapt the PR as-needed.

zeronewb:
would really nice to have, everytime build from sources kinda PITA

janeyx99:
@zeronewb slightly tangential to uv but if you wanted prebuilt FA3 wheels (they work with any torch version 2.9+ cuz they're ABI stable with torch :D), you can install them from our index: https://x.com/liangel02/status/2030025693380133103

bbuschkaemper:
> @zeronewb slightly tangential to uv but if you wanted prebuilt FA3 wheels (they work with any torch version 2.9+ cuz they're ABI stable with torch :D), you can install them from our index: https://x.com/liangel02/status/2030025693380133103

If you can add these properly to uv as source index then its a good alternative. Also: how are the prebuilt wheels packaging fa3? Do you import the interface module directly or is it packaged as `flash_attn_3` (my PR exports both since downstream NVIDIA repos have both direct- and package imports)?

janeyx99:
cc @liangel-02 ^

liangel-02:
@bbuschkaemper the prebuilt wheels supports both direct import via `import flash_attn_interface` and package import via `import flash_attn_3`

bbuschkaemper:
> @bbuschkaemper the prebuilt wheels supports both direct import via `import flash_attn_interface` and package import via `import flash_attn_3`

Works great! I would still like to include this PR for anyone who wants to build themselves as well as for completeness (i.e. making it `uv`/`pyproject.toml` compatible for any environment or future CUDA versions).

bbuschkaemper:
@Johnsonms if you find the time, could you take a look at this? We plan a protein language model pretraining run and a single reproducible `pyproject.toml` would be nice for clean apptainer images.

Since some downstream packages directly `import flash_attn_interface` and some do `from flash_attn_3 import flash_attn_interface`, supporting both versions cleanly with `uv` would be nice to have.

Johnsonms:
Reviewed and tested on an H100 (SM90). The change is minimal, backwards-compatible, and resolves the issue as intended.

Note for reviewers: the PR description and the Copilot comments predate a force-push and describe the earlier FLASH_ATTENTION_AS_PACKAGE approach; they no longer match the current shim-based commit. Updating the description would avoid confusion.

Verification:
- uv pip install -e . fails on main (could not create 'flash_attn_3/_C.abi3.so') and succeeds on this branch.
- Both import flash_attn_interface and from flash_attn_3 import flash_attn_interface resolve identically.
- fp16 hdim128 forward matches a torch reference (max_abs_err ≤ 2e-3), causal and non-causal.

I pushed a follow-up commit (5d833fd) addressing four items found during testing:
1. Added flash_attn_3/flash_attn_config.py so from flash_attn_3 import flash_attn_config also works (per the description's interface + config intent).
2. Un-ignored that file in .gitignore; the bare flash_attn_config.py pattern was silently dropping it.
3. __version__ now reads from package metadata with a fallback rather than a hardcoded string.
4. Moved dependencies under [project] in the README uv snippet for valid PEP 621.

Johnsonms (approved):

PR #2459, opened 2026-04-15
Handle linter for flash mla file

jayhshah:
I would prefer to exclude `flash_fwd_mla_sm100.py` from ruff format checks to preserve table formatting readability. Do you have a preference @tridao ?


tridao:
for the sections of the code you don't want formatted you can mark them w comment (i forget, `# noqa` or sth like that)

Johnsonms (approved):

Johnsonms (approved):

PR #2462, opened 2026-04-15
[FA4] Add missing fence.proxy.async SM90 backward

lw:
The CUDA memory model requires a fence.proxy.async.shared::cta when shared memory transitions from generic-proxy access back to async-proxy access. In the SM90 backward kernel, sLSE is co-loaded with Q into the same pipeline stage via TMA (async proxy), and then read by threads via generic ld.shared. Before releasing the pipeline stage (consumer_release) — which allows TMA to overwrite it with the next tile — the kernel must fence to signal that the generic proxy is done with the memory.

In the is_dQ_wg=True path, an unrelated fence_view_async_shared() for the dQaccum store incidentally covered sLSE. In the is_dQ_wg=False path (used when dQ_single_wg=True), no such fence existed, leaving a window where TMA could begin overwriting the pipeline stage while generic-proxy reads of sLSE were still formally in flight.

Add fence_view_async_shared() before pipeline_Q.consumer_release in the is_dQ_wg=False path.

### DIFF
--- a/flash_attn/cute/flash_bwd_sm90.py
+++ b/flash_attn/cute/flash_bwd_sm90.py
@@ -1592,6 +1592,9 @@ class FlashAttentionBackwardSm90:
                 mma_dsq_fn(tCrA=tdKrdS, B_idx=smem_idx_Q, zero_init=not dKV_accumulate, wg_wait=1)
             pipeline_dO.consumer_release(consumer_state_dO_cur)
             warpgroup.wait_group(0)
+            # Fence generic-proxy reads of sLSE before releasing the pipeline_Q
+            # stage back to TMA for overwriting.
+            cute.arch.fence_view_async_shared()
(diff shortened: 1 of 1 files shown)


lw:
Stack from [ghstack](https://github.com/ezyang/ghstack/tree/0.14.0) (oldest at bottom):
* __->__ #2462

drisspg (approved):

tridao:
There's already a fence here. Does this not cover LSE?
https://github.com/Dao-AILab/flash-attention/blob/d7f60e6639250626589c528a943aaaba6fe5955a/flash_attn/cute/flash_bwd_sm90.py#L1550

lw:
Yes I think you're right. I saw that there was a fence in the other branch of the `if` statement, so I thought we needed one here too, but I didn't look above. Sorry for the false positive.

PR #2455, opened 2026-04-13
Add CLC scheduler heuristic

drisspg:
# CLC Scheduler Heuristic

Human note: I had Pi summarize all the things we ran. We pair-wrote compile + run scripts around the benchmark helper from Inductor / transformer-nuggets that flushes L2 and returns sample statistics, then bootstrapped a paired 95% confidence interval around `speedup_on_vs_off`.

## Shipped heuristic

Disable CLC (use **STATIC** scheduling) when either:

1. **varlen MHA** — varlen with `q_heads == kv_heads`
2. **dense noncausal** — non-varlen, non-causal, non-local-windowed

Otherwise, keep the CLC path available.

## Decision metric

All comparisons use:

- **`speedup_on_vs_off`**
- paired **95% confidence interval**
- significance flag **`ci95_excludes_1x`**

| Value | Meaning |
|---|---|
| `speedup_on_vs_off > 1.0` | CLC is faster |
| `speedup_on_vs_off < 1.0` | CLC is slower |
| `ci95_excludes_1x = True` | difference is statistically stable |

## Aggregate summary by workload

| Workload | Cases | Significant | Wins | Losses |
|---|---:|---:|---:|---:|
| Dense | 750 | 693 | 324 | 369 |
| Varlen | 7200 | 5353 | 3674 | 1679 |
| Block-sparse | 1872 | 878 | 427 | 451 |

### Aggregate result

| Cases | Significant | Wins | Losses | Mean Δ% |
|---|---:|---:|---:|---:|
| 750 | 693 | 324 | 369 | +5.74 |

Dense is mixed overall, but the shipped heuristic explicitly gates the dense noncausal bucket.

### Dense by causal

| Setting | Cases | Significant | Wins | Losses | Mean Δ% |
|---|---:|---:|---:|---:|---:|
| Causal | 375 | 357 | 257 | 100 | +14.02 |
| Noncausal | 375 | 336 | 67 | 269 | -2.53 |

**Dense causal is a strong net win (+14.02%), while dense noncausal is a net regression (-2.53%).** The mixed aggregate (+5.74%) is the average of a very positive causal story and a clearly negative noncausal one.

### Dense by head mode

| Mode | Cases | Significant | Wins | Losses | Mean Δ% |
|---|---:|---:|---:|---:|---:|
| `mha` | 150 | 149 | 67 | 82 | +3.75 |
| `gqa2` | 150 | 135 | 56 | 79 | +4.32 |
[...]

PR #2463, opened 2026-04-16
fix causal calcs

drisspg:
# Summary

- **Fix lower right causal bug in clc script causing insane tflop numbers; add the clc log parser

Raw data:
https://gist.github.com/drisspg/e0b510ee69c993a66b7e97770b06aca6

clc viz:
image


Edenzzzz:
I got this 
ref: https://github.com/Dao-AILab/flash-attention/pull/2455#issuecomment-4256672505
image

tridao (approved):

PR #2470, opened 2026-04-17
[Cute,Fwd,Sm100] Fix the crash when seqlen_k=0

Johnsonms:
## Summary
                                                                                     
  Fix crash when K/V has `seqlen == 0` (e.g. vLLM CUDA-graph capture with an         
  all-padding batch).
                                                                                     
  Two crash sites, one root cause (`seqlen_k == 0` flows into tensor shapes):        
   
  - **`causal=False`** — `StaticPersistentTileScheduler` passes host setup;          
    kernel launches; the first TMA over `k.shape=(B, 0, H, D)` goes OOB →
    **PTX IllegalInstruction**.                                                      
  - **`causal=True`** — `SingleTileLPTScheduler.Params.create` evaluates             
    `size_l2 // size_one_head`, where                                                
    `size_one_head = seqlen_k * (d + d_v) * elem_size == 0` → **host SIGFPE**        
    before the kernel launches.                                                      
                                                                                     
  ## Fix                                                                             
                                                                                     
  Host-side early return in `_flash_attn_fwd` (`flash_attn/cute/interface.py`):      
  when `seqlen_k == 0`, write zeros to `out` and `-inf` to `lse`, then skip the
  kernel launch entirely. This is math-correct for an empty softmax and              
  mirrors FA2's handling in `csrc/flash_attn/flash_api.cpp:497-504`.                 
                                                                                     
  No kernel changes.                                                                 
                                                                                    

### (on main without fix)

```python
q = torch.randn(4, 1, 16, 128, device="cuda", dtype=torch.bfloat16)
k = torch.empty(4, 0, 16, 128, device="cuda", dtype=torch.bfloat16)
v = torch.empty(4, 0, 16, 128, device="cuda", dtype=torch.bfloat16)
out, lse = flash_attn_func(q, k, v, causal=False)  # cudaErrorIllegalInstruction
out, lse = flash_attn_func(q, k, v, causal=True)   # Fatal: Floating point exception
```

### After fix

```python
out, lse = flash_attn_func(q, k, v, causal=False)  # all-zeros output, -inf LSE
out, lse = flash_attn_func(q, k, v, causal=True)   # all-zeros output, -inf LSE
```

  ## Test plan                                                                       
  - [x] Repro confirmed on `main`: `cudaErrorIllegalInstruction` (causal=False),
  SIGFPE (causal=True)                                                               
  - [x] Fix verified: all-zeros output, `-inf` LSE, no crash
  - [x] `test_flash_attn_seqlen_k_zero` — 16 cases pass on B200                      
  - [x] `tests/cute/test_flash_attn.py -k test_flash_attn_output` — no regression    
  - [x] Pre-commit (ruff check + format) passes  

PR authored by @yunzhongOvO  @Johnsonms @Orion34-lanbo

### DIFF
--- a/tests/cute/test_flash_attn.py
+++ b/tests/cute/test_flash_attn.py
@@ -2410,3 +2410,65 @@ def test_flash_attn_mla_absorbed_varlen(
                 # print(f"out vs out2 max diff: {(out_cmp - out2).abs().max().item()}, {iter=}")
                 # print(f"out vs out2 mean diff: {(out_cmp - out2).abs().mean().item()}, {iter=}")
                 assert torch.equal(out_cmp, out2), f"non-deterministic with max diff = {(out_cmp - out2).abs().max().item()} on {iter=}"
+
+
+# ---------------------------------------------------------------------------
+# Regression test: seqlen_k=0 must not crash (CUDA graph padding scenario)
+# ---------------------------------------------------------------------------
+
+@pytest.mark.parametrize("causal", [False, True])
+@pytest.mark.parametrize("d", [128, 192])
+@pytest.mark.parametrize("seqlen_q", [1, 64, 128, 256])
+@maybe_fake_tensor_mode(USE_FAKE_TENSOR)
+def test_flash_attn_seqlen_k_zero(seqlen_q, d, causal):
+    """K/V with physical seqlen dim == 0 must not crash.
+
+    seqlen_k == 0 violates two downstream invariants, producing two
+    different crashes depending on the mask:
+
+      causal=False -> TMA descriptor over a 0-length K tensor goes OOB
+                      on first tile load -> PTX IllegalInstruction.
+
+      causal=True  -> SingleTileLPTScheduler's L2-swizzle heuristic in
+                      tile_scheduler.py evaluates
+                          size_l2 // (seqlen_k * (d + d_v) * elem_size)
+                      -> host SIGFPE before the kernel launches.
+
+    Varlen paths (cu_seqlens_k / seqused_k with K physical seqlen > 0)
+    are not exercised here: per-batch empty slots are already handled
+    by the kernel's fake-iteration path and do not hit either invariant.
+    """
+    if IS_SM90:
+        pytest.skip("SM90 uses a different kernel path")
+
+    device = "cuda"
+    dtype = torch.bfloat16
+    dv = 128 if d == 192 else d
+    batch_size = 4
+    nheads = 16
+    nheads_kv = 16
+
+    torch.manual_seed(0)
+
+    q = torch.randn(batch_size, seqlen_q, nheads, d, device=device, dtype=dtype)
+    # K/V have physical seqlen dim == 0 — this is what crashes on unpatched FA4.
+    # causal=False hits GPU IllegalInstruction (TMA OOB on 0-length K tensor).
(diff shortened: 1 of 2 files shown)


jayhshah:
Is this specifically about seqlen_k = 0? For varlen k, we were always testing possibly zero length seqlen_k batches (tests pass on main) and relying on one cta to go through one fake work iteration (loading zero for operands, masking output and accumulating zero to O) to write zero to O. For split we don't need to do this since combine kernel handles the write zero with the lse known.

Of course, we could also write zero to O directly.

tests/cute/test_flash_attn.py
```diff
@@ -2409,4 +2409,62 @@ def _gen_unused_masks(padding_mask, add_unused, max_seq_len, bs, device):
...
+
+@pytest.mark.parametrize("causal", [False, True])
+@pytest.mark.parametrize("d", [128, 192])
+@pytest.mark.parametrize("seqlen_q", [1, 64, 128, 256])
+@maybe_fake_tensor_mode(USE_FAKE_TENSOR)
+def test_flash_attn_seqlen_k_zero(seqlen_q, d, causal):
+    """Regression: seqlen_k=0 must not crash in the SM100 forward kernel.
+
+    When vLLM captures CUDA graphs with padded batches, padding entries have
+    seqlen_k=0.  This makes n_block_max=0, so n_block_first = n_block_max - 1
```
jayhshah: Don't we already prevent this via
`n_block_first = n_block_max - 1 if n_block_max > 0 else 0`?

Johnsonms:
> Is this specifically about seqlen_k = 0? For varlen k, we were always testing possibly zero length seqlen_k batches (tests pass on main) and relying on one cta to go through one fake work iteration (loading zero for operands, masking output and accumulating zero to O) to write zero to O. For split we don't need to do this since combine kernel handles the write zero with the lse known.
> 
> Of course, we could also write zero to O directly.

Thanks @jayhshah  — you’re right that this is specifically the **non-varlen `seqlen_k = 0`** case. I verified this on `main`:

* **Varlen** with `cu_seqlens_k[b+1] == cu_seqlens_k[b]` and physical K seqlen > 0 works in both causal and non-causal modes. The per-batch fake iteration correctly writes zeros, so no change is needed there.
* **Non-varlen** K/V with physical seqlen dimension = 0 crashes in two different ways, depending on the masking mode.

I traced the two failures to distinct downstream invariants (details are in the refactored commit message), and the fake-iteration path cannot cover them because:

* **Non-causal:** the TMA descriptor over the zero-length K tensor goes out of bounds on the first load, before any kernel-side guard can run.
* **Causal:** it fails on the host in `SingleTileLPTScheduler.Params.create`, where the L2 swizzle heuristic divides `size_l2 // size_one_head`, and `size_one_head = seqlen_k * (d + d_v) * elem_size = 0`, so this happens before the kernel is even launched.

I agree that writing zeros directly to `O` is the cleaner solution here, and that is what the early return does. The scope of the fix is limited to the non-varlen path.

tests/cute/test_flash_attn.py
Johnsonms: Yes, it is exactly and perfect. Thanks @jayhshah leading me here

jayhshah (approved):

PR #2471, opened 2026-04-17
Add [STATIC] tile trace for CLC-off work distribution

Edenzzzz:
## Summary
- Adds a `[STATIC] query sm=... cta=... (m_blk=...,h=...,b=...,s=...) valid=...` trace line to `SingleTileLPTScheduler.get_current_work()` on the non-CLC path
- Matches the existing `[CLC] query ...` format so `parse_clc_log.py` can produce work-distribution histograms for both CLC-on and CLC-off modes
- Guarded by `FA_LOG_LEVEL=3` and `is_valid` — zero cost when logging is off, no invalid/exhausted tile spam
- Remove invalid/exhausted tile as for non-persistent each tile will have an "invalid" tail.
image

## Usage
```bash
# CLC off trace
FA_LOG_LEVEL=3 FA_CLC=0 python your_script.py > static_trace.log 2>&1

# Convert [STATIC] -> [CLC] for the existing parser
sed 's/\[STATIC\]/[CLC]/g' static_trace.log > converted.log
python AI/parse_clc_log.py converted.log --html -o static_histogram.html
```

## Test plan
- [x] Verified trace output on B200 for `mqa_causal_q16384_k1024_h96` (16:1, d=96)
- [x] Confirmed 2048 valid tile lines (128 Q-blocks × 16 heads), no invalid tile spam
- [x] `parse_clc_log.py --html` generates correct histogram after `[STATIC]` → `[CLC]` conversion

cc @drisspg

### DIFF
--- a/flash_attn/cute/tile_scheduler.py
+++ b/flash_attn/cute/tile_scheduler.py
@@ -21,6 +21,8 @@ from quack.cute_dsl_utils import ParamsBase
 
 import flash_attn.cute.utils as utils
 from flash_attn.cute.fast_math import clz
+from flash_attn.cute.fa_logging import fa_printf
+from flash_attn.cute.utils import smid
 
 
 class SchedulingMode(IntEnum):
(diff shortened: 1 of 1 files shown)


drisspg (approved):
Looks good, thanks

PR #2474, opened 2026-04-19
[Cute,Sm120] fix Spark forward and backward regressions

CuriousCaliBoi:
## Summary
- restore the intended SM80-style control flow for `FlashAttentionForwardSm120` so DGX Spark does not fall into the TMA epilogue path at runtime
- initialize the SM120 backward `dQ_single_wg` state and keep the shared SM80/SM120 backward launcher on concrete softmax-scale values so the Spark backward path compiles cleanly again
- add a regression test that verifies the SM120 forward object keeps its effective arch pinned to `Arch.sm_80`

## Test plan
- [x] Force a fresh Spark compile with cache disabled and run a full FA4 forward+backward smoke test on `NVIDIA GB10` / compute capability `12.1`
- [x] Benchmark forward on Spark with `B=8, S=8192, H=32, D=64, causal=False`: FA4 `46.52 ms / 94.5 TFLOPS`, SDPA `50.18 ms / 87.6 TFLOPS`
- [x] Instantiate `FlashAttentionForwardSm120` directly and verify the effective arch stays `Arch.sm_80`

### DIFF
--- a/flash_attn/cute/flash_bwd.py
+++ b/flash_attn/cute/flash_bwd.py
@@ -433,7 +433,13 @@ class FlashAttentionBackwardSm80:
         tile_sched_params = TileScheduler.to_underlying_arguments(tile_sched_args)
         grid_dim = TileScheduler.get_grid_shape(tile_sched_params)
 
-        softmax_scale_log2, softmax_scale = utils.compute_softmax_scale_log2(softmax_scale, self.score_mod)
+        softmax_scale_for_scoremod = softmax_scale
+        if cutlass.const_expr(self.score_mod is None):
+            softmax_scale_log2 = softmax_scale * math.log2(math.e)
+        else:
+            softmax_scale_log2, softmax_scale_for_scoremod = utils.compute_softmax_scale_log2(
+                softmax_scale, self.score_mod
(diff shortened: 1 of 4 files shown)


hebo1221:
Validated this on DGX Spark / GB10 (SM121).

Env:
- driver 580.126.09, CUDA 13.0.88
- PyTorch 2.11.0+cu130
- nvidia-cutlass-dsl 4.4.2
- commit b417fd266d18977252ddee6fa60c74339a0bab89

Results:
- current main ba59def9 reproduces the TMA-O crash (`tma_atom_O=None`)
- #2484 fixes forward but still fails backward with `dQ_single_wg` unbound
- this branch passes dense fwd+bwd smoke for fp16/bf16, D=64/128, causal/non-causal
- this branch passes varlen fwd+bwd smoke for fp16/bf16, D=64, causal/non-causal
- `python -m pytest -q tests/cute/test_sm120.py`: 1 passed when run against the FA4 editable package path

Max diffs vs PyTorch SDPA stayed in the expected fp16/bf16 range; worst observed was `dv=0.003906` for bf16 D=128 causal.

PR #2475, opened 2026-04-19
[Cute] tolerate both nvvm.atomicrmw Python bindings

CuriousCaliBoi:
## Summary
- make `flash_attn.cute.utils.atomic_add_fp32` tolerate CUTLASS DSL wheels whose generated `nvvm.atomicrmw` binding omits the explicit `res` argument
- add a focused regression test that locks in the signature detection logic
- fix the fresh `pip install \"flash-attn-4[cu13]\"` backward-compile failure reproduced on GB10 where both envs reported `nvidia-cutlass-dsl==4.4.2` but exposed different Python bindings

## Test plan
- [x] `python -m pytest tests/cute/test_utils.py`
- [x] In a fresh `flash-attn-4[cu13]` env on GB10, run a small `flash_attn_func(..., causal=True)` forward+backward smoke after reinstalling the patched package
- [x] Confirm the same small backward smoke still passes in the known-good source-tree Spark env

## Notes
- Raw upstream `main` on GB10 still first hits the separate SM120 forward issue tracked in #2474, so end-to-end Spark runtime validation for this branch is still gated on that earlier fix.

### DIFF
--- a/flash_attn/cute/utils.py
+++ b/flash_attn/cute/utils.py
@@ -89,6 +89,11 @@ def _get_disable_2cta_default() -> bool:
     return _fa_disable_2cta_enabled or _fa_disable_2cta_cuda12
 
 
+def _nvvm_atomicrmw_accepts_res(fn: Callable = nvvm.atomicrmw) -> bool:
+    """Handle CUTLASS DSL bindings with and without an explicit result type."""
+    return "res" in inspect.signature(fn).parameters
+
+
 def _compute_base_hash(func: Callable) -> str:
     """Compute hash from source code or bytecode and closure values."""
     try:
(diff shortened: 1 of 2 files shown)


hebo1221:
Validated this on a DGX Spark / GB10 setup.

Environment:
- GPU: NVIDIA GB10, compute capability 12.1
- CUDA 13.0, PyTorch 2.11.0+cu130
- nvidia-cutlass-dsl 4.4.2
- observed `nvvm.atomicrmw` signature includes `res`, and `_nvvm_atomicrmw_accepts_res()` returns `True`

Checks:
- `python -m pytest tests/cute/test_utils.py`: 12 passed
- Applied this PR on top of #2474, then ran dense fp16/bf16, D=64/128, causal/non-causal forward+backward smoke against PyTorch SDPA on GB10: passed

I tested on top of #2474 because this branch alone is still gated by the separate SM120 forward issue before reaching the backward `atomic_add_fp32` path.

Issue #2424, opened 2026-04-02
FA4: Autograd wrappers don't pass mask_mod/score_mod/block_sparse to backward — silent wrong gradients


asolergi-nv:
`FlashAttnFunc` and `FlashAttnVarlenFunc` in `flash_attn/cute/interface.py` accept `mask_mod`, `score_mod`, `aux_tensors`, and block sparsity tensors in the forward pass, but **do not save them to `ctx`** and **do not pass them to `_flash_attn_bwd`** in the backward pass.

This means calling `.backward()` on the output of `flash_attn_func(..., mask_mod=my_mask)` will produce **silently incorrect gradients** — no error, no warning.

### `FlashAttnFunc` (line ~1578)

The forward method accepts `mask_mod` (line 1593) and block sparse tensors (lines 1594-1598) and correctly passes them to `_flash_attn_fwd` (lines 1623-1624). However:

- `mask_mod` is **not saved** to `ctx` (lines 1627-1634)
- `block_sparse_tensors` is **not saved** to `ctx`
- The backward method (lines 1644-1658) calls `_flash_attn_bwd` without `mask_mod` or `block_sparse_tensors`

### `FlashAttnVarlenFunc` (line ~1662)

Similarly, the forward accepts `score_mod` (line 1684) and `aux_tensors` (line 1685) and passes them to `_flash_attn_fwd`. However:

- `score_mod` is **not saved** to `ctx` (lines 1711-1720)
- `aux_tensors` is **not saved** to `ctx`
- The backward (lines 1731-1753) calls `_flash_attn_bwd` without `score_mod`, `score_mod_bwd`, or `aux_tensors`

## Why tests don't catch this

The existing tests for `mask_mod` backward (`tests/cute/test_mask_mod.py`, `run_cute_mask_bwd` at line ~1091) and `score_mod` backward (`tests/cute/test_score_mod.py`, `run_cute_flash_bwd` at line ~755) call `_flash_attn_fwd` and `_flash_attn_bwd` **directly**, bypassing the `FlashAttnFunc` / `FlashAttnVarlenFunc` autograd wrappers entirely.

No test calls `.backward()` on the output of `flash_attn_func` or `flash_attn_varlen_func` with `mask_mod`, `score_mod`, or block sparse tensors enabled, so the missing wiring is never exercised.

Is this expected?

EricLina:
+1, it seems that block sparse doesn't supports backward yet.

reubenconducts:
[...]

PR #2485, opened 2026-04-23
[CuTe,Flex] Wire up interface for flex autograd support

reubenconducts:
Addresses #2424. Added `use_autograd` as an option for testing flex backward. cc @drisspg

### DIFF
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -1787,7 +1787,10 @@ class FlashAttnFunc(torch.autograd.Function):
         num_splits: int = 1,
         pack_gqa: Optional[bool] = None,
         deterministic: bool = False,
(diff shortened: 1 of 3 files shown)


drisspg:
Looks good, I also had opus writeup the custom op deffs such that we can get graph break freak + a pattern Richard mentioned for handling the callables: https://gist.github.com/drisspg/629ace33785a395940af62f89f7e0f47

drisspg (approved):

wangsiyu:
@reubenconducts Hi，it seems that varlen unit test FAILED because varlen caller does not compat with new added param.

Johnsonms:
Thanks @wangsiyu and @reubenconducts,
Let's add nightly tests later Cc: @tridao

PR #2406, opened 2026-03-28
[CuTe, SM120] Forward kernel with optimized TMA path and full features support

sisgrad:
This PR introduces separate sm120 forward kernel with optimized TMA path and full features support (except block sparsity).

I started developing it about two months ago with primary focus on improving performance relative to sm80 (which also executable on sm120 hardware), then this PR was opened 
 https://github.com/Dao-AILab/flash-attention/pull/2349 and I discovered that its performance is slightly worse than mine and other features were split into separate PR's

Summary about performance relative to sm80 kernel: 
- In most cases TMA version is slightly better by 2-3%, but there are cases of about 5% better and worth both. Detailed performance and correctness tables are below.
- Its performance slightly improves with greater batches and degrades with greater heads count. Performance delta with `pack_gqa` is smaller.
- I think further speed up attempts should utilize quantization in some way, not necessary in using sm120 mma ops for quantized inputs, but maybe transferring big quantized tiles into smem and upcast to half in registers. TMA benefits are limited on sm120 because of its smem capacity.

## Feature matrix comparing with https://github.com/Dao-AILab/flash-attention/pull/2349

| Area | `Sm120Tma` | `Sm120TMAOptimized` |
|---|---|---|
| Core transfer model | TMA for Q/K/V | TMA for K/V always; Q can be TMA, cp.async, or direct gmem->regs decode |
| O global store | SMEM->GMEM copy path only (`use_tma_O = False`) | Supports TMA S2G O-store in eligible cases; otherwise SMEM/copy path |
| Decode q1 direct Q gmem->regs | Not implemented | Implemented via `direct_q_gmem_regs_decode`, it can be faster in some cases |
| pack-GQA handling | Supports pack-GQA mode, but no dedicated packed-Q TMA atom path | Dedicated pack-GQA Q TMA atom path (`make_packgqa_tiled_tma_atom`) + pack-aware load/store helpers. Changes in pack_gqa.py were necessary for kernel correctness when `cute.arch.WARP_SIZE % threads_per_row != 0` (head dims 96, 192) |
| Split-KV | Not supported (`num_splits = 1`, `is_split_kv = False`) | Supported (rank-based split detection, split-aware LSE/O writes) |
| Paged KV (`mPageTable`) | Explicitly unsupported (assert) | Supported with restriction: K/V must be batched dense (no varlen K with page table) |
| Varlen (`cu_seqlens`) | Supported | Supported |
[...]

PR #2499, opened 2026-04-27
GeForce RTX 50 SM120 FA4 support

alecco:
This my ongoing work on comprehensive support for SM120 in FA4. The idea is to have broad coverage like SM100 by building bottom up the scaffolding. I tried to follow FA4's code architecture as close as possible and tried to make the commits as small as possible for easy reviewing.

**It is unfinished**, unoptimized, only aiming at correctness. And it has a few minor problems I plan to push later today (e.g. one leftover helper function). It also contains a couple of commits not meant for the final PR including agent instructions.

But it has plenty of tests.

I tried to not touch the other architectures as much as possible.

I present it here early to open the discussion.

There was really good work already present in https://github.com/Dao-AILab/flash-attention/pull/2406 that I completely missed. Hopefully we can join forces.

### DIFF
--- a/AI/SM120_FA4_PLAN.md
+++ b/AI/SM120_FA4_PLAN.md
@@ -1,0 +1,750 @@
+You are working as a careful upstream-oriented engineering agent in a local checkout of a fork of Dao-AILab/flash-attention.
+
+Primary repo:
+  /home/agent/Soft/flash-attention
+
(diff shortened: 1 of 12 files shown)


hg0428:
This looks interesting. Hopefully we get this.

PR #2501, opened 2026-04-27
Parallel compile/prewarm to speed up testing with pytest-xdist (README.md)

alecco:


alecco:
Duh, this was in AGENTS.md already. Sorry for the noise.

PR #2505, opened 2026-04-27
Fix: pass `stream` to SM100 MLA kernel

MatthewBonanni:
`FlashAttentionMLAForwardSm100.__call__()` does not pass `stream` through to `self.kernel().launch()`, so the operation is carried out on the default CUDA stream, leading to potential races with PyTorch ops. Simple 1-line fix.


jayhshah:
Thanks!

jayhshah (approved):

PR #2507, opened 2026-04-27
[CuTe, Bwd] Fix backward compile key churn due to pickling, max_seqlen is a tensor

michaelxu-msft:
## Problem

When `max_seqlen_q` / `max_seqlen_k` are passed as `torch.Tensor` (which HuggingFace Transformers does via `_prepare_from_posids` in the varlen path), the backward compile key in `_flash_attn_bwd` contains `torch.Tensor` values instead of Python scalars.

Specifically, `seqlen_q` and `seqlen_k` remain tensors, so `seqlen_q_rounded // m_block_size == 1` evaluates to `torch.Tensor(False)` rather than `bool`. Since `pickle.dumps(torch.Tensor)` produces a unique hash per tensor object, every backward call generates a new compile key even though the actual kernel parameters are identical. This means that each call will not reuse any previous JIT compiled kernel and instead recompile each time, producing thousands of identical .o files in the JIT cache (if enabled, or recomputes in memory each time), and also takes ~8-13s, massively slowing down training.

### Root Cause

In `_flash_attn_bwd` (line 1323):
```python
seqlen_q = max_seqlen_q if max_seqlen_q is not None else total_q
```

When `max_seqlen_q` is a `torch.Tensor`, all downstream arithmetic produces tensors that leak into the compile key tuple elements `[29]` and `[30]`.

## Fix

Just need to cast `max_seqlen_q` / `max_seqlen_k` to `int()` before they enter the sequence length computation:

```python
seqlen_q = int(max_seqlen_q) if max_seqlen_q is not None else total_q
seqlen_k = int(max_seqlen_k) if max_seqlen_k is not None else total_k
```

## Verification

Tested on NVIDIA B200 (SM 10.0) with `flash-attn-4==4.0.0b10` and `transformers==5.6.2`:

**Before fix** — 10 steps with varying packed seq lengths via HF `_flash_attention_forward`:
```
step 0: bwd_keys=1  disk=1
step 1: bwd_keys=2  disk=2
...
step 9: bwd_keys=10 disk=10   # new kernel every step...
```

**After fix**:
```
step 0: bwd_keys=1  disk=4  [JIT]
step 1: bwd_keys=1  disk=4  [cached]
...
step 9: bwd_keys=1  disk=4  [cached]   # 1 kernel for all steps
```

### Reproduction

```python
from flash_attn.cute import flash_attn_varlen_func
import torch

for max_sl in [512, 768, 1024]:
    cu = torch.tensor([0, max_sl], dtype=torch.int32, device="cuda")
    max_t = torch.tensor(max_sl, dtype=torch.int32, device="cuda")  # tensor, not int
    q = torch.randn(max_sl, 8, 128, dtype=torch.bfloat16, device="cuda", requires_grad=True)
    k = torch.randn(max_sl, 8, 128, dtype=torch.bfloat16, device="cuda", requires_grad=True)
    v = torch.randn(max_sl, 8, 128, dtype=torch.bfloat16, device="cuda", requires_grad=True)
    out = flash_attn_varlen_func(q, k, v, cu_seqlens_q=cu, cu_seqlens_k=cu,
                                  max_seqlen_q=max_t, max_seqlen_k=max_t, causal=True)
    out[0].sum().backward()
    # Without fix: 3 unique compile keys. With fix: 1.
```

### DIFF
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -1320,7 +1320,7 @@ def _flash_attn_bwd(
     else:
         batch_size = cu_seqlens_q.shape[0] - 1
         total_q = q.shape[0]
-        seqlen_q = max_seqlen_q if max_seqlen_q is not None else total_q
+        seqlen_q = int(max_seqlen_q) if max_seqlen_q is not None else total_q
 
     if cu_seqlens_k is None:
         batch_size, seqlen_k = k.shape[:2]
@@ -1328,7 +1328,7 @@ def _flash_attn_bwd(
     else:
         batch_size = cu_seqlens_k.shape[0] - 1
         total_k = k.shape[0]
-        seqlen_k = max_seqlen_k if max_seqlen_k is not None else total_k
+        seqlen_k = int(max_seqlen_k) if max_seqlen_k is not None else total_k
 
     num_head_kv = k.shape[-2]
 
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -1320,7 +1320,7 @@ def _flash_attn_bwd(
     else:
         batch_size = cu_seqlens_q.shape[0] - 1
         total_q = q.shape[0]
-        seqlen_q = int(max_seqlen_q) if max_seqlen_q is not None else total_q
+        seqlen_q = max_seqlen_q if max_seqlen_q is not None else total_q
 
     if cu_seqlens_k is None:
         batch_size, seqlen_k = k.shape[:2]
@@ -1328,7 +1328,7 @@ def _flash_attn_bwd(
     else:
         batch_size = cu_seqlens_k.shape[0] - 1
         total_k = k.shape[0]
-        seqlen_k = int(max_seqlen_k) if max_seqlen_k is not None else total_k
+        seqlen_k = max_seqlen_k if max_seqlen_k is not None else total_k
 
     num_head_kv = k.shape[-2]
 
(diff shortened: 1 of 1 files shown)


LuisRobaina:
Thanks @michaelxu-msft , I was able to get much better performance applying the fix on this PR.

Training steps/min (orange = with the fix).

Screenshot 2026-05-21 at 7 56 49 PM

tridao:
wouldn't `int` cause a cpu-gpu sync?

yuchenwang3:
+1 — independently reproduced on a production stack (Qwen3.5-35B-A3B MoE, Megatron-LM + TransformerEngine + ms-swift, Blackwell): the tensor-valued `compile_key` entries make the FA4 backward recompile every microbatch (~7–8× slower steps), because TE passes `max_seqlen` as a 0-dim CUDA tensor. Same root cause as #2571 / #2635.

On the CPU–GPU sync concern with `int(max_seqlen_*)`: the two affected key entries —

```python
(seqlen_q_rounded // m_block_size == 1),
(seqlen_k_rounded // n_block_size == 1),
```

are pure host-side **branch predicates** that only select a compiled-kernel variant; they aren't passed to the kernel. So one way to avoid both the churn *and* a per-backward `.item()` sync is to specialize on them only when the seqlen is already a host scalar, and fall back to a fixed value when it's a tensor, e.g.:

```python
single_q_block = (not torch.is_tensor(seqlen_q_rounded)) and (seqlen_q_rounded // m_block_size == 1)
```

That keeps the key stable for tensor callers with no device sync.

One thing I'd defer to you on: the comment on those lines says *"prevent TVM stride poisoning when only one block is present"* — if that single-block branch is a **correctness** requirement (not just an optimization) for genuinely single-block inputs, then defaulting tensor callers to the multi-block path would be wrong, and the coercion would instead need to happen once on the host at the autograd `Function` boundary (where the value is often already a Python int from `cu_seqlens`) rather than per-backward. Either way the per-step sync is avoidable. Happy to put up a PR along whichever direction you prefer.

YunfanZhang42:
Got hit with this problem today. Would appreciate if this PR gets merged. Thanks a lot for your work @michaelxu-msft !

Johnsonms:
@YunfanZhang42 Could you please workaround this with @michaelxu-msft PR? I will take more time to dig into this issue later

YunfanZhang42:
@Johnsonms Yes I have applied the patch to my local flash attention 4 fork. Can confirm it solves the problem for me (HF Transformers 5.4, Qwen3 dense, 8x B200, PyTorch FSDP)

michaelxu-msft:
I've updated my PR to match @yuchenwang3 suggestion, hopefully that resolves any concerns about CPU-GPU sync!

Johnsonms:
Verified on a B300 (sm_103) — LGTM. 👍

Root cause is right: when max_seqlen is a tensor, seqlen_*_rounded // block == 1 becomes a fresh 0-d tensor each call, and the cache dict hashes it by object identity → recompile every backward step. The not torch.is_tensor(...) guard yields a stable Python bool and reads no value off the tensor → no CPU–GPU sync.

Checks:
- Churn A/B (varlen bwd, tensor max_seqlen, 6 calls): baseline compile_cache grows 1→6, with the PR it's flat at 1. Fixed.
- Correctness safe by construction: the flag is only in the compile_key, never a kernel arg — tensor callers already produced correct results (just recompiling), so this can't regress output.
- Scope complete: both bwd branches fixed; it's the only tensor-derived key entry; forward has no analogous predicate.

Non-blocking: a regression test asserting the bwd compile_cache stays size 1 across two tensor-max_seqlen calls would lock it in.

Johnsonms (approved):

PR #2510, opened 2026-04-28
[Cute,Bwd,Sm90] Fix determinism for GQA, port Sm100 approach in

v0i0:


flash_attn/cute/flash_bwd_sm90.py
```diff
@@ -402,6 +402,15 @@ def _qkv_transpose(t):
         if const_expr(self.deterministic):
             assert mdQ_semaphore is not None
             mdQ_semaphore = layout_utils.select(mdQ_semaphore, mode=[2, 3, 1, 0])
+        if const_expr(self.deterministic and self.qhead_per_kvhead > 1):
+            assert mdK_semaphore is not None
```
drisspg, line 406: I havent looked at sm100 code, but 2 semaphores as opposed to mdkv is for overlap?

drisspg:
LGTM, nit test

flash_attn/cute/flash_bwd_sm90.py
v0i0: good question, not sure (i just ported over the sm100 code). we should benchmark and see.

v0i0:
> LGTM, nit test

added tests

flash_attn/cute/flash_bwd_sm90.py
v0i0: benchmarked, two semaphores is faster, presumable smaller critical section

drisspg (approved):

Issue #2426, opened 2026-04-02
FA4: ~20-25us Python dispatch overhead per call — optimization targets


hypnopump:
## Summary

CONSORT profiling of FA4's Python-level overhead reveals a constant ~20-25 us dispatch overhead per `flash_attn_func` call (both forward and backward). This is significant for small/medium problem sizes:

- **24% overhead** at batch=1, seqlen=128, heads=4, hdim=64
- **17% overhead** at batch=1, seqlen=2048, heads=16, hdim=128
- **<4% overhead** at batch=4, seqlen=4096 (large problems)

### Forward (total Python: ~73 us/call)

| Function | us/call | % | Actionable? |
|----------|---------|---|-------------|
| `_flash_attn_fwd` body (validation, tensor setup, compile key construction) | 22.1 | 30.2% | Yes — fast-path for repeated shapes |
| CuTeDSL `wrapper` (TVM FFI) | 9.3 | 12.7% | Upstream (cutlass-dsl) |
| `active_fake_mode()` via `is_fake_mode()` (3x/call) | 7.1 | 9.7% | Yes — cache or check once |
| `torch.empty` (output alloc) | 5.3 | 7.2% | Yes — pre-alloc workspace |

### Backward (total Python: ~95 us/call)

| Function | us/call | % | Actionable? |
|----------|---------|---|-------------|
| `_flash_attn_bwd` body (validation, tile config, compile key) | 22.6 | 23.8% | Yes — fast-path |
| `active_fake_mode()` (4x/call) | 9.4 | 9.9% | Yes — cache or check once |
| `_bwd_preprocess` dispatch wrapper | 8.5 | 8.9% | Yes — inline/reduce overhead |
| `torch.empty_like` (6 tensor allocs) | 7.9 | 8.3% | Yes — workspace pre-alloc |
| CuTeDSL `wrapper` | 6.7 | 7.1% | Upstream |

## Suggested Optimizations

1. **Cache `is_fake_mode()` at call entry**: Check once in `flash_attn_func` and thread through, instead of re-checking 3-4x per call via `is_fake_mode()` -> `active_fake_mode()` -> `_get_current_dispatch_mode_stack()`.

2. **Cache `get_device_properties()`**: Called every forward to determine tile sizes. The result never changes for a given device. Use `@lru_cache` or module-level singleton.

[...]

PR #2516, opened 2026-04-28
[FA4] Reduce per-call Python dispatch overhead

shivam2199:
## Summary

Targets the ~20-25us host-side overhead per `flash_attn_func` call reported in #2426. Focuses on the cacheable, pure-function work hit every forward and backward, without changing kernel semantics.

### What changed

1. **Cache `is_fake_mode()` per dispatch.** `is_fake_mode()` re-enters torch's dispatch-mode stack (~2us each) and was being called 3-4 times per forward (and similarly for backward). Computed once at `_flash_attn_fwd` / `_flash_attn_bwd` entry, threaded through `_validate_tensor` and the internal helpers (`_bwd_preprocess`, `_bwd_postprocess_convert`, `_flash_attn_fwd_combine`) via an optional `fake_mode=` kwarg that defaults to `None` (falls back to the original call).

2. **Cache `num_SMs` per device.** `torch.cuda.get_device_properties(device).multi_processor_count` runs every forward for the SplitKV heuristic. Added `@lru_cache` `_get_num_sms(device_index)` matching the existing pattern on `_get_device_arch`.

3. **Cache pure config helpers.** Added `@lru_cache(maxsize=None)` to `_validate_head_dims`, `_tile_size_fwd_sm90`, `_tile_size_bwd_sm90`. All three take small ints and return frozen dataclasses (or `None`); they run on every dispatch. `lru_cache` does not cache exceptions, so invalid shapes still raise.

4. **Benchmark script** at `benchmarks/benchmark_dispatch_overhead.py` that reproduces the wall-vs-gpu-time methodology from #2426 so the improvement is measurable and reviewable.

### What this PR does not do

- **No shape-keyed fast-path that memoizes the full `_flash_attn_fwd` setup block.** That block interleaves pure-function resolution with per-call tensor allocations (`out`, `lse`, `out_partial`) and shape-dependent normalization (block sparsity, aux tensors). Safely memoizing it would need a restructure. Happy to follow up if maintainers want that.
- **No backward workspace pool.** Torch's caching allocator already makes `torch.empty_like` cheap (~1-2us amortized), and a pool introduces stream-ordering / concurrent-backward correctness risk. Not worth it here.
- **CuTeDSL / TVM FFI `wrapper` overhead (~7-10us)** is upstream in `nvidia-cutlass-dsl`, out of scope.

### Expected impact

The OP's profile of `_flash_attn_fwd` body attributes ~7us to `is_fake_mode()` calls and ~1-2us to device-property lookups — Changes 1+2 should recover that directly. The `@lru_cache` decorators on pure helpers save additional dict/hash work on each call. Absolute numbers will vary by host (the OP measured on H100 SXM); the benchmark script is included so reviewers can reproduce on their own hardware.

I don't have a GPU host to attach before/after numbers to the PR body; will fill in once someone runs `benchmarks/benchmark_dispatch_overhead.py` on the same H100 / Blackwell setup.

### Review surface

The behavioral risk is concentrated in the `fake_mode=` plumbing: on cache miss the helper still calls `is_fake_mode()` itself (unchanged behavior), so existing callers that don't pass the kwarg see no difference. The `@lru_cache` decorators are on functions whose arguments are all hashable scalars.

## Test plan

- [ ] `pytest tests/cute/test_flash_attn.py -k "fake_tensor or seqlen_k_zero" -x` still passes
- [ ] `pytest tests/cute/test_flash_attn.py -x` smoke suite
- [ ] `pytest tests/cute/test_flash_attn_varlen.py -x`
- [ ] `pytest tests/cute/test_flash_attn_combine.py -x` (exercises `_flash_attn_fwd_combine` with new `fake_mode=` kwarg)
- [ ] Run `python benchmarks/benchmark_dispatch_overhead.py` on main and on this branch; report overhead deltas in the PR thread

Refs #2426.

### DIFF
--- a/benchmarks/benchmark_dispatch_overhead.py
+++ b/benchmarks/benchmark_dispatch_overhead.py
@@ -1,0 +1,80 @@
+"""Measure FA4 CuTe Python dispatch overhead.
+
+Reproduces the methodology from
+https://github.com/Dao-AILab/flash-attention/issues/2426 to compare host-side
+dispatch cost (wall time minus GPU kernel time) across sequence lengths.
+
+Run:
+    python benchmarks/benchmark_dispatch_overhead.py
+
+Outputs a table of (seqlen, wall_ms_median, gpu_ms_median, overhead_us,
+overhead_pct). The overhead column is the metric optimized by #2515.
+"""
+
+import argparse
+import statistics
+import time
+
+import torch
+
+from flash_attn.cute import flash_attn_func
+
+
+def bench_one(seqlen: int, batch: int, heads: int, head_dim: int, causal: bool, iters: int):
+    device = "cuda"
+    dtype = torch.bfloat16
+
+    q = torch.randn(batch, seqlen, heads, head_dim, device=device, dtype=dtype)
+    k = torch.randn(batch, seqlen, heads, head_dim, device=device, dtype=dtype)
+    v = torch.randn(batch, seqlen, heads, head_dim, device=device, dtype=dtype)
+
+    # Warmup to populate compile caches.
+    for _ in range(5):
+        out, _ = flash_attn_func(q, k, v, causal=causal)
+    torch.cuda.synchronize()
+
+    wall = []
+    gpu = []
+    start_evt = torch.cuda.Event(enable_timing=True)
+    end_evt = torch.cuda.Event(enable_timing=True)
+    for _ in range(iters):
+        torch.cuda.synchronize()
+        t0 = time.perf_counter_ns()
+        start_evt.record()
+        out, _ = flash_attn_func(q, k, v, causal=causal)
+        end_evt.record()
+        torch.cuda.synchronize()
+        t1 = time.perf_counter_ns()
+        wall.append((t1 - t0) / 1e6)  # ms
+        gpu.append(start_evt.elapsed_time(end_evt))  # ms
+
+    return statistics.median(wall), statistics.median(gpu)
+
+
+def main():
+    parser = argparse.ArgumentParser()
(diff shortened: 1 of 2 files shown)


shivam2199:
Ran `benchmarks/benchmark_dispatch_overhead.py` on A100 80GB (AWS p4de), CUDA 12.8 / torch 2.x / cu128, 500 iterations per config, warm DSL cache:

```
                  before (main)    after (this PR)    delta
seqlen=64         19.8us           19.5us             -0.3us
seqlen=128        20.0us           19.8us             -0.2us
seqlen=256        20.6us           19.7us             -0.9us
seqlen=512        20.0us           20.0us              0.0us
seqlen=1024       21.2us           20.9us             -0.3us
seqlen=2048       22.6us           22.6us              0.0us
seqlen=4096       23.2us           22.7us             -0.5us
```

Mean delta across all seqlens: **-0.3us on a ~20us baseline** — within iteration-level jitter on this hardware. The cheap Python-side caches in this PR (`@lru_cache` on tile-size helpers, deduped `is_fake_mode()`, cached `multi_processor_count`) each save <1us, which does not add up to a measurable improvement.

The ~20us floor on A100 appears dominated by CuTeDSL FFI dispatch and `autograd.Function` entry — both outside the scope of this PR. The OP's original profile flagged FFI as upstream in `nvidia-cutlass-dsl`; that matches what I'm seeing.

Closing: no measurable win justifies the maintenance cost of the `fake_mode=` kwarg plumbing. Happy to reopen if someone sees a path that actually moves the needle, or if a reviewer wants me to land just the `@lru_cache` decorators (no API change) as a pure readability improvement.

PR #2224, opened 2026-02-02
[CuTe,Flex] varlen blocksparsity

reubenconducts:
This PR extends blocksparsity to the variable sequence length case. Whereas in batched blocksparsity the metadata tensors take the shapes
```python
*_block_cnt: [batch_size, num_heads, num_m_blocks]
*_block_idx: [batch_size, num_heads, num_m_blocks, num_n_blocks],
```
in varlen blocksparsity, we pack our metadata tensors to take the shapes
```python
*_block_cnt: [num_heads, total_m_blocks]
*_block_idx: [num_heads, total_n_blocks]
```
where `total_m_blocks` is the sum of all `m` blocks per head (equiv. number of work tiles per head) and `total_n_blocks` is the total of all `n` blocks potentially processed per head across all sequences in the batch. For example, consider a varlen batch with sequences contained in `seqlens_q` and `seqlens_k`. At batch index `b`, we let 
```tex
num_m(b) = ceildiv(seqlens_q[b], tile_m)
num_n(b) = ceildiv(seqlens_k[b], tile_n)
```
and define
```tex
total_m_blocks = sum_{b \in B} num_m(b)
total_n_blocks = sum_{b \in B} num_m(b) * num_n(b)
```

To properly index into the blocksparsity tensors, we use auxiliary `mCuTotalMBlocks` and `mBlockIdxOffsets` tensors, which can be prepared on host. 

cc @drisspg @v0i0 

NOT INTENDED FOR THIS PR:
- bwd support

PR #2520, opened 2026-04-29
[Cute] Fix O(N^2) per-CTA scan in SingleTileVarlenScheduler

horakka5:
## Summary

This PR fixes the **O(N²) per-CTA scan in `SingleTileVarlenScheduler`** (`flash_attn/cute/tile_scheduler.py`). It does *not* touch the hd=256 dedicated path, which uses a different scheduler (`Sm100FmhaStaticTileScheduler`) and has its own — distinct — varlen scaling issue (see *Out of scope* below).

`SingleTileVarlenScheduler._varlen_coord_map` mapped a flat CTA index to `(batch, head, m_block)` by linearly scanning batches in groups of 31 (one warp) on the GPU, per CTA. With N batches, every CTA did up to ~N/31 warp-level prefix-sum iterations, giving total scheduling work O(N²). Pathological for many short sequences — at N=8192 / seq_len=34 / 12h / hd=128 / bf16, forward varlen is **73× slower** than the same compute over a dense (padded) layout.

The fix adds the metadata pass from FA4 paper §3.3 ("Scheduling"): the host precomputes per-batch cumulative tile counts (`_compute_tile_cumsum` in `interface.py`); the kernel binary-searches it to skip to the right group of 31 batches in O(log N), then runs the existing in-group warp prefix-sum unchanged. The LPT batch sort the paper also describes is left for a follow-up — the metadata this PR adds is a prerequisite for it.

## Numbers

A/B captured by `git checkout` between commits in this PR (B=8192, seq=34, 12h, hd=128, bf16):

| metric | pre-fix | post-fix | speedup |
|---|---:|---:|---:|
| forward varlen | 68.9 ms | 4.95 ms | **13.9x** |
| backward varlen | 308 ms | 27.1 ms | **11.4x** |
| forward vl/dn ratio | 73.5x | 5.3x | |
| backward vl/dn ratio | 23.7x | 2.1x | |
| per-seq fwd cost (us, flat across N=256..8192) | 0.5 → 8.4 | 0.59 → 0.65 | |

Linear-fit R² on post-fix forward: 0.99990 (so per-seq cost is essentially flat — the asymptotic was the bug).

Reproduce with `benchmarks/benchmark_varlen_scheduler.py` (added in this PR).

The remaining 5.3x vl/dn ratio at the worst-case (N=8192, seq=34) is real varlen overhead (per-batch TMA descriptor setup, padded `dq_accum`); the scheduler is no longer a contributor. Once seq_len rises to 1024 the ratio collapses to 1.6x.

## Scope

`SingleTileVarlenScheduler` is instantiated by 9 kernel files; this PR threads the optional `mTileCumsum: cute.Tensor` argument through every one of them:

- `flash_fwd.py` (SM80 + SM120 via inheritance)
- `flash_fwd_sm90.py`
- `flash_fwd_sm100.py` (SM100/SM110)
- `flash_fwd_mla_sm100.py` (MLA absorbed)
- `flash_bwd.py` (SM80 + SM120)
- `flash_bwd_sm90.py`
- `flash_bwd_sm100.py`
- `flash_bwd_preprocess.py`
- `flash_bwd_postprocess.py`

The hd=256 dedicated kernel uses a different scheduler (`Sm100FmhaStaticTileScheduler`) without an `mTileCumsum` slot, so the `interface.py` plumbing is gated behind `not use_dedicated_hd256_kernel`.

Public API is unchanged. The kernel parameter is optional with a `None` fallback to the original linear-scan path, which preserves the direct-kernel-callable contract documented in `usage.md`. `interface.py` always passes a real cumsum when varlen is enabled, so the production path never falls back.

## Correctness

- **Bit-identical** to the linear-scan path for forward, dq, dv across MHA / GQA / MQA at B in {64, 128, 512, 4096} (verified by `torch.equal`).
- dk on GQA/MQA shows ~6e-5 max diff vs the linear scan — but the same magnitude as cumsum-on vs cumsum-on with identical inputs (1.5e-5), i.e. it's the FA4 backward's intrinsic atomic-add non-determinism described in paper §3.2.4 ("Deterministic backward pass") and is unchanged by this patch.
- vs PyTorch SDPA reference at B=64 / S=128 / H=8 / D=128 / bf16: fwd max_abs 2.4e-04, dq 3e-05, dk 6e-05, dv 0.0 — well within bf16 tolerance.
- `pytest tests/cute/test_flash_attn_varlen.py -k "(7 or 20) and dtype1 and (mha or gqa or mqa) and 64"`: **828 passed, 0 failed** (covers MHA/GQA/MQA × hd in {64, 128, 256} × the existing parametrize matrix).

## Tests added

| Test | Marker | Purpose |
|---|---|---|
| `test_varlen_scheduler_scaling` | `@slow` | Asserts `per_seq_us[8192] < 2 * per_seq_us[128]`. Guards against re-introduction of O(N²). With the fix the ratio is ~1; without it, ~13. Lives at the regression-test commit so reviewers can see it fail on the previous commit. |
| `test_varlen_high_batch_skewed_correctness` | `@slow` | B=1024 with seqlens uniformly random in [1, 128]; validates fwd+bwd against PyTorch SDPA reference. The existing `test_varlen` caps at B=20; this is the high-batch+skewed regime where the scheduler change matters most. |

Slow tests are excluded by the existing CI filter, so default CI stays green.

## Pre-existing issues observed during verification (NOT introduced by this PR)

- **sm_120a forward FakeTensor compile fails** with `ValueError: Operation creation failed` in `flash_fwd.py:386` (the `epilogue` function, SM80 base via SM120 inheritance). Present on pristine `upstream/main` (`ba59def`).
- **hd=256 dedicated path also exhibits super-linear varlen scaling**, but for a *different* reason: `BlackwellFusedMultiHeadAttentionForward.__call__` in `flash_attn/cute/sm100_hd256_2cta_fmha_forward.py:280-282` sets the M dimension of the FMHA grid from `q_tensor.shape[0]` (= `total_q` = `N × mean_seqlen` for varlen) instead of from `max_seqlen_q`. Result: total tiles `M × B × H` ≈ `(N × mean_seqlen / tile_m) × N × H` = O(N²), most of them no-ops filtered by `check_valid_work_for_seqlen_q`. Reproduces identically on pristine `upstream/main`. Filing as a separate issue #2521 . Fix shape: thread `max_seqlen_q` (already on the public API) into the hd256 kernels and use it for `s_q_total` when varlen — non-trivial because it touches recently-landed code (#2412) across 3 hd256 kernel files.

## Out of scope

- **LPT batch sort** (FA4 paper §3.3 second half) — orthogonal load-balancing improvement, requires its own re-validation (changes the order of inter-Q-head atomic adds on GQA backward), and is most useful for mixed prefill+decode workloads. Best as a follow-up PR; the cumsum metadata this PR adds is a prerequisite for it.
- **hd=256 varlen grid sizing** (different scheduler, different mechanism, different kernel — described above). See #2521 

## Commits

1. `[Cute,Bench] Add varlen tile-scheduler stress benchmark` — script that demonstrates the regression
2. `[Cute,Test] Regression test for FA4 varlen scheduler scaling` — fails on this commit, passes after the fix
3. `[Cute] Fix O(N²) varlen tile scheduler via precomputed cumsum` — the fix
4. `[Cute,Test] High-batch correctness for FA4 varlen vs SDPA` — independent additional coverage

### DIFF
--- a/benchmarks/benchmark_varlen_scheduler.py
+++ b/benchmarks/benchmark_varlen_scheduler.py
@@ -1,0 +1,189 @@
+"""Stress benchmark for the FA4 (CuTe-DSL) varlen tile scheduler.
+
+Targets the kernels in ``flash_attn/cute/`` (FA4) — specifically
+``SingleTileVarlenScheduler`` in ``flash_attn/cute/tile_scheduler.py`` — and is
+not applicable to the older FA2 (``csrc/``) or FA3 (``hopper/``) generations.
+
+The scheduler maps a flat CTA index to ``(block, head, batch)`` by scanning
+batches in groups of 31 (one warp) on the GPU, per CTA. Without precomputed
+metadata each CTA does an O(num_batches / 31) linear scan, giving
+O(num_batches^2) total scheduling work across the launched grid. The
+pathological regime is many short sequences, where per-tile scheduling cost
+swamps the actual attention compute.
+
+This script sweeps num_seqs with the per-seq length fixed to a small value
+(seq_len=34 by default, the regime where the scheduler dominates). True
+attention compute scales linearly in num_seqs at fixed seq_len, so any
+super-linear growth in the per-iter time comes from the scheduler.
+"""
+
+import argparse
+
+import torch
+
+from flash_attn.cute import flash_attn_func, flash_attn_varlen_func
+
+
+def make_fixed_len_data(num_seqs, seq_len, num_heads, head_dim, device, dtype):
+    total = num_seqs * seq_len
+    q = torch.randn(total, num_heads, head_dim, dtype=dtype, device=device) * 0.1
+    k = torch.randn(total, num_heads, head_dim, dtype=dtype, device=device) * 0.1
+    v = torch.randn(total, num_heads, head_dim, dtype=dtype, device=device) * 0.1
+    cu = torch.arange(0, total + 1, seq_len, dtype=torch.int32, device=device)
+    return q, k, v, cu
+
+
+def make_padded_view(q, k, v, num_seqs, seq_len):
+    """Reshape packed (total, H, D) tensors to dense (num_seqs, seq_len, H, D) — no copy."""
+    H, D = q.shape[1], q.shape[2]
+    return (
+        q.view(num_seqs, seq_len, H, D),
+        k.view(num_seqs, seq_len, H, D),
+        v.view(num_seqs, seq_len, H, D),
+    )
+
+
+def cuda_bench(fn, warmup=3, iters=20):
+    """Time a callable with CUDA events. Returns mean ms per call."""
+    for _ in range(warmup):
+        fn()
+    torch.cuda.synchronize()
+    starts = [torch.cuda.Event(enable_timing=True) for _ in range(iters)]
+    ends = [torch.cuda.Event(enable_timing=True) for _ in range(iters)]
+    for s, e in zip(starts, ends):
+        s.record()
+        fn()
+        e.record()
+    torch.cuda.synchronize()
+    times = [s.elapsed_time(e) for s, e in zip(starts, ends)]
+    return sum(times) / len(times)
+
+
+def bench_one(num_seqs, seq_len, num_heads, head_dim, dtype, device, with_bwd):
+    q, k, v, cu = make_fixed_len_data(num_seqs, seq_len, num_heads, head_dim, device, dtype)
+    qp, kp, vp = make_padded_view(q, k, v, num_seqs, seq_len)
+
+    def fwd_varlen():
+        out = flash_attn_varlen_func(
+            q, k, v,
+            cu_seqlens_q=cu, cu_seqlens_k=cu,
+            max_seqlen_q=seq_len, max_seqlen_k=seq_len,
+        )
+        if isinstance(out, tuple):
+            out = out[0]
+        return out
+
+    def fwd_dense():
+        out = flash_attn_func(qp, kp, vp)
+        if isinstance(out, tuple):
+            out = out[0]
+        return out
+
+    fwd_varlen_ms = cuda_bench(fwd_varlen)
+    fwd_dense_ms = cuda_bench(fwd_dense)
+
+    fb_varlen_ms = fb_dense_ms = float("nan")
+    if with_bwd:
+        def fb_varlen():
+            qg = q.detach().requires_grad_(True)
+            kg = k.detach().requires_grad_(True)
+            vg = v.detach().requires_grad_(True)
+            out = flash_attn_varlen_func(
+                qg, kg, vg,
+                cu_seqlens_q=cu, cu_seqlens_k=cu,
+                max_seqlen_q=seq_len, max_seqlen_k=seq_len,
+            )
+            if isinstance(out, tuple):
+                out = out[0]
+            out.sum().backward()
+
+        def fb_dense():
+            qg = qp.detach().requires_grad_(True)
+            kg = kp.detach().requires_grad_(True)
+            vg = vp.detach().requires_grad_(True)
+            out = flash_attn_func(qg, kg, vg)
+            if isinstance(out, tuple):
+                out = out[0]
+            out.sum().backward()
+
+        fb_varlen_ms = cuda_bench(fb_varlen, iters=10)
+        fb_dense_ms = cuda_bench(fb_dense, iters=10)
+
+    return fwd_varlen_ms, fwd_dense_ms, fb_varlen_ms, fb_dense_ms
+
+
+def main():
+    p = argparse.ArgumentParser()
+    p.add_argument("--seq-len", type=int, default=34,
+                   help="per-sequence length (small => scheduler dominates)")
+    p.add_argument("--num-heads", type=int, default=12)
+    p.add_argument("--head-dim", type=int, default=128)
+    p.add_argument("--dtype", choices=["bf16", "fp16"], default="bf16")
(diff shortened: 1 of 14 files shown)


tridao:
This makes sense. However I was thinking typically num batch <= 128. What's the use case for very large batch with short seqlen?
Computing tile_cumsum would incur an extra kernel launch for LLM decode (one can amortize that by computing once and use for all layers, but the interface gets a bit more complicated).
Cc @jayhshah @drisspg: what do you think about the tradeoff here?

jayhshah:
I've noticed this as well and instead locally implemented a switch to non-varlen persistent scheduler conditional on `seqlen_q_packgqa <= 2 * tile_m` as deduced from `max_seqlen_q` along with the tma store for O using ragged tma; I would expect that to still be faster than varlen even with the binary search (and it avoids the extra kernel call). However this solution makes sense for the general case.

For small batch we get a slowdown -- can we see a comparison for that?

tridao:
One (over-engineered) soln is if we can assume batch <= 1024, we load cu_seqlens from gmem -> smem, build 2 cumsum arrays (1 array size 1024 for partial cumsum of each 32 elements, one array of size 32 for cumsum of (sum of 32 segments)). Then to look up, you do binary search twice (one of the 32 sized array, one on the 1024 sized array).
This can all happen in the attn kernel done by each CTA, no extra kernel needed. 
The downside is that we have to assume batch <= 1024, and it's kinda annoying to impl.

horakka5:
Thanks @tridao @jayhshah. I spent some time chasing correctness issues I hit during verification; these should be fixed in the commit above. I have tested the fixes for the SM100 path only. For other platforms I had to rely on a code-level audit.

Fair point on small batches. The regression below ~B=256 is real (table below). The motivating workload is BERT-style encoders trained with in-batch contrastive loss, where effective B sits in the 1k–8k range with sequences averaging tens to hundreds of tokens, and master's per-CTA O(B) scan is the bottleneck. One option would be to gate the cumsum on `num_batch >= 512` so master's behavior is preserved on small batches (passing `None` for `mTileCumsum` already triggers the linear-scan fallback).

A/B per-call latency on B-class hardware, MHA hd=128 bf16. Ratio = `on / off` (this PR / master); <1 means this PR is faster.

| S \ B | 8 | 64 | 256 | 512 | 1024 | 2048 | 4096 | 8192 |
|---:|---:|---:|---:|---:|---:|---:|---:|---:|
| 34   | 2.95× | 2.98× | 1.24× | 0.81× | 0.53× | 0.32× | 0.17× | 0.07× |
| 2048 | 1.18× | 1.05× | 1.00× | 0.95× | 0.89× | 0.73× | — | — |
| 8192 | 1.10× | 1.02× | 1.00× | 0.99× | 0.96× | 0.92× | — | — |

Absolute times for S=2048 to anchor the ratios:

| B | master (μs) | this PR (μs) | Δ (μs) |
|---:|---:|---:|---:|
| 8    | 114   | 135   | +21 |
| 64   | 768   | 810   | +42 |
| 256  | 3504  | 3516  | +12 |
| 512  | 7484  | 7131  | -353 |
| 1024 | 16266 | 14461 | -1805 |
| 2048 | 39538 | 29026 | -10512 |

The per-call regression at small B comes from two sources: a roughly constant ~60μs of host-side cumsum compute (4 small kernel launches), and the in-kernel batch lookup (binary search vs warp scan) which costs more at moderate B and saves a lot at high B. I tried to isolate them quantitatively but the diffs are within timing noise on the small-B configs, so I'm not posting a breakdown.

@tridao: on the amortization — agreed it would complicate the interface, and I haven't tried it. The gate proposal above (only enabling cumsum at `num_batch >= 512`) would skip the cumsum entirely for typical decode batch sizes, which I think handles the same concern without any interface change. For training-style workloads where the gate fires, amortizing across layers would still be a win, but probably worth doing only if there's a clear way to do it without exposing extra state.

The in-kernel two-level scan you suggested is a more substantial redesign — I'd leave that call to you.

@jayhshah: thanks, I don't fully understand all the implications of that approach. One question that came to mind — does the `seqlen_q_packgqa <= 2*tile_m` condition hold up well when most sequences in a batch are short but a few are long (e.g., short queries mixed with longer documents)? I read it as gating on `max_seqlen_q`, in which case a single outlier would disable the switch — but I may be misreading. Genuinely curious how it behaves in that regime.

reubenconducts:
Could you change the name `mTileCumsum` to `mCuTotalMBlocks` to match that in #2224? Additionally, there is no need to build `num_heads` (and hence `pack_gqa`) into `mCuTotalMBlocks` - it is handled properly already in the tile scheduler. I've split out the binary search into a separate utility in #2556, so you can use that.

horakka5:
@reubenconducts Thank you for the review and pointers to #2224 and #2556. I have introduced the following changes:
- renamed `mTileCumsum` to `mCuTotalMBlocks` and applied the change across all 9 kernels
- replaced inline binary search with `utils.get_batch_from_cu_tensor`
- dropped `num_head` from cumsum (as well as the architecture-based branching it required)

I have only limited insight into the codebase, so I kept the following in the host helper: When `pack_gqa`, I apply `seqlen *= qhead_per_kvhead` before computing the per-batch `m_block` count. To the best of my understanding, this reflects the logic from kernel-side `_get_num_m_blocks` and therefore is needed (when I tried dropping it, several GQA/MQA tests started failing). Happy to change if there is a better place for this.

I have verified correctness on SM100 using the extended test suite. I do not have easy access to other hardware to test on.

PR #2525, opened 2026-04-30
gc.collect needs to be run for every test case to avoid oom on py314 for hopper

smoothsmooth:


smoothsmooth:
@tridao Would you mind taking a look at this PR? Py314 has some issues in gc. Thanks!

PR #2526, opened 2026-04-30
Fix CK varlen num_splits plumbing

TideDra:
## Summary
- Add the trailing `num_splits` argument to the ROCm CK `mha_varlen_fwd` declaration and definition.
- Use the caller-provided `num_splits` value instead of hardcoding `0` before CK split-kv selection.
- Align the CK backend with the Python wrapper and CUDA implementation after #2448/#2476.

## Tests
- `git diff --check`
- `python3` signature alignment check


ZiguanWang:
meet the same problem, thanks

PR #2527, opened 2026-04-30
Fix test_flash_attn_fast varlen call after qv positional insert

henrylhtsang:
## Summary
- `flash_attn_varlen_func` now takes `qv` as its 4th positional parameter, which shifted the existing positional call in `tests/cute/test_flash_attn_fast.py` so `cu_seqlens_q/k` and `max_seqlen_q/k` got bound to the wrong slots.
- This caused 132 `AttributeError: 'int' object has no attribute 'shape'` failures at `flash_attn/cute/interface.py:370`, e.g. `test_flash_attn_varlen_output[128-64-False-mha-dtype0]`.
- Fix: switch the call site to keyword arguments. After the fix, the suite passes 240/240.

## Test plan
- [x] `pytest tests/cute/test_flash_attn_fast.py` — 240 passed

cc @jayhshah

### DIFF
--- a/tests/cute/test_flash_attn_fast.py
+++ b/tests/cute/test_flash_attn_fast.py
@@ -139,8 +139,8 @@ def test_flash_attn_varlen_output(seqlen, d, causal, mha_type, dtype):
 
     out_varlen, lse = flash_attn_varlen_func(
         q_varlen, k_varlen, v_varlen,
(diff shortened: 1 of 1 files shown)


jayhshah:
Thanks!

jayhshah (approved):

PR #2542, opened 2026-05-06
Fully enable varlen split KV 

liangel-02:
**Context**

today, for split KV varlen with paging, the input gets reshaped to include batch dimension before it hits the split KV kernel, so it's not actually varlen. this PR fully enables varlen split KV by keeping the ragged sequences. we need to modify the way we compute address indices to convert between rectangular and padded layouts. 

**What changed**

in the combine kernel for split KV, the partial results for oaccum (partial weight output) + lseaccum (partial lse) get merged. first it reads the partial results from all splits and then writes the global result to the output lse/output tensors. but for varlen, we need to make sure that that the correct addresses are computed. we do this by using `cu_seqlens_q` to properly calculate the boundaries of the tensors to skip the padding rows. 

(the padding comes from here where the rectangular buffers get allocated) 
https://github.com/Dao-AILab/flash-attention/blob/8a8b2f10ddca88fd46db406c3e143e1ab0af977f/csrc/flash_attn/flash_api.cpp#L320-L321

**Testing**

Screenshot 2026-05-19 at 12 49 13 PM
Screenshot 2026-05-19 at 12 26 30 PM

### DIFF
--- a/csrc/flash_attn/flash_api.cpp
+++ b/csrc/flash_attn/flash_api.cpp
@@ -698,14 +698,11 @@ mha_varlen_fwd(at::Tensor &q,  // total_q x num_heads x head_size, total_q := \s
     params.page_block_size = page_block_size;
     // Keep references to these tensors to extend their lifetime
     at::Tensor softmax_lse_accum, out_accum;
-    if (seqlenq_ngroups_swapped) {
+    if (seqlenq_ngroups_swapped || paged_KV) {
         std::tie(softmax_lse_accum, out_accum) =
             set_params_splitkv(params, batch_size, num_heads, head_size,
                                max_seqlen_k, max_seqlen_q, head_size_rounded,
                                p_dropout, num_splits, get_num_sm(get_current_device()), opts);
-    } else if (paged_KV) {
-        TORCH_CHECK(num_splits <= 1, "num_splits > 1 is not supported for varlen paged KV");
-        params.num_splits = num_splits;
     }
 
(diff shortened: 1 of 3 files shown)


drisspg:
This is my understanding:

Before #2448, mha_varlen_fwd only called set_params_splitkv(...) here:
```c++
if (seqlenq_ngroups_swapped) {
    set_params_splitkv(..., /*num_splits*/ 0, ...);
}
```
And seqlenq_ngroups_swapped required: `max_seqlen_q == 1`
That means varlen forward only got splitKV when decode q_size=1  + packgqa. Which intern

```c++
cu_seqlens_q_d = nullptr;
max_seqlen_q = ngroups;
```

After #2448, the condition became effectively:
```c++
if (paged_KV || seqlenq_ngroups_swapped) {
    set_params_splitkv(..., num_splits, ...);
}
```
So now varlen can enter the multi-split combine path without the max_seqlen_q == 1 / ngroups-swapped transformation. 

And now we are improperly indexing into the output tensor. 


More generally I think this codepath could use some very targeted fuzzing. I think the Combine path is probably making other assumptions, that could have effects on correctness. We are now sending ragged sequences do we need to account for padding from Max_seqlen_q being beyond each group? 

Can we run compute sanitizer to be sure. It will probably not say anything because we are still in valid mem. Regarldess we should do more testing

liangel-02:
closing, performance gains are minimal 
Screenshot 2026-05-21 at 1 21 53 PM

PR #2543, opened 2026-05-06
[CuTe,Bwd,Sm100] don't disable 2cta due to cuda 12 in bwd

reubenconducts:


jayhshah:
This PR follows up on https://github.com/Dao-AILab/flash-attention/pull/2461 to only disable 2cta for fwd on cuda 12. 2cta for backward is still faster in general and is necessary for hdim 192.

jayhshah (approved):

PR #2549, opened 2026-05-08
[Cute,Bwd,Sm100] fix incorrect calculation of n_block global max for bwd deterministic

jayhshah:
PR https://github.com/Dao-AILab/flash-attention/pull/2253 refactored the bwd dQ deterministic code by introducing a new `get_n_block_max_for_m_block` method, however this took in an incorrect calculation of `n_block_global_max` by using the cta rather than the cluster n tile (thus doubling the value), which sometimes breaks deterministic mode for 2cta backward. Note that BlockInfo's tile_n is the indeed the cluster tile_n by definition in the kernel.

This PR fixes the bug by determining `n_block_global_max` inside `get_n_block_max_for_m_block`, matching the existing logic in `get_n_block_min_max`.

@drisspg

### DIFF
--- a/flash_attn/cute/block_info.py
+++ b/flash_attn/cute/block_info.py
@@ -143,8 +143,8 @@ class BlockInfo:
         self,
         seqlen_info: SeqlenInfoQK,
         m_block: Int32,
-        n_block_global_max: Int32,
     ) -> Int32:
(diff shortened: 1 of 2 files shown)


drisspg (approved):
Ahh good find, another reason why the tiling+cluster meaning can be hard to grok. I wonder if non colliding names would be helpful here

PR #2468, opened 2026-04-17
Add page table support for MLA SM100 forward kernel

Johnsonms:
### Summary

Motivated from https://github.com/Dao-AILab/flash-attention/pull/2441

Enable paged KV cache for the MLA (Multi-head Latent Attention) forward kernel on SM100/Blackwell, targeting inference serving with DeepSeek-family models (vLLM/SGLang).

- **Phase 1 (TMA paged KV)**: Page table support for MLA with `page_size == tile_n` (128), using TMA bulk copies for zero-overhead paged KV access.
- **Phase 2 (cp.async paged KV)**: Extends paged KV to arbitrary `page_size` (16, 64, etc.) using cp.async loads, enabling compatibility with vLLM/SGLang page sizes.
- **Benchmark**: Adds `tests/cute/benchmark_mla_paged_kv.py` for paged vs non-paged MLA performance comparison.

### Benchmark Results (B200, batch=128, seqlen_q=1, MQA-128, bf16)

| Seqlen_k | Non-paged | Paged-128 (TMA) | Paged-64 (cp.async) | Paged-16 (cp.async) |
|----------|-----------|-----------------|---------------------|---------------------|
| 2K | 800 TFLOPS | 799 TFLOPS (-0.1%) | 616 TFLOPS (-23%) | 607 TFLOPS (-24%) |
| 4K | 1013 TFLOPS | 1020 TFLOPS (+0.7%) | 822 TFLOPS (-19%) | 811 TFLOPS (-20%) |
| 8K | 1178 TFLOPS | 1176 TFLOPS (-0.2%) | 966 TFLOPS (-18%) | 952 TFLOPS (-19%) |
| 16K | 1240 TFLOPS | 1231 TFLOPS (-0.7%) | 1006 TFLOPS (-19%) | 1019 TFLOPS (-18%) |
| 32K | 1259 TFLOPS | 1248 TFLOPS (-0.9%) | 1018 TFLOPS (-19%) | 1039 TFLOPS (-17%) |
| 64K | 1262 TFLOPS | 1252 TFLOPS (-0.8%) | 1061 TFLOPS (-16%) | 1053 TFLOPS (-17%) |

- **TMA paged (page_size=128)**: zero overhead vs non-paged
- **cp.async (page_size 16/64)**: ~16-24% overhead (expected for MLA dv=512 bandwidth)
- **Peak**: 1262 TFLOPS at seqlen=64K

### Test plan
[...]

PR #2558, opened 2026-05-13
[CuTe,Fwd,Sm100] refactor mla sm100 forward and add page table

jayhshah:
Superseding https://github.com/Dao-AILab/flash-attention/pull/2468.

We refactor the kernel file to remove duplication around splitting V and add page table support.

We also support not providing Q and K (which semantically mean q_pe and k_pe). This computes attention according to the formula
```
O = softmax(scale * (Qv @ V.T)) @ V
```
as in Deepseek v4 core attention.

### DIFF
--- a/benchmarks/benchmark_mla_paged_kv.py
+++ b/benchmarks/benchmark_mla_paged_kv.py
@@ -1,0 +1,108 @@
+# Copyright (c) 2025, Johnsonms.
+
+# We recommend locking GPU clocks before running the benchmark to ensure consistent results.
+# This can be done using the following commands (2619 MHz is the max clock for B200):
+# sudo nvidia-smi -i 0 -pm 1
+# sudo nvidia-smi -i 0 --lock-gpu-clocks 2619,2619
+# See more here: https://github.com/triton-lang/triton/blob/d9f10ebdc5da53f73eb852fde73d8d7d80b679d1/python/triton/testing.py#L487
+
+import time
+import torch
+
+from triton.testing import do_bench
+
+from flash_attn.cute.interface import flash_attn_varlen_func
+
+
+device = "cuda"
+dtype = torch.bfloat16
+seqlen_q = 1
+nheads_q = 128
+nheads_kv = 1  # MQA-128
+headdim = 64
+headdim_v = 512
+causal = True
+
+batch_size = 128
+page_sizes = [None, 16, 64, 128]  # None = non-paged baseline
+
+torch.manual_seed(0)
+
+print(f"\nMLA paged KV, nheads_q = {nheads_q}, nheads_kv = {nheads_kv}, headdim = {headdim}, headdim_v = {headdim_v}, causal = {causal}")
+
+for seqlen in [s * 1024 for s in [1, 2, 4, 8, 16, 32, 64]]:
+    # Varlen format: (total_tokens, nheads, hdim)
+    total_q = batch_size * seqlen_q
+    total_k = batch_size * seqlen
+
+    try:
+        q = torch.randn(total_q, nheads_q, headdim, dtype=dtype, device=device)
+        k = torch.randn(total_k, nheads_kv, headdim, dtype=dtype, device=device)
+        v = torch.randn(total_k, nheads_kv, headdim_v, dtype=dtype, device=device)
+        qv = torch.randn(total_q, nheads_q, headdim_v, dtype=dtype, device=device)
+    except torch.OutOfMemoryError:
+        continue
+
+    cu_seqlens_q = torch.arange(0, total_q + seqlen_q, seqlen_q, dtype=torch.int32, device=device)
+    cu_seqlens_k = torch.arange(0, total_k + seqlen, seqlen, dtype=torch.int32, device=device)
+
+    # Mem I/O: KV read + Q/QV read + O write
+    total_seqlen = seqlen * batch_size
+    mem_io = (
+        total_seqlen * nheads_kv * (headdim + headdim_v) * 2  # K + V read
+        + q.numel() * 2 + qv.numel() * 2  # Q + QV read
+        + total_q * nheads_q * headdim_v * 2  # O write
+    )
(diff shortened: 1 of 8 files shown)


flash_attn/cute/flash_fwd_sm90.py
```diff
@@ -557,7 +557,9 @@ def kernel(
                 blocksparse_tensors.cu_total_m_blocks if blocksparse_tensors is not None else None
             ),
             mCuBlockIdxOffsets=(
-                blocksparse_tensors.cu_block_idx_offsets if blocksparse_tensors is not None else None
+                blocksparse_tensors.cu_block_idx_offsets
```
jayhshah, line 560: This change is just to fix the linter error.

jayhshah:
Note that 64k test in benchmark script throws IMA on cutlass 4.5.0 in the refactor, but this is due to issue identified here: https://github.com/NVIDIA/cutlass/issues/3208

jayhshah:
Benchmark shows ~15% dropoff for paged using cp async loads.

[benchmark_mla_paged.txt](https://github.com/user-attachments/files/27683939/benchmark_mla_paged.txt)

flash_attn/cute/interface.py
```diff
@@ -419,25 +433,33 @@ def _flash_attn_fwd(
...
     qhead_per_kvhead = num_head // num_head_kv
     if pack_gqa is None:
         pack_gqa = qhead_per_kvhead > 1
 
-    is_fp8 = q.dtype in (torch.float8_e4m3fn, torch.float8_e5m2)
-    if is_fp8 and (q.requires_grad or k.requires_grad or v.requires_grad):
+    is_fp8 = v.dtype in (torch.float8_e4m3fn, torch.float8_e5m2)
+    requires_grad = any(t is not None and t.requires_grad for t in [q, k, v, qv])
+    if is_fp8 and requires_grad:
         raise NotImplementedError("FA4 CuTe FP8 backward is not supported yet (forward-only).")
```
Johnsonms, line 458: Suggest adding a small interim guard here. The current MLA `qv` path does not support backward yet:

* `FlashAttnFunc.forward` does not `save_for_backward(qv)`
* `_flash_attn_bwd` has no `qv` parameter
* `q.shape[:2]` at line 1384 raises when `q is None` (MLA absorbed path)

The most dangerous case is when `q`, `k`, `v`, and `qv` all require grad: backward runs without crashing but silently computes incorrect gradients as if `qv` were not used in forward.

A lightweight guard converts this into a clear error and also avoids the deep `AttributeError` in the `q=None` path. Since both `FlashAttnFunc` and `FlashAttnVarlenFunc` funnel through `_flash_attn_fwd`, one guard here covers both.

```python
if is_fp8 and requires_grad:
    raise NotImplementedError(
        "FA4 CuTe FP8 backward is not supported yet (forward-only)."
    )

if qv is not None and requires_grad:
    raise NotImplementedError(
        "FA4 CuTe MLA `qv` backward is not implemented yet. "
        "Use `.detach()` on inputs or run under `torch.no_grad()`."
    )
```

Happy to follow up with full MLA backward support if that is in scope for this refactor; otherwise this seems like a cheap and safe first step.

flash_attn/cute/interface.py
```diff
@@ -419,25 +433,33 @@ def _flash_attn_fwd(
...
+    device = v.device
     q_batch_seqlen_shape = (batch_size, seqlen_q) if cu_seqlens_q is None else (total_q,)
-    lse_shape = (batch_size, num_head, seqlen_q) if cu_seqlens_q is None else (num_head, total_q)
-    requires_grad = q.requires_grad or k.requires_grad or v.requires_grad
+
+    if qv is None:
+        lse_shape = (batch_size, num_head, seqlen_q) if cu_seqlens_q is None else (num_head, total_q)
+    else:
+        # num_head contiguous better for MQA in MLA absorbed
+        lse_shape = (batch_size, seqlen_q, num_head) if cu_seqlens_q is None else (total_q, num_head)
```
Johnsonms, line 467: The returned `lse` layout depends on `qv`:
- `qv is None`: `(batch, nheads, seqlen_q)` — existing public contract
- `qv is not None`: `(batch, seqlen_q, nheads)` — transposed

Silent contract change for `return_lse=True` callers: not in the public docstrings, and existing tests only check `lse` for non-NaN, so a transposed return goes unnoticed.

Either transpose back at the public API boundary, or document the divergence and add a shape assertion on both paths.

benchmarks/benchmark_mla_paged_kv.py
```diff
@@ -0,0 +1,108 @@
...
+dtype = torch.bfloat16
+seqlen_q = 1
+nheads_q = 128
+nheads_kv = 1  # MQA-128
+headdim = 64
+headdim_v = 512
+causal = True
+
+batch_size = 128
+page_sizes = [None, 16, 64, 128]  # None = non-paged baseline
```
Johnsonms, line 27: benchmarks/benchmark_mla_paged_kv.py crashes with cudaErrorIllegalAddress  on page_size=16 (cp.async) at seqlen=65536 (total_pages = 128 × 4096 = 524288, V-side indexable ~8 GB). TMA (page_size=128) is clean at the same seqlen — likely 32-bit overflow in byte-offset arithmetic on the cp.async pointer path (cpasync_paged_load_KV @ flash_fwd_mla_sm100.py:1712, compute_X_ptr @ paged_kv.py:158).

Johnsonms (approved):

flash_attn/cute/interface.py
jayhshah: I'll be adding backward in a follow-up PR.

benchmarks/benchmark_mla_paged_kv.py
jayhshah: This should be fixed in cutlass 4.5.2

flash_attn/cute/interface.py
jayhshah: Now documented in docstring for varlen func

PR #2559, opened 2026-05-13
[CuTe,Sm100] Varlen Dynamic Persistent scheduler and metadata

reubenconducts:
This PR adds the `VarlenDynamicPersistentScheduler`, seen in FA3, to FA4.
- Accepts scheduler metadata tensors, which can be prepared in the `prepare_scheduler_metadata.py` `FlashPrepareScheduler` kernel:
	- `num_m_blocks_ptr` to precompute number of m blocks for the lpt sort (unused on SM100)
	- `num_splits_dynamic_ptr`: holds `num_splits` for each sequence in a batch, which can vary in mixed workloads
	- `virtual_batch_idx_ptr`: unused currently; used to permute sequences according to "virtual" batch indices, e.g. when sorting a batch for load balancing (to be utilized in a subsequent PR)
	- `num_nheads_in_l2_ptr`: used for head swizzle computation in the tile scheduler
	- `tile_count_semaphore`: zeroed out here, used in main kernel for dynamic persistence
- Threads metadata through interface with options to disable, call `FlashPrepareKernel`, or pass in pre-computed cached metadata to amortize cost across layers.
- Refactors `tile_scheduler.py` so that dynamic persistent schedulers (CLC and "traditional", like that in this PR) share methods and so that varlen schedulers (`SingleTile` and `DynamicPersistent`) reuse common methods.

Additionally, adds in a `mCuTotalMBlocks` tensor to be used by the `SingleTileVarlenScheduler` to perform a binary search for the current batch. This is due to @horakka5 in #2520, with an additional `mCuTotalSplitsMBlocks` pointer used jointly to deduce dynamic num splits, when appropriate.

Comprehensive performance numbers are attached. The regressions seen (e.g. with large batch, small seqlen) are attributable mainly to the fact that the combine kernel does not early-exit for 1-split sequences. We see that "traditional" dynamic persistent outperforms CLC (to which I've also wired up the appropriate metadata tensors) almost always. It's worth noting that in the few tests where prepare kernel latency appears extreme, a large proportion of that latency comes from torch.empty (known issue on Grace CPUs; ostensibly fixed with cuda 13.x, but I ran on 12.x). 

[varlen_dynamic_scheduler_perf.txt](https://github.com/user-attachments/files/27735757/varlen_dynamic_scheduler_perf.txt)

To-do in follow-up PRs:
- Currently the combine kernel does not early-exit for 1-split sequences; this explodes in many cases from over-launching zero-work CTAs. Solution is to use a persistent scheduler. 
- Implement lpt batch sort in the `FlashPrepareKernel`. This is proven to have an enormous impact on mixed prefill/decode workloads (including with CLC, from my preliminary testing). (In addition to this varlen case, batch sort will be helpful for load balancing with blocksparsity.)


flash_attn/cute/flash_fwd.py
```diff
@@ -637,6 +637,8 @@ def __call__(
         learnable_sink: Optional[cute.Tensor] = None,
         blocksparse_tensors: Optional[BlockSparseTensors] = None,
         aux_tensors=None,
+        mCuTotalMBlocks: Optional[cute.Tensor] = None,
```
drisspg, line 641: nit: can we turn this into a namedTuple for keepign them colocated?

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -676,8 +712,15 @@ def __call__(
             cutlass.max(cute.cosize(sQ_layout), cute.cosize(sO_layout) * self.o_dtype.width // self.q_dtype.width)
         )
 
-        clc_response_size = self.sched_stages * 4 if self.use_clc_scheduler else 0
-        clc_mbar_size = self.sched_stages * 2 if self.use_clc_scheduler else 0
+        sched_response_size = (
+            self.sched_stages * 4
```
drisspg: what bout if is_persistent?

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -698,12 +742,12 @@ class SharedStorage:
...
             sScale: cute.struct.MemRange[Float32, self.q_stage * self.m_block_size * 2]
-            # CLC buffers placed here to utilize padding before sO's 1024-byte alignment.
-            # This avoids adding bytes at the end when we're at the smem limit.
-            # PipelineClcFetchAsync expects 2 * sched_stages mbarriers (full + empty).
-            clc_mbar_ptr: cute.struct.MemRange[cutlass.Int64, clc_mbar_size]
-            # CLC response storage (16 bytes per stage, stored as 4 Int32s).
-            clc_response: cute.struct.MemRange[Int32, clc_response_size]
+            # Scheduler buffers placed here to utilize padding before sO's 1024-byte
+            # alignment. PipelineClcFetchAsync / PipelineAsync both expect
+            # 2 * sched_stages mbarriers (full + empty). Response is 4 Int32 per stage
```
drisspg, line 761: can we keep  the note though that if we were to move this to end it would fail due to fragmentation of smem

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1014,6 +1068,22 @@ def kernel(
                 defer_sync=True,
             )
 
+        pipeline_load_epi = None
+        if const_expr(self.overlap_sO_sQ and self.is_persistent):
```
drisspg, line 1086: maybe a small note here

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1087,60 +1160,79 @@ def kernel(
...
-            clc_pipeline_producer_group = cutlass_pipeline.CooperativeGroup(
+        sched_ctx = None
+        if const_expr(self.use_clc_scheduler or self.dynamic_persistent):
+            sched_response_ptr = storage.sched_response.data_ptr()
+            sched_mbar_ptr = storage.sched_mbar_ptr.data_ptr()
+            sched_producer_group = cutlass_pipeline.CooperativeGroup(
                 cutlass_pipeline.Agent.Thread
             )
-            num_clc_consumer_warps_per_cta = self.threads_per_cta // cute.arch.WARP_SIZE
-            # NB on CTA0 warp15 == scheduler on CTA1 == empty but still both consume
```
drisspg, line 1109: can we leave this note unless that has changed?

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -2928,30 +3064,37 @@ def epilogue_s2g(
...
-                    "[CLC] query sm={} cta={} (m_blk={},h={},b={},s={}) valid={}\n",
-                    smid(),
-                    cute.arch.block_idx()[0],
-                    work_tile.tile_idx[0],
-                    work_tile.tile_idx[1],
-                    work_tile.tile_idx[2],
-                    work_tile.tile_idx[3],
-                    work_tile.is_valid_tile,
-                )
+            if const_expr(self.use_clc_scheduler):
```
drisspg: should we make this log support both dynamic schedulers?I do find this very helpful for debugging

flash_attn/cute/interface.py
```diff
@@ -1,6 +1,7 @@
 # Copyright (c) 2025, Jay Shah, Ganesh Bikshandi, Ying Zhang, Vijay Thakkar, Pradeep Ramani, Tri Dao.
 # [2025-07-04] Version in Cute-DSL, for Hopper and Blackwell. You'll need install nvidia-cutlass-dsl==4.2.0.
 
+from numpy._core.defchararray import zfill
```
drisspg: hmmmm

flash_attn/cute/interface.py
```diff
@@ -2554,6 +2909,345 @@ def flash_attn_combine(
         lse,
         cu_seqlens,
         seqused,
-        varlen_batch_idx=varlen_batch_idx,
+        virtual_batch_idx=virtual_batch_idx,
     )
     return out, lse
+
+
+def _get_scheduler_metadata(
```
drisspg, line 3621: you think some of this should live inteh scheduler file?

flash_attn/cute/tile_scheduler.py
```diff
@@ -38,35 +38,59 @@ class SchedulingMode(IntEnum):
...
+    def advance_consumer_state(self, *, loc=None, ip=None):
+        self._consumer_state.advance(loc=loc, ip=ip)
+
+    def producer_tail(self, *, loc=None, ip=None):
+        self._pipeline.producer_tail(self._producer_state, loc=loc, ip=ip)
+
+
+@dataclass
+class ClcSchedulerState(SchedulerState):
+    """CLC-backed: `prefetch_next_work` issues the HW query."""
```
drisspg: can we leave the full doc :)

flash_attn/cute/tile_scheduler.py
```diff
@@ -768,23 +833,305 @@ def __new_from_mlir_values__(self, values):
...
+                                if num_n_blocks * 8 <= self.max_kvblock_in_l2
+                                else (
+                                    4
+                                    if num_n_blocks * 4 <= self.max_kvblock_in_l2
+                                    else (2 if num_n_blocks * 2 <= self.max_kvblock_in_l2 else 1)
+                                )
+                            )
+                        )
+                        nheads_in_l2 = min(nheads_in_l2, self.num_head)
+                    mh_in_l2 = nheads_in_l2 * num_m_blocks
```
drisspg, line 1151: how different is this swizziling (havent read) should it be shared between LPT or at least parts?

flash_attn/cute/flash_fwd_sm100.py
reubenconducts: `is_persistent` means `is_static_persistent or dynamic_persistent` and dynamic persistent includes clc (so this conditional is redundant; it should just be `if self.dynamic_persistent`)

flash_attn/cute/flash_fwd.py
reubenconducts: It's a bit tricky. This is what `SchedulerMetadataTensorsTorch` is for, but this PR doesn't add general scheduler metadata to the non-sm100 kernels.

flash_attn/cute/interface.py
reubenconducts: I think it's best left in interface, for example to hint that tile size selection needs to agree.

flash_attn/cute/tile_scheduler.py
```diff
@@ -816,43 +1175,36 @@ def create(
...
-                max_kvblock_in_l2=max_kvblock_in_l2,
-                tile_shape_mn=args.tile_shape_mn,
-                mCuSeqlensQ=args.mCuSeqlensQ,
-                mSeqUsedQ=args.mSeqUsedQ,
-                qhead_per_kvhead_packgqa=args.qhead_per_kvhead_packgqa,
-                lpt=args.lpt,
-                is_split_kv=args.is_split_kv,
+            # TODO: Support varlen CLC with cluster_shape_m > 1 by refactoring the
+            # flattened-tile decode so cluster unpacking semantics are explicit.
+            assert scheduling_mode != SchedulingMode.CLC or args.cluster_shape_mn[0] == 1, (
```
jayhshah: We have addressed this TODO on main; reintroducing the assert would break MLA absorbed varlen

flash_attn/cute/tile_scheduler.py
```diff
@@ -1110,13 +1360,198 @@ def __new_from_mlir_values__(self, values):
...
+            assert args.mCuSeqlensQ is not None or args.mSeqUsedQ is not None, (
+                "At least one of mCuSeqlensQ or mSeqUsedQ must be provided"
+            )
+            decoder = VarlenDecoder.create(
+                args,
+                fold_splits_into_scan=True,
+                scheduling_mode=SchedulingMode.DYNAMIC,
+                loc=loc,
+                ip=ip,
+            )
```
jayhshah, line 1417: as written dynamic w/ semaphore doesn't support non-trivial cluster. We should add an assert for this and can update in a follow-on PR

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -289,7 +308,7 @@ def __init__(
             self.empty_warp_ids = self.empty_warp_ids + self.epilogue_warp_ids
             self.epilogue_warp_ids = self.correction_warp_ids
 
-        self.clc_scheduler_warp_id = self.empty_warp_ids[0] if self.use_clc_scheduler else None
+        self.scheduler_warp_id = self.empty_warp_ids[0] if self.dynamic_persistent else None
```
jayhshah, line 326: We should assert that empty_warp_ids is non-empty for dynamic persistent to ensure we indeed have a warp to assign as a scheduler warp.

There's a subtle dependency here: we write that "CLC does not compose" with use_tma_KV = false to avoid the problem of warp 15 being repurposed as a load warp and therefore being unavailable for use as the scheduler warp. However, for varlen dynamic persistent we always have use_correction_warps_for_epi = True, freeing up warp 13 for use as the scheduler warp.

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1089,60 +1160,80 @@ def kernel(
...
             tile_scheduler = self.tile_scheduler_cls.create(tile_sched_params)
         assert isinstance(tile_scheduler, TileSchedulerProtocol), f"tile_scheduler is not a TileSchedulerProtocol: {type(tile_scheduler)}"
 
         # ///////////////////////////////////////////////////////////////////////////////
-        #  EMPTY / CLC SCHEDULER WARP
+        #  EMPTY / SCHEDULER WARP
         # ///////////////////////////////////////////////////////////////////////////////
-        if const_expr(self.use_clc_scheduler):
-            if warp_idx == self.clc_scheduler_warp_id:
+        if const_expr(self.use_clc_scheduler or self.dynamic_persistent):
```
jayhshah: The dynamic_persistent boolean subsumes use_clc_scheduler, so this conditional should only be on dynamic_persistent.

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -1089,60 +1160,80 @@ def kernel(
...
+            if warp_idx == self.scheduler_warp_id:
                 cute.arch.setmaxregister_decrease(self.num_regs_other)
-                if is_leader_cta:
-                    self.clc_scheduler_warp(tile_scheduler)
+                # CLC: only leader CTA produces.
+                if const_expr(self.dynamic_persistent) or is_leader_cta:
```
jayhshah: Should just be is_leader_cta; see preceding comment.

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -2931,30 +3066,38 @@ def epilogue_s2g(
...
-                    "[CLC] query sm={} cta={} (m_blk={},h={},b={},s={}) valid={}\n",
-                    smid(),
-                    cute.arch.block_idx()[0],
-                    work_tile.tile_idx[0],
-                    work_tile.tile_idx[1],
-                    work_tile.tile_idx[2],
-                    work_tile.tile_idx[3],
-                    work_tile.is_valid_tile,
-                )
+            if const_expr(self.dynamic_persistent):
```
jayhshah: This conditional is superfluous

flash_attn/cute/block_info.py
```diff
@@ -19,13 +19,17 @@ class BlockInfo:
...
+    num_splits_dynamic_ptr: Optional[cute.Tensor] = None
+    num_n_blocks_per_split: Optional[cutlass.Constexpr[Int32]] = None
 
     @cute.jit
     def get_n_block_min_max(
         self,
         seqlen_info: SeqlenInfoQK,
         m_block: Int32,
         split_idx: Int32 = 0,
+        batch_idx: Int32 = 0,
```
jayhshah: This added argument is unused and can be removed.

Correspondingly you'll have to update the call sites but it should simplify the merge since the signature is now unchanged.

flash_attn/cute/interface.py
```diff
@@ -1097,7 +1417,12 @@ def _flash_attn_fwd(
             lse.transpose(-1, -2) if lse is not None else None,
             cu_seqlens_q,
             seqused_q,
+            num_splits_dynamic_ptr=num_splits_dynamic if has_scheduler_metadata else None,
+            virtual_batch_idx=virtual_batch_idx if has_scheduler_metadata else None,
         )
+    if reuse_scheduler_metadata and tile_count_semaphore is not None:
+        # combine kernel does this for us
```
jayhshah: Can you write this as a TODO or otherwise pass the tile_count_semaphore to the combine kernel and only zero out when is_split_kv=False?

If we pass the semaphore to the combine kernel, we may need to have CTA 0 zero out the semaphore instead of the last CTA to ensure the CTA isn't one that is prematurely exited.

flash_attn/cute/interface.py
```diff
@@ -717,6 +882,110 @@ def _flash_attn_fwd(
...
+        num_splits_dynamic = None
+        virtual_batch_idx = None
+        num_nheads_in_l2 = None
+        tile_count_semaphore = None
+
+    # use binary batch search in SingleTileVarlenScheduler to avoid
+    # O(N^2) lookup; observed to be faster only for batch_size > BIN_BATCH_SEARCH_THRESH; this is tunable
+    cu_total_m_blocks = None
+    cu_total_splits_m_blocks = None
+    use_single_tile_varlen_scheduler = use_clc_scheduler or tile_count_semaphore is None
```
jayhshah: Why is this keyed to use_clc_scheduler? It should just be tile_count_semaphore is None since we want to overwrite clc with dynamic persistent if the semaphore is present as determined by above settings.

flash_attn/cute/interface.py
```diff
@@ -817,10 +1093,29 @@ def _flash_attn_fwd(
...
+                num_splits_dynamic,
+                tile_count_semaphore,
+                virtual_batch_idx,
+                num_nheads_in_l2,
+                cu_total_m_blocks,
+                cu_total_splits_m_blocks,
+            )
+        ]
+
+        qv_tensor = to_cute_tensor(qv) if qv is not None else None
```
jayhshah: You don't have to restore the "if x is not None else None" pattern since to_cute_tensor now returns None if passed None.

flash_attn/cute/block_info.py
```diff
@@ -45,11 +49,20 @@ def get_n_block_min_max(
             n_idx_left = n_idx - self.window_size_left
             n_block_min = cutlass.max(n_idx_left // self.tile_n, 0)
         if cutlass.const_expr(self.is_split_kv):
-            num_n_blocks_per_split = (
-                Int32(0)
-                if n_block_max <= n_block_min
-                else (n_block_max - n_block_min + num_splits - 1) // num_splits
-            )
+            if const_expr(self.num_splits_dynamic_ptr is not None):
```
jayhshah: Since we're not using num_splits_dynamic_ptr as a pointer we could rename this to be more reflective of its role, e.g. "pack_split_idx" boolean

jayhshah:
LGTM now though the varlen performance matrix has now gotten very complex in terms of scheduler dispatch and the multitude of heuristics. We will have to do a performance pass in a follow-up PR.

jayhshah (approved):

PR #2560, opened 2026-05-13
Normalize scalar tensors in CuTe JIT cache keys

xyzhang626:
CuTeDSL compile keys are compared and hashed on the Python side. If a scalar tensor gets into the key, two keys with the same scalar value but produced by different layers compare by object identity and miss both the in-memory and persistent JIT caches.

This can happen in FA4 varlen backward when `max_seqlen_q/k` are scalar tensors and the backward compile key includes derived single-block predicates such as `seqlen_q_rounded // m_block_size == 1`. Each layer can produce a fresh scalar tensor, causing repeated backward JIT compilation for the same kernel shape.

This PR normalizes scalar tensors in CuTe JIT cache keys before in-memory lookup/storage and before computing the persistent cache hash.

Tests added:
- in-memory cache lookup with distinct scalar tensor objects but identical values
- persistent cache hash stability for scalar tensor keys

Local validation:
- manual CPU scalar cache normalization check passed
- manual CUDA scalar cache normalization check passed on GPU7

Could not run pytest locally because pytest is not installed in the available environment.

### DIFF
--- a/flash_attn/cute/cache_utils.py
+++ b/flash_attn/cute/cache_utils.py
@@ -16,6 +16,7 @@ import ctypes
 import cutlass
 import cutlass.cute as cute
 import tvm_ffi
+import torch
 from cutlass.cutlass_dsl import JitCompiledFunction
 from flash_attn.cute.fa_logging import fa_log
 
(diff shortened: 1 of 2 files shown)


xyzhang626:
Just realized this some overlaps with [#2507](https://github.com/Dao-AILab/flash-attention/pull/2507). #2507 fixes the specific bwd max_seqlen source, while this PR normalizes scalar tensors at the JIT cache boundary so future scalar tensor compile-key entries do not miss the in-memory or persistent cache.

PR #2568, opened 2026-05-15
[Cute,Sm100] allow for zero length sequences in hdim 256 kernels

jayhshah:
We add support for zero-length Q and KV sequences for varlen mode in the sm100 hdim 256 kernels. Changes are as follows:

Forward: have softmax execute one dummy iteration to not hang for zero-length K. We also add row sum check for zero or NaN to not output NaN in this case.

Backward dQ: for guarding work tile per-iteration, check also that trip count is non-zero. If it is zero, write zero as the output.

Backward dKV: zero-length Q and K was nominally supported but the write zero logic was broken and yielded IMA; now fixed in the PR.

### DIFF
--- a/flash_attn/cute/sm100_hd256_2cta_fmha_backward.py
+++ b/flash_attn/cute/sm100_hd256_2cta_fmha_backward.py
@@ -21,6 +21,7 @@ from flash_attn.cute.sm100_hd256_2cta_fmha_backward_dqkernel import (
 from flash_attn.cute.sm100_hd256_2cta_fmha_backward_dkdvkernel import (
     BlackwellFusedMultiHeadAttentionBackwardDKDVKernel,
 )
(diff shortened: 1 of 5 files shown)


Johnsonms (approved):

Johnsonms:
PR fixes the bug.

  Before patch (b11 baseline):
  - forward OK → out.backward(g) → cudaErrorIllegalAddress in _flash_attn_bwd (sm100 hd=256 kernel)
  
  After patch (jshah/hdim256-varlen-zero-lengths, commit 75db52f):
  - forward OK
  - backward OK
  - grads finite, non-zero, and structurally correct — dk[0:2538] (the K rows paired with the zero-length Q segment) is
   exactly zero as expected, while dk[2538:] carries signal.

PR #2573, opened 2026-05-19
Fix FA2 varlen SplitKV early-exit LSE offset

DCwashing:
## Summary

Fix the LSE write offset used by the FA2 SplitKV early-exit path when writing directly to unpadded `softmax_lse`.

Upstream varlen forward stores LSE in the packed unpadded layout by setting `params.unpadded_lse = true`. The normal epilogue path already handles this layout, but the SplitKV early-exit path still uses the padded `(batch, head, seqlen_q)` offset.

## Problem

When running a Qwen 235B model on a single node with 8x L40S GPUs after enabling DCP, execution can hang and eventually report:

`torch.AcceleratorError: CUDA error: an illegal memory access was encountered`

The issue is in the FA2 SplitKV early-exit path. It always computes `row_offset_lseaccum` with the padded SplitKV-style layout:

`((n_split_idx * b + bidb) * h + bidh) * seqlen_q + m_block * kBlockM`

That layout is correct for SplitKV/LSE accumulation buffers, but not for direct unpadded LSE output in varlen mode.

For varlen forward, `softmax_lse` is allocated as `{num_heads, total_q}` and `params.unpadded_lse` is set to `true`. In this case, the early-exit path must use the same packed varlen LSE layout as the normal epilogue path.

## Fix

Use the packed varlen LSE offset when writing directly to unpadded `softmax_lse`:

`bidh * params.total_q + binfo.q_offset(params.seqlen_q, 1, bidb) + m_block * kBlockM`

Keep the existing padded offset for SplitKV accumulation buffers and padded LSE output.

### DIFF
--- a/csrc/flash_attn/src/flash_fwd_kernel.h
+++ b/csrc/flash_attn/src/flash_fwd_kernel.h
@@ -542,7 +542,10 @@ inline __device__ void compute_attn_1rowblock_splitkv(const Params &params, cons
             + m_block * kBlockM * params.o_row_stride + bidh * params.o_head_stride;
         const index_t row_offset_oaccum = (((n_split_idx * params.b + bidb) * params.h + bidh) * params.seqlen_q
             + m_block * kBlockM) * params.d_rounded;
-        const index_t row_offset_lseaccum = ((n_split_idx * params.b + bidb) * params.h + bidh) * params.seqlen_q + m_block * kBlockM;
+        const index_t row_offset_lseaccum = (Split || !params.unpadded_lse
+            ? ((n_split_idx * params.b + bidb) * params.h + bidh) * params.seqlen_q
+            : bidh * params.total_q + binfo.q_offset(params.seqlen_q, 1, bidb)
+        ) + m_block * kBlockM;
         Tensor gOaccum = make_tensor(make_gmem_ptr(reinterpret_cast<ElementO *>(Split ? params.oaccum_ptr : params.o_ptr) + (Split ? row_offset_oaccum : row_offset_o)),
                                       Shape<Int<kBlockM>, Int<kHeadDim>>{},
                                      make_stride(Split ? kHeadDim : params.o_row_stride, _1{}));


jagane-memwize:
We hit this same bug independently and can confirm both the root cause and that this exact fix resolves it, but with a different, fully deterministic symptom that may help review: instead of an illegal memory access, we saw silent LSE corruption, where empty sequences splatter their +inf early-exit marker over other sequences' LSE entries.
Observed behavior:
  - With no seqused_k == 0 rows in the batch, every LSE is correct.
  - Each row e with seqused_k[e] == 0 writes its early-exit +inf LSE marker to the wrong addresses: it lands on the LSE entries of ~`num_heads_k*other* sequences (near columne * num_heads_k), on the first GQA group's q-heads. With num_heads_k = 1` the write happens to land at the correct column - the mis-addressing scales with the kv-head count, exactly as the padded-vs-unpadded offset mismatch predicts.
  - out is correct everywhere (the early-exit's O write uses the varlen-aware binfo.q_offset addressing — only the LSE offset is wrong).

Why an IMA in the OP's case vs. silent corruption in ours: the wrong offset is computed against softmax_lse_ptr's padded (b, h, seqlen_q) extent instead of the actual unpadded (h, total_q) allocation. Depending on shapes it either stays inside the allocation (corrupting neighbors' LSE, our case) or runs past it (OP's CUDA illegal memory access). Same bug, two manifestations.

Validation of this fix: we applied the identical change (making the early-exit row_offset_lseaccum mirror the normal epilogue's Split || !params.unpadded_lse conditional) and rebuilt:
- Isolated repro: an empty row at index e now writes +inf at exactly column e, all heads, nowhere else. A B=1536 production-shaped batch went from 192 corrupted LSE rows to 0; out unchanged and still correct.
- End-to-end: our serving stack computes an exact reference LSE in torch and counts disagreements with the kernel's LSE. Against the stock kernel this counted 50K–460K corrupted rows per run; with this patch it reads zero across all instances - two independent implementations now agree everywhere.

One nuance for reviewers: strictly, the seqlenq_ngroups_swapped case also needs the swapped stride order (the full get_lse_tile treatment) for the early-exit write to land at its own exact position. In practice the patched offset was sufficient in all our testing — every +inf landed on the correct sequence — but it may be worth a follow-up to unify the early exit on get_lse_tile like the non-splitkv kernel already does (lines ~109, ~462).

I am uploading a single python file test script for validation. I'm also happy to share our test environment details.

Bug was found and test script generated using Claude Code.

jagane-memwize:
[fa2_lse_earlyexit_repro.py](https://github.com/user-attachments/files/30804474/fa2_lse_earlyexit_repro.py)

Failed run:
===================================================================================

(venv-vllm) root@ubuntu:~/working/vllm-memwize-attention# FA2_SO=/root/working/vllm/vllm/vllm_flash_attn/_vllm_fa2_C.abi3.so.stock python prototype/fa2_lse_earlyexit_repro.py
kernel: /root/working/vllm/vllm/vllm_flash_attn/_vllm_fa2_C.abi3.so.stock  md5=0f949d139cff7433ae06c1559ef5b35f
batch B=64 heads H=8 kv-heads HK=2, empty rows [1, 10]
FAIL: 24 corrupted LSE entries in 4 victim columns
  seq   1 (seqused_k=  0) heads [0, 1, 2, 3, 4, 5, 6, 7]: lse=[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0] expected=[inf, inf, inf, inf, inf, inf, inf, inf]
  seq   3 (seqused_k= 61) heads [0, 1, 2, 3]: lse=[inf, inf, inf, inf] expected=[4.487, 4.896, 5.105, 4.582]
  seq  10 (seqused_k=  0) heads [0, 1, 2, 3, 4, 5, 6, 7]: lse=[0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0] expected=[inf, inf, inf, inf, inf, inf, inf, inf]
  seq  21 (seqused_k= 50) heads [0, 1, 2, 3]: lse=[inf, inf, inf, inf] expected=[4.643, 4.295, 4.112, 4.231]
note: expected splatter near columns [2, 20] (= empty_row * num_heads_k), first H/HK=4 q-heads

Successful run, with patched code:
===================================================================================
(venv-vllm) root@ubuntu:~/working/vllm-memwize-attention# FA2_SO=/root/working/vllm/vllm/vllm_flash_attn/_vllm_fa2_C.abi3.so python prototype/fa2_lse_earlyexit_repro.py
kernel: /root/working/vllm/vllm/vllm_flash_attn/_vllm_fa2_C.abi3.so  md5=fd4689150b2697a7320b01b1611f74d5
batch B=64 heads H=8 kv-heads HK=2, empty rows [1, 10]
PASS: all LSE entries correct (empty rows inf in their own column only)

csrc/flash_attn/src/flash_fwd_kernel.h
```diff
@@ -542,7 +542,10 @@ inline __device__ void compute_attn_1rowblock_splitkv(const Params &params, cons
             + m_block * kBlockM * params.o_row_stride + bidh * params.o_head_stride;
         const index_t row_offset_oaccum = (((n_split_idx * params.b + bidb) * params.h + bidh) * params.seqlen_q
             + m_block * kBlockM) * params.d_rounded;
-        const index_t row_offset_lseaccum = ((n_split_idx * params.b + bidb) * params.h + bidh) * params.seqlen_q + m_block * kBlockM;
+        const index_t row_offset_lseaccum = (Split || !params.unpadded_lse
```
bvolpato, line 545: Could `seqlenq_ngroups_swapped` need its distinct LSE stride here? `get_lse_tile` writes that path as `(h, seqlen_q, b)`, but this offset appears to use `(h, b, seqlen_q)`. For an empty K row, could this place `+inf` in another sequence’s LSE? Is this a concern, and would reusing `get_lse_tile` here cover both layouts?

bvolpato (commented):
Other than this comment, this looks good to me.

jagane-memwize:
Your point is well taken, @bvolpato - I have reworked the PR by reusing get_lse_tile. I also enhanced the test to cover this case. I created the PR against the OP's fork/branch, i.e. repo DCwashing/flash-attention branch fix-fa2-varlen-splitkv-early-exit-lse - LMK if you'd like me to create a PR against main on this repo

jagane-memwize:
@bvolpato - I'm uploading a patch against the latest source code, it's easier to read. 
[fix-fa2-varlen-splitkv-early-exit-using-get-lse-tile.patch](https://github.com/user-attachments/files/31106935/fix-fa2-varlen-splitkv-early-exit-using-get-lse-tile.patch)

jagane-memwize:
I have created a new PR#2851 which is cleaner. It includes a fix to get_lse_tile. @bvolpato  - I would appreciate it if you could take a look at that PR.

PR #2589, opened 2026-05-25
Use is_family_of for sm_90 and sm_103 arch checks

Johnsonms:
Follow-up to #2572 — apply the same is_family_of pattern to the two remaining range-style arch checks for consistency:

- flash_fwd_sm90.py:69 (SM 9.x assert)
- flash_fwd_sm100.py:195 (is_sm103 flag)

Same semantic narrowing as #2572: bare-base SMs (`sm_90`, `sm_103`) are excluded — they were accepted by the old range check but  cannot actually run these kernels, since wgmma / UMMA / 2CTA paths require the `a`/`f` PTX variant. The new assertion fails earlier with a clearer error instead of letting compilation get to ptxas before failing.

### DIFF
--- a/flash_attn/cute/flash_fwd_sm100.py
+++ b/flash_attn/cute/flash_fwd_sm100.py
@@ -192,7 +192,7 @@ class FlashAttentionForwardSm100:
         self.mask_vec_size: cutlass.Constexpr = getattr(mask_mod, "__vec_size__", 1)
         # Does S1 need to wait for S0 to finish
         # self.s0_s1_barrier = self.head_dim_padded in [64, 96] and (not self.is_causal and not self.is_local)
-        is_sm103 = self.arch >= Arch.sm_103 and self.arch <= Arch.sm_103f
+        is_sm103 = self.arch.is_family_of(Arch.sm_103f)
         self.is_sm103 = is_sm103
         # enable_ex2_emu is derived: True if tuning config has freq > 0, else fallback to default logic
         _default_enable_ex2_emu = (self.head_dim_padded <= 128 or (self.head_dim_padded == 192 and self.use_2cta_instrs and not self.is_causal and not self.is_local)) and not is_sm103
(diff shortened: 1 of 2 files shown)


Johnsonms:
Cc, @ocss884 @lingolin128 helping review

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -192,7 +192,7 @@ def __init__(
         self.mask_vec_size: cutlass.Constexpr = getattr(mask_mod, "__vec_size__", 1)
         # Does S1 need to wait for S0 to finish
         # self.s0_s1_barrier = self.head_dim_padded in [64, 96] and (not self.is_causal and not self.is_local)
-        is_sm103 = self.arch >= Arch.sm_103 and self.arch <= Arch.sm_103f
+        is_sm103 = self.arch.is_family_of(Arch.sm_103f)
```
ocss884, line 199: is_family_of will test `(self.major == arch.major and self.minor >= arch.minor)`, so it is correct for now. If there is another sm10.x hardware with `x > 3` in the future then it would be wrong.

ocss884 (approved):

lingolin128:
LGTM, @ocss884's consideration is completely correct in terms of code.
btw, It's unlikely there will be `10.x` hardware with `x > 3`.

flash_attn/cute/flash_fwd_sm100.py
Johnsonms: @ocss884 @lingolin128 Good catch — added an inline comment in 5833281 to make the forward-inclusive semantics explicit. The flag gates ex2 emulation, and since sm_103 (B300) brought fast hardware ex2, any future sm_10x Blackwellvariant should inherit the same behavior. So `is_family_of`'s `minor >= 3` behavior is actually what we want here, even though the  literal name `is_sm103` is a bit misleading.

jayhshah (approved):

PR #2594, opened 2026-05-26
Clamp kv_stage to avoid SMEM overflow for small head_dims on SM100

### LINKED ISSUE #2591: [FA4, CuTe, SM100] `_flash_attn_fwd` fails at launch for `head_dim < 32` on B200
On B200 (SM100), the FA4 cute forward kernel raises `RuntimeError: CUDA Error: cudaErrorInvalidValue` at kernel launch when `head_dim < 32`, even though `_validate_head_dims` advertises `8 <= head_dim <= 128` as supported on SM100.

## Script to reproduce the error

Pinned versions:

```bash
uv venv --python=3.12
uv pip install torch==2.12.0 "nvidia-cutlass-dsl[cu13]==4.4.2" "flash-attn-4[cu13]==4.0.0b14"
uv run python fa4_head_dim_repro.py
```

Python script:

```python
import math
import sys
import torch


def fa4_call(head_dim: int):
    batch, seqlen, num_heads = 1, 2048, 2
    dtype = torch.bfloat16
    device = "cuda"

    q = torch.randn(batch, seqlen, num_heads, head_dim, dtype=dtype, device=device)
    k = torch.randn(batch, seqlen, num_heads, head_dim, dtype=dtype, device=device)
    v = torch.randn(batch, seqlen, num_heads, head_dim, dtype=dtype, device=device)

    torch.cuda.synchronize()
    try:
        out, lse = _flash_attn_fwd(
            q,
            k,
            v,
            softmax_scale=1.0 / math.sqrt(head_dim),
            causal=True,
            return_lse=True,
        )
        torch.cuda.synchronize()
    except Exception as exc:
        return f"head_dim={head_dim:>3}: FAIL  {type(exc).__name__}: {exc}"

    return (
        f"head_dim={head_dim:>3}: OK    "
        f"out.shape={tuple(out.shape)}  lse.shape={tuple(lse.shape)}"
    )



def main():
    cap = torch.cuda.get_device_capability()
    print(f"python: {sys.version.split()[0]}")
    print(f"torch:  {torch.__version__}")
    print(
        f"device: {torch.cuda.get_device_name(0)} "
        f"sm_{cap[0]}{cap[1]}  (capability={cap})"
    )
    print("Calling flash_attn.cute._flash_attn_fwd with varying head_dim:")
    print("(batch=1, seqlen=2048, num_heads=2, bf16, causal=True)")

    results = [fa4_call(hd) for hd in (8, 16, 32, 64, 96, 128)]
    print("\n".join(results))


if __name__ == "__main__":
    main()
```

### Output

```
python: 3.12.3
torch:  2.12.0+cu130
device: NVIDIA B200 sm_100  (capability=(10, 0))
Calling flash_attn.cute._flash_attn_fwd with varying head_dim:
(batch=1, seqlen=2048, num_heads=2, bf16, causal=True)
head_dim=  8: FAIL  RuntimeError: CUDA Error: cudaErrorInvalidValue
head_dim= 16: FAIL  RuntimeError: CUDA Error: cudaErrorInvalidValue
head_dim= 32: OK    out.shape=(1, 2048, 2, 32)  lse.shape=(1, 2, 2048)
head_dim= 64: OK    out.shape=(1, 2048, 2, 64)  lse.shape=(1, 2, 2048)
head_dim= 96: OK    out.shape=(1, 2048, 2, 96)  lse.shape=(1, 2, 2048)
head_dim=128: OK    out.shape=(1, 2048, 2, 128)  lse.shape=(1, 2, 2048)
```

Johnsonms:
## Summary

Fixes #2591.

`flash_fwd_sm100.py:335` computes `kv_stage` without accounting for per-stage overhead (`mbar_*`, `sScale`, pipeline counters). For `head_dim_padded=16`, the heuristic can select 52+ stages, exceeding the `sm_100a` shared memory limit by ~1 KB and causing kernel launch failure:

```text
cudaErrorInvalidValue: launch shared memory exceeds current GPU arch sm_100a allowed
Allocated: 233472 bytes. Max: 232448 bytes.
```

This affects small head dimensions (`head_dim ∈ [8,16]`) with non-trivial sequence lengths where `q_stage=2`.

## Fix

Clamp `kv_stage` to 32.

This is sufficient to avoid the SMEM overflow and does not impact performance in practice, since attention pipelining saturates well before 32 stages. Existing `head_dim_padded >= 64` configurations are unchanged because they already use ≤24 stages.

## Test plan

* Added `test_flash_attn_small_head_dim` covering:

  * `d ∈ {8,16,32}`
  * `causal ∈ {True, False}`
  * `seqlen ∈ {128, 2048}`
* Verified original #2591 repro now passes.
* Verified on B200 (`sm_100a`): all new tests pass with max error vs Torch SDPA ≤ 0.0078.
* Confirmed no behavior change for `head_dim_padded >= 64`.

## Notes

An alternative fix would incorporate per-stage overhead directly into the heuristic:

```python id="jlwmju"
(BUDGET - smem_qo) // (smem_kv + PER_STAGE_OVERHEAD)
```

but that reduces stage count across all configurations. The clamp is a smaller, lower-risk fix for this bug.

`_validate_head_dims` remains unchanged: small head dimensions are valid on `sm_100/110`; only the staging heuristic was incorrect.

### DIFF
--- a/flash_attn/cute/flash_fwd_sm100.py
+++ b/flash_attn/cute/flash_fwd_sm100.py
@@ -336,7 +336,10 @@ class FlashAttentionForwardSm100:
         smem_size_k_per_stage = self.n_block_size * self.head_dim_padded * self.k_dtype.width // 8
         smem_size_v_per_stage = self.n_block_size * self.head_dim_v_padded * self.v_dtype.width // 8
         smem_size_kv_per_stage = max(smem_size_k_per_stage, smem_size_v_per_stage) // self.cta_group_size
-        kv_stage = (224 * 1024 - smem_size_q_o) // smem_size_kv_per_stage
+        # Cap small head_dim from over-staging: the 224*1024 budget undercounts
+        # per-stage state, so at hd_padded=16 the unbounded formula picks 52 stages
+        # and overflows the 227 KB SMEM cap. No-op for hd_padded >= 32 (max 26).
+        kv_stage = min((224 * 1024 - smem_size_q_o) // smem_size_kv_per_stage, 32)
         if self.head_dim_padded == 192 and self.head_dim_v_padded == 128 and kv_stage == 2:
             # For hdim 192,128, we can fit 3 stages if we use uneven_kv_smem
              kv_stage = 3
--- a/tests/cute/test_flash_attn.py
+++ b/tests/cute/test_flash_attn.py
@@ -438,6 +438,50 @@ def test_flash_attn_output(
             ).abs().max().item() + dv_atol
 
 
+# Regression test for #2591: SMEM overflow at small head_dims on SM100. The main
+# test_flash_attn_output skips d < 64, but _validate_head_dims accepts head_dim >= 8
+# for sm_100/110, so this path needs coverage. Trigger requires
+# seqlen_q_packgqa > tile_m to push q_stage 1->2.
+@pytest.mark.parametrize("dtype", [torch.bfloat16])
+@pytest.mark.parametrize("causal", [False, True])
+@pytest.mark.parametrize("d", [8, 16, 32])
+@pytest.mark.parametrize("seqlen_q,seqlen_k", [(128, 128), (2048, 2048)])
+@retry_on_oom
+@maybe_fake_tensor_mode(USE_FAKE_TENSOR)
+def test_flash_attn_small_head_dim(seqlen_q, seqlen_k, d, causal, dtype):
+    device = "cuda"
+    seed = 0
+    random.seed(seed)
+    torch.random.manual_seed(seed)
+    torch.cuda.empty_cache()
+    torch.cuda.synchronize()
+    batch_size = 2
+    nheads = 2
+    nheads_kv = nheads
+    dtype_ref = dtype
+    q_ref = torch.randn(
+        batch_size, seqlen_q, nheads, d, device=device, dtype=dtype_ref
+    ).requires_grad_()
+    k_ref = torch.randn(
+        batch_size, seqlen_k, nheads_kv, d, device=device, dtype=dtype_ref
+    ).requires_grad_()
+    v_ref = torch.randn(
+        batch_size, seqlen_k, nheads_kv, d, device=device, dtype=dtype_ref
+    ).requires_grad_()
+    q, k, v = [x.detach().to(dtype).requires_grad_() for x in (q_ref, k_ref, v_ref)]
+    out_ref, _ = attention_ref(q_ref, k_ref, v_ref, None, None, causal=causal)
+    out_pt, _ = attention_ref(
+        q_ref, k_ref, v_ref, None, None, causal=causal, upcast=False, reorder_ops=True
+    )
+    out, _ = flash_attn_func(q, k, v, causal=causal)
+    if is_fake_mode():
+        return
+    fwd_atol = 2 * (out_ref + 0.3 - 0.3 - out_ref).abs().max().item()
+    assert (out - out_ref).abs().max().item() <= 2 * (
+        out_pt - out_ref
+    ).abs().max().item() + fwd_atol
+
+
 # @pytest.mark.parametrize("dtype", [torch.float16, torch.bfloat16, torch.float8_e4m3fn])
 @pytest.mark.parametrize("dtype", [torch.bfloat16])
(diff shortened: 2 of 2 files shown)


Johnsonms:
Cc: @rm-wu for review

rm-wu:
LGTM! I verified again on B200 and all `head_dim` in `8, 16, 32, 64, 96, 128` pass with this PR using the repro script in #2591 :
```sh
python: 3.12.3
torch:  2.12.0+cu130
device: NVIDIA B200 sm_100  (capability=(10, 0))
Calling flash_attn.cute._flash_attn_fwd with varying head_dim:
(batch=1, seqlen=2048, num_heads=2, bf16, causal=True)
head_dim=  8: OK    out.shape=(1, 2048, 2, 8)  lse.shape=(1, 2, 2048)
head_dim= 16: OK    out.shape=(1, 2048, 2, 16)  lse.shape=(1, 2, 2048)
head_dim= 32: OK    out.shape=(1, 2048, 2, 32)  lse.shape=(1, 2, 2048)
head_dim= 64: OK    out.shape=(1, 2048, 2, 64)  lse.shape=(1, 2, 2048)
head_dim= 96: OK    out.shape=(1, 2048, 2, 96)  lse.shape=(1, 2, 2048)
head_dim=128: OK    out.shape=(1, 2048, 2, 128)  lse.shape=(1, 2, 2048)
```

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -332,7 +332,11 @@ def _setup_attributes(self):
         smem_size_k_per_stage = self.n_block_size * self.head_dim_padded * self.k_dtype.width // 8
         smem_size_v_per_stage = self.n_block_size * self.head_dim_v_padded * self.v_dtype.width // 8
         smem_size_kv_per_stage = max(smem_size_k_per_stage, smem_size_v_per_stage) // self.cta_group_size
-        kv_stage = (224 * 1024 - smem_size_q_o) // smem_size_kv_per_stage
+        # The 224*1024 budget excludes per-stage state (mbarriers, sScale) and so undercounts
+        # actual SMEM usage. At hd_padded=16 the unbounded formula yields 52+ stages and
+        # overflows the sm_100a 227 KB cap. Clamp to an upper bound safely past the
+        # pipelining-saturation point; attention pipelining saturates well before 32 stages.
+        kv_stage = min((224 * 1024 - smem_size_q_o) // smem_size_kv_per_stage, 32)
```
drisspg, line 342: do we actually want to be using  32 kv stages?

Johnsonms: Good catch, @drisspg  
32 is essentially "the next round number above 26" — picked to be surgical to the broken case. The unbounded formula only exceeds 32 at `head_dim_padded ∈ {8, 16}`:

| hd_padded | q=1 | q=2 |
|---|---|---|
|  8 → 16 | 108 | 104 |
| 16 | 108 |  52 |
| 32 |  26 |  24 |
| 64 |  12 |  10 |
| 96 |   7 |   5 |
| 128 |   5 |   3 |

(1CTA path; 2CTA is gated to hd_padded ∈ {128, 192} at `interface.py:572` and its max value there is 10, also below 32.)

So `min(..., 32)` only fires at hd_padded ∈ {8, 16}; anything from 27 to ~50 would be equally surgical. Dropping below 26 starts perturbing kernel staging for hd_padded ∈ {32, 64} (clamp=8 would give 3× fewer stages at hd=32) — and we only have perf data at hd=16, not those.

At hd=16 itself, swept clamp on B200 (`batch=4, nheads=16, hd=16, bf16, causal=False`):

| clamp | sl=4096 ms | TFLOPS | sl=16384 ms | TFLOPS |
|---|---|---|---|---|
|  2 | 0.3778 | 181.9 | 1.4705 | 186.9 |
|  4 | 0.3818 | 180.0 | 1.4858 | 185.0 |
|  8 | 0.3819 | 180.0 | 1.4860 | 185.0 |
| 16 | 0.3824 | 179.7 | 1.4874 | 184.8 |
| 32 | 0.3830 | 179.4 | 1.4883 | 184.7 |

Everything is within ~1% — clamp value doesn't measurably matter at hd=16 either, so keeping 32 just avoids changing kernel staging anywhere outside the broken case.

drisspg (approved):

PR #2597, opened 2026-05-27
[Cute Sm100] support dropout

endurehero:
# FA4 SM100 Dropout Support

Adds FA2-aligned dropout to **FA4 SM100/SM110** (forward + backward), developed with **CodeAgent**.

## Summary

- New `flash_attn_func` args: `p_dropout`, `rng_state`, `return_dropout_mask`.
- Per-element mask is **bit-identical to FA2** under the same Philox `(seed, offset)`; fwd / `dq,dk,dv` errors bounded by FA2's own bf16-vs-fp32 error.
- Validated end-to-end on an internal MoE model (GB200): per-step train-loss diff vs an FA2-dropout run stays in **1e-3 – 1e-4**; downstream eval matches within run-to-run noise.

## Performance

GB200, bf16, cuDNN 9.24.0, `nheads=16`, `total_seqlen≈32k`.

- **Dropout backward**: FA4 beats cuDNN on every config (~1.4–1.7× at hdim=64, ~1.1–1.2× at hdim=128).
- **Dropout forward**: FA4 and cuDNN trade wins; FA4 leads on causal long sequences.
- Dropout costs ~20–40% over no-drop FA4, roughly constant across seqlen (fwd ≈ 75–85%, bwd ≈ 60–70% of no-drop TFLOPS).
- Benched at `p_dropout=0.125` (cuDNN requires a multiple of 1/16; FA4 has no such limit).

### Forward
benchmark_dropout_fwd

### Backward
benchmark_dropout_bwd

```bash
PYTHONPATH=. python tests/cute/benchmark_dropout.py
```

## Correctness

`tests/cute/test_flash_attn_dropout.py`: asserts per-cell mask match vs FA2, bounds fwd/bwd error, and checks kept-fraction ≈ `1−p` (±2%). Coverage: `causal ∈ {F, T}`, `p ∈ {0.1, 0.25}`, `d ∈ {64, 128}`, 3 shapes.

```bash
PYTHONPATH=. pytest -q tests/cute/test_flash_attn_dropout.py
```

## Implementation

Reuses FA2's Philox layout; the mask is generated **and** consumed inline in the softmax tile loop — no global mask tensor on the critical path.

- **Forward**: Philox bits split into bytes, compared against an 8-bit threshold from `p_dropout` (keep ⇔ `byte ≥ threshold`); kept elements scaled by `rp = 1/(1−p)`, dropped ones zeroed in-registers, fused into the rP fragment (MMA layout unchanged). `return_dropout_mask=True` optionally exports a `(b, h, sq, sk)` uint8 mask (test/debug only; backward never reads it).
- **Backward**: no stored mask — replays the Philox stream via `rng_state` and regenerates the mask per tile, folded into the `dS` chain. Several internal register-budget knobs auto-enable when dropout is on (`merge_exp_dropout_ds`, `packed_dropout_fma`, `direct_sdS_xchg_write`, `skip_drop_bits_pack`).

### API

```python
out, lse, rng_state, dropout_mask = flash_attn_func(
    q, k, v, causal=...,
    p_dropout=0.1,             # NEW
    rng_state=None,            # NEW: [seed, offset] int64; reproduce/pin seed
    return_dropout_mask=True,  # NEW: uint8 (b, h, sq, sk), test/debug only
    return_lse=True,
)
```

- `p_dropout` is rounded to 6 decimals in the kernel cache key (nearby rates share a binary).
- `rng_state=None` with `p_dropout>0` → sampled from the default CUDA generator and returned for backward/replay.
- Dropout asserted off for `qv` (MLA), `split_kv`, `block_sparsity`, and the dedicated hdim-256 kernel.


endurehero:
hi @tridao, this is a function implemented according to my actual production requirements on GB200. I would be very happy if you could give me any suggestions

reubenconducts:
Hi @endurehero, I have a few organization-type comments:

- `dropout_rate` doesn't need to be a compile-time parameter. Instead, just use the boolean `use_dropout` (I prefer that nomenclature to `is_dropout`) to `__init__` and pass `dropout_rate` to `__call__`, in both forward and backward
- please limit comments to what is necessary, such as notes about experimental results (e.g. "Sweeping showed split=64 (50%) yields ~4% speedup over split=96 (75%)" in `flash_fwd_sm100.py`). Note the comment concision currently in the codebase. (More generally, please try to keep diff minimal unless necessary)
- The index arithmetic in `mask.py` is unclear; it can be reformulated in layout algebra and avoid any magic numbers. Additionally, try to deduplicate shared code.
- Remove any dead code paths, such as 	`philox_precompute_round_keys` and `mRngState`, or otherwise utilize them
- I'd suggest splitting the dropout helpers that are currently in `softmax.py` out into a separate `dropout.py` module. 
- Could you rebase on main and re-benchmark?

endurehero:
@reubenconducts Thank you very much for your suggestion. I think it is very reasonable. I will modify it today

endurehero:
Hi @reubenconducts, I've done some refactoring based on your comments, but have not changed the cache mechanism for dropout_rate. I tried making it a runtime arg, but forward got 17–38% slower on GB200. The rate constants live through the whole softmax loop — baked in they're free, but as runtime values they blow the register budget (spill 56→232 B) and, worse, kill constant folding/scheduling in the exp2/FMA path (still ~25–30% slower even after fixing the spill). The rate never changes within a model, so compiling once per rate and running at full speed is just the better deal.  Here are the latest performance figures：(based on cudnn v9.24.0. It looks like cuDNN has a performance issue in the varlen scenario.)

- headdim=64
image

- headim=128
image

PR #2609, opened 2026-05-31
Wire can_implement checks before JIT compile in interface.py

fxdv:
## Summary

Adds early `can_implement()` guards before `cute.compile()` so invalid tile, head-dimension, or shared-memory configurations fail with a descriptive `RuntimeError` instead of an opaque CuTeDSL error.

Reachable guarded paths:

- Forward: `FlashAttentionForwardSm80`, `FlashAttentionForwardSm120`
- Backward: `FlashAttentionBackwardSm90`, `FlashAttentionBackwardSm120`

SM90/SM100 forward and SM100 backward do not expose compatible predicates and remain unchanged. SM80 backward is not supported by the current interface, so this PR no longer claims or tests that interface path.

The rebase exposed a stale SM90 backward predicate: `num_threads` includes a 128-thread producer warp group, while the old SM80-style divisibility rule rejected the dispatcher's valid 384-thread configurations. The predicate now validates complete producer/MMA warp groups and is tested against current SM90 tiles.

## Merge dependency

#2751 owns the SM80 forward tile heuristic and architecture-aware SMEM-capacity parameter. This PR intentionally does not duplicate it. Once #2751 lands, the SM8x forward guard here must pass `smem_capacity_arch=f"sm_{arch}"` only for SM8x and add the `_arch=89`, `head_dim=256`, `tile_mn=(128, 64)` pre-compile rejection regression. Please do not merge #2609 before that integration is complete.

Backward SM8x capacity parameterization is left to a separate change if SM8x backward support is restored.

## Tests

- `tests/cute/test_can_implement.py`: **10 passed** with `FakeTensorMode` on a remote Python/CUDA stack.
- #2751's `tests/cute/test_sm80_head_dim256_smem.py`: **15 passed** independently against #2751; its implementation was not copied here.
- Ruff check and format check: passed.

A real FA4 CUDA execution is still required before requesting review. The available host has only V100/SM70 GPUs, while FA4 requires SM80 or newer, so no hardware execution is claimed.

### DIFF
--- a/flash_attn/cute/flash_bwd_sm90.py
+++ b/flash_attn/cute/flash_bwd_sm90.py
@@ -161,9 +161,11 @@ class FlashAttentionBackwardSm90:
             return False
         if tile_n % 16 != 0:
             return False
-        if num_threads % 32 != 0:
-            return False
-        if (tile_m * 2) % num_threads != 0:
+        # num_threads includes one 128-thread producer warp group in addition to the
+        # MMA warp groups. The tuned SM90 configs use 384 total threads with tile_m
+        # values such as 64, 80, and 128, so the SM80-style tile_m divisibility check
+        # does not apply here. Require only a structurally valid warp-group count.
+        if num_threads < 256 or num_threads % 128 != 0:
             return False
         return True
 
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -979,7 +979,20 @@ def _flash_attn_fwd(
             raise ValueError(
                 f"Unsupported compute capability: {arch}. Supported: 8.x, 9.x, 10.x, 11.x, 12.x"
             )
-        # TODO: check @can_implement
+        # Validate the kernel configuration before JIT compilation so invalid tile /
+        # head-dim / SMEM configs fail with a clear error instead of an opaque CuTeDSL
+        # compile error. Only the SM80 and SM120 forward kernels expose can_implement;
+        # the SM90 and SM100 forward kernels do not define it, so they are skipped here.
+        if arch // 10 in [8, 12]:
+            if not fa_fwd.can_implement(
+                dtype, head_dim, head_dim_v, tile_m, tile_n, 1, num_threads, causal,
+            ):
+                raise RuntimeError(
+                    f"FlashAttention forward kernel ({type(fa_fwd).__name__}) cannot be "
+                    f"implemented with given parameters: dtype={dtype}, head_dim={head_dim}, "
+                    f"head_dim_v={head_dim_v}, tile_m={tile_m}, tile_n={tile_n}, "
+                    f"num_threads={num_threads}, causal={causal}"
+                )
         if qv is not None:
             _flash_attn_fwd.compile_cache[compile_key] = cute.compile(
                 fa_fwd,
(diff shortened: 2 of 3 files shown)


fxdv:
Hi @drisspg @Johnsonms — would either of you be able to take a look when you have time?

Adds early `can_implement()` guards in the fwd/bwd compile paths (mirroring the existing combine-kernel pattern), so invalid tile/head-dim/SMEM configs raise a clear `RuntimeError` instead of failing deep inside CuTeDSL compile. Covers SM80/SM120 fwd and SM80/SM90/SM120 bwd where `can_implement` already exists.

Includes GPU-free tests in `tests/cute/test_can_implement.py`. Happy to rebase on latest `main` (currently ~12 commits behind). Thanks!

arbi-dev:
Nice, this is a real gap that's worth closing. One thing worth fixing before/alongside merge: `FlashAttentionForwardSm80.can_implement` (and `FlashAttentionBackwardSm80.can_implement`, same pattern), which this PR wires up, still compute their SMEM budget from a hardcoded `"sm_80"` string:

```python
# flash_attn/cute/flash_fwd.py:168-169 (main @ c75d019)
# TODO: sm86 and sm89
smem_capacity = utils_basic.get_smem_capacity_in_bytes("sm_80")
```

`SMEM_CAPACITY_MAP` in `cutlass.utils` gives `sm_80: 166912` (A100) but `sm_86: 101376` and `sm_89: 101376` (RTX 30xx/40xx) — a 39% smaller budget. So on sm_86/sm_89 this guard checks against the wrong (larger) capacity and will incorrectly pass configs that actually overflow at launch.

Concretely: `_flash_attn_fwd`'s SM80 branch (`interface.py:539`, `fwd_cfg = FwdConfig(128, 64, True, True)  # SM80, should tune`) picks the same `tile_m=128, tile_n=64` for every head_dim. At `head_dim=256` that needs `2*(128*256) + 2*64*(256+256) = 131072` bytes of static SMEM. `can_implement` as wired by this PR would compute `131072 <= 166912` → `True` (guard passes) on an RTX 4090 (sm_89), but the real launch-time budget there is `101376` bytes, so the kernel still crashes with a raw `cudaErrorInvalidValue: Allocated: 131072 bytes. Max: 101376 bytes.` — the exact opaque failure this PR is trying to turn into a clean error. I reproduced this on an actual RTX 4090.

I'm filing a follow-up issue + PR for the underlying root cause (the SM80 tile heuristic has no head-dim branching at all, unlike `_tile_size_fwd_sm90`) plus a fix for this capacity-constant bug specifically, since it's a separate issue from what this PR addresses (wiring the guard calls). My fix parameterizes `can_implement`'s capacity lookup (defaulting to `"sm_80"` for backward compatibility with existing call sites) so a caller — including this PR's guards, once rebased — can pass the actual detected arch (e.g. `f"sm_{arch}"`) instead of the hardcoded constant.

Thanks for pushing this forward — the early-guard idea here is the right shape, this is just one more piece needed for it to actually catch the sm_86/sm_89 case.

fxdv:
Thanks for the detailed reproduction and for opening #2750 / #2751 — agreed that wiring the guard without passing the actual SM8x architecture leaves a false-positive path on sm_86/sm_89.

I’ll avoid duplicating the forward fix from #2751. I’m rebasing #2609 onto current `main` now and will keep the integration order explicit: once #2751 lands, this PR will pass `smem_capacity_arch=f"sm_{arch}"` only on the SM8x forward path and add a regression for `_arch=89`, `head_dim=256`, and explicit `tile_mn=(128, 64)`, asserting that the guard rejects it before compilation.

One nuance I confirmed while revalidating the scope: current `_flash_attn_bwd` rejects SM8x at entry (supported families are SM90/SM100/SM110/SM120), so `FlashAttentionBackwardSm80.can_implement` is not reachable through this interface path. I’ll narrow this PR’s backward guard/tests and description to the reachable SM90 and SM120 paths, and leave backward SM8x capacity parameterization for a separate change if that support is restored.

Thanks again — the live 4090 data and the split between the tile fix and guard wiring are very helpful.

PR #2354, opened 2026-03-15
[CuTe, Sm100] PackGQA for backward

reubenconducts:
EDIT 2026-06-10: Re-opening (as draft, until I've rebased), because we've found pack-gqa to beneficial in SWA.

This PR implements PackGQA for the backward pass of Sm100. This is beneficial for two reasons:
1. Bitwise equivalence with PackGQA in the forward pass, which is not currently the case when using e2e
2. Meaningful performance increase at shorter sequence lengths, coming in part from the fact that dK/dV postprocess is no longer needed. Some benchmarks are attached below. 
There is nothing preventing this from being added to Sm90, I just haven't done so yet. It's less necessary as there is no bitwise invariance issue. 

One design choice comment: I have introduced a `get_m_block_from_iter` method into the main bwd kernel to "unswizzle" the head swizzle inherent in packgqa, which had led to significant L2 spilling and a meaningful performance decrease. This could be tuned further, both for packgqa and other configurations. 

cc @tridao @jayhshah 

```
### hdim=128, causal=False, s_q=32768, b=1, nheads=32, nheads_kv=4, varlen=False, det=False ###
FA Python bwd:          34.999ms, 1256.6 TFLOPS
FA Python bwd pack_gqa: 33.994ms, 1293.8 TFLOPS

### hdim=128, causal=True, s_q=32768, b=1, nheads=32, nheads_kv=4, varlen=False, det=False ###
FA Python bwd:          17.597ms, 1249.6 TFLOPS
FA Python bwd pack_gqa: 18.099ms, 1215.0 TFLOPS

### hdim=128, causal=False, s_q=16384, b=2, nheads=32, nheads_kv=4, varlen=False, det=False ###
FA Python bwd:          17.952ms, 1225.0 TFLOPS
FA Python bwd pack_gqa: 17.435ms, 1261.3 TFLOPS

### hdim=128, causal=True, s_q=16384, b=2, nheads=32, nheads_kv=4, varlen=False, det=False ###
FA Python bwd:          9.369ms, 1173.6 TFLOPS
FA Python bwd pack_gqa: 9.693ms, 1134.3 TFLOPS

### hdim=128, causal=False, s_q=8192, b=4, nheads=32, nheads_kv=4, varlen=False, det=False ###
FA Python bwd:          9.353ms, 1175.6 TFLOPS
FA Python bwd pack_gqa: 8.862ms, 1240.7 TFLOPS

[...]

PR #2610, opened 2026-05-31
Enable pack_gqa backward on SM100/SM110

fxdv:
## Summary

This PR enables `pack_gqa` in the SM100/SM110 (Blackwell) backward path, which was previously hardcoded off. The forward kernel already packs multiple query heads per KV head; this change mirrors that packing into the backward kernel so a single KV-head tile processes all of its query heads.

**What changed**
- Thread `pack_gqa` from the forward path through to the backward path via `FlashAttnFunc`/`ctx` and `_flash_attn_bwd`.
- Pack the `mQ`/`mdO`/`mLSE`/`mdPsum` layouts in `FlashAttentionBackwardSm100`.
- Set `qhead_per_kvhead_packgqa` in the backward tile scheduler, mask, and `SeqlenInfoQK`, and map `head_idx` to the KV head when packing.
- Scale the backward m-block count in `block_info.get_m_block_min_max` by the pack-gqa factor while keeping the logical `seqlen_q` for masking/score_mod.
- Use a packed `dq_accum` buffer for the kernel, run the (unchanged) dQ postprocess on the packed `dq` buffer, then scatter back to the dense `dq` tensor. dK/dV accumulate directly in the packed tiles, so their accum+postprocess steps are skipped.

**Scope / gating**
Packing is only enabled when all of the following hold:
- Architecture is SM100/SM110.
- Dense (non-varlen, non-block-sparse).
- Non-causal / non-local.
- Non-deterministic.
- Not head_dim 256.
- `m_block_size % qhead_per_kvhead == 0`.

SM80/SM90/SM120 still force `pack_gqa` off in the backward path. SM90 backward `pack_gqa` remains unimplemented and is still a separate xfail.

**Why**
Improves performance for GQA/MQA training on Blackwell, bringing the backward path in line with the forward path which already packs heads.

## Test plan

- `tests/cute/test_score_mod.py::test_cute_vs_flex_attention_backward_pack_gqa` now exercises the SM100 packed backward path (it remains xfail only on SM90).

## Validation caveat

This change has been implemented and statically validated (ruff reports no new errors; AST parse OK), but it has **not** yet been run on hardware — the author has no NVIDIA GPU available. It needs SM100 (Blackwell) CI/hardware validation before merge, specifically of:
- the packed `dq_accum` buffer and the dQ postprocess scatter back to dense `dq`,
- mask boundaries under packing,
- the packed TMA layouts (`mQ`/`mdO`/`mLSE`/`mdPsum`).

### DIFF
--- a/flash_attn/cute/block_info.py
+++ b/flash_attn/cute/block_info.py
@@ -56,7 +56,15 @@ class BlockInfo:
 
     @cute.jit
     def get_m_block_min_max(self, seqlen_info: SeqlenInfoQK, n_block: Int32) -> Tuple[Int32, Int32]:
-        m_block_max = cute.ceil_div(seqlen_info.seqlen_q, self.tile_m)
+        # With PackGQA the backward iterates over the packed Q dimension
+        # (qhead_per_kvhead_packgqa * seqlen_q rows per kv-head), so the number of
+        # m_blocks scales by the packing factor. seqlen_info.seqlen_q stays the
+        # logical (unpacked) sequence length so masking/score_mod can recover the
+        # per-head query position.
+        seqlen_q_packed = seqlen_info.seqlen_q
+        if const_expr(self.qhead_per_kvhead_packgqa > 1):
+            seqlen_q_packed = seqlen_info.seqlen_q * self.qhead_per_kvhead_packgqa
+        m_block_max = cute.ceil_div(seqlen_q_packed, self.tile_m)
         m_block_min = 0
         if const_expr(self.is_causal or (self.is_local and self.window_size_right is not None)):
             n_idx_min = n_block * self.tile_n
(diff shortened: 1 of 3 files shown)


fxdv:
Hi @Johnsonms — would you be able to review this when convenient?

Enables `pack_gqa` in the SM100/SM110 **backward** path (previously hardcoded off): threads `pack_gqa` fwd→bwd via `FlashAttnFunc`, packs mQ/mdO/mLSE layouts, sets `qhead_per_kvhead_packgqa` in the bwd scheduler/mask, and handles packed dq_accum + dQ postprocess scatter.

Gated to dense, non-causal/local, non-varlen, non-block-sparse configs on SM100/110. SM90 remains xfailed in tests.

**Note:** This is kernel-level and I have not been able to run GPU tests locally (no CUDA host). Would appreciate a sanity check on SM100 hardware — especially the packed dq_accum/postprocess scatter. Also happy to rebase on latest `main` (~12 commits behind; aware of #2629 removing forward packgqa guards). Thanks!

Johnsonms:
Will take a look by end of this week, Thanks

jayhshah:
We have an open PR https://github.com/Dao-AILab/flash-attention/pull/2354 on this that we're planning on merging. I note that pack gqa is generally non-performant for backward so the hardest part about this will be getting the heuristic right.

PR #2616, opened 2026-06-01
Add aux-scalars to interface to enable dynamic ints and floats in expressions

drisspg:
Add aux-scalars to interface to enable dynamic ints and floats in expressions

## Summary

Right now we have Aux-tensors mixin data coming from global tensors,

This pr provides the same functionality but for scalar values via  `aux_scalars`. 

With this we can write mask/score mods that are dynamic over python ints and floats instead of needing to recompute. The most common case is for decode attention. Today that is done via 1 off seq_len info but in general this could have been done via generic query and key size or a dynamic integer offset ( handy if you cant derive from input tensor sizes).

### Examples

```Python
   @cute.jit
   def shifted_causal_mask(batch, head, q_idx, kv_idx, seqlen_info, aux_tensors, aux_scalars):
       q_offset = cutlass.Int32(aux_scalars[0])
       return kv_idx <= q_idx + q_offset


   out = flash_attn_func(
       q,
       k,
       v,
       mask_mod=shifted_causal_mask,
       aux_scalars=[cutlass.Int32(cache_position)],
   )
  ```

I have the PyTorch PR here: https://github.com/pytorch/pytorch/pull/186019
This removes the largest coverage gap between triton and flex/flash :)

## Review guidance

I can merge the two in to an AuxStruct if we want to reduce the number of kwargs for internal kernel funcs -> down for whatever

### DIFF
--- a/flash_attn/cute/block_sparse_utils.py
+++ b/flash_attn/cute/block_sparse_utils.py
@@ -18,6 +18,7 @@ from quack import copy_utils
 from flash_attn.cute.block_sparsity import BlockSparseTensors
 from flash_attn.cute.named_barrier import NamedBarrierBwd
 from flash_attn.cute.seqlen_info import SeqlenInfoQK
+from flash_attn.cute.utils import AuxData
 
 
 @cute.jit
@@ -1363,7 +1364,7 @@ def consume_block_sparse_mma_bwd_sm90(
     score_mod_bwd_fn=None,
     subtile_factor: cutlass.Constexpr = 1,
     m_block_max: int = 0,
-    aux_tensors=None,
+    aux_data: AuxData = AuxData(),
     fastdiv_mods=(None, None),
 ):
     """SM90 backward block sparse MMA consumption with separate partial/full loops.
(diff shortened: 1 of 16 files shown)

### CODE REVIEW

**Review by tridao — APPROVED** [review]

**Comment by tridao** [comment]
If both scalar and tensor are passed in, do we prioritize one?

**Review by Johnsonms — APPROVED** [review]

**Comment by drisspg** [comment]
Should be orthogonal both are just scratch pads that can be indexed into, added a test showing how to use both: 

```Py

@cute.jit
def tensor_bias_and_scalar_scale_score(
    score, b_idx, h_idx, q_idx, kv_idx, seqlen_info, aux_tensors, aux_scalars
):
    batch_bias = aux_tensors[0]
    b_frag = cute.make_rmem_tensor(1, cutlass.Int32)
    b_frag.store(b_idx)
    bias_frag = cute.make_rmem_tensor(1, batch_bias.element_type)
    bias_frag[0] = batch_bias[b_frag[0]]
    return score * cute.full_like(score, cutlass.Float32(aux_scalars[0])) + bias_frag.load().to(
        cutlass.Float32
    )

```

PR #2614, opened 2026-06-01
[ROCm] Bump Triton to >=3.6.0 and update aiter submodule

micmelesse:
Fixes #2580. We bump the Triton version and update the aiter submodule. We also update the Flash Attention v3 setup.py so we can start testing the v3 API on ROCm.

CI run on the aiter side passes: https://github.com/ROCm/aiter/actions/runs/26600146840/job/78381581829

PR #2620, opened 2026-06-02
[Triton] Fix graph capture issues and env var

### LINKED ISSUE #2580: ROCM + triton build broken
I'm trying to build in a new venv with the instructions from the README but the build is not working when it tries to build _aiter_.

```sh
cd flash-attention
FLASH_ATTENTION_TRITON_AMD_ENABLE="TRUE" pip install --no-build-isolation .
```

**output**
```
INFO: pip is looking at multiple versions of amd-aiter to determine which version is compatible with other requirements. This could take a while.
ERROR: Could not find a version that satisfies the requirement flydsl==0.1.1.dev409 (from amd-aiter) (from versions: 0.1.5.dev515, 0.1.5, 0.1.6.dev529, 0.1.6, 0.1.7.dev551, 0.1.7, 0.1.8.dev574, 0.1.8)
ERROR: No matching distribution found for flydsl==0.1.1.dev409
```

Any ideas how I should get (or opt out of) `flydsl` or otherwise get the build to work? cc @micmelesse 

I'm using python3.13 on Arch Linux.

Resolved by #2540

micmelesse:
Bumps aiter to pick up the graph-capture fix (ROCm/aiter#2764) and removes the `AITER_TRITON_ONLY=1` env variable in setup.py which was added in #2614. It was added to fix #2580 but replacing a full build of aiter with a slim, triton only build might cause issues for other users.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -227,11 +227,8 @@ if IS_ROCM:
             assert os.path.isdir("third_party/aiter"), (
                 "third_party/aiter is missing, please use source distribution or git clone"
             )
-        aiter_env = os.environ.copy()
-        aiter_env.setdefault("AITER_TRITON_ONLY", "1")
         subprocess.run(
             [sys.executable, "-m", "pip", "install", "--no-build-isolation", "third_party/aiter"],
-            env=aiter_env,
             check=True,
         )
     elif ROCM_BACKEND == "ck":


tridao (approved):

micmelesse:
Full CI run for the aiter commit is green. https://github.com/ROCm/aiter/actions/runs/26849965459/job/79179565895

PR #2621, opened 2026-06-03
[Cute,Bwd,Sm100] add sparse MLA (Deepseek v4) backward kernels

jayhshah:
We add a backward pass for the MLA weight absorbed attention formula

`O = softmax(scale * (Q @ K.T + Qv @ V.T)) @ V`

in hdim-hdimv=64-512 and sparse KV, MQA 128 case. Eliminating Q, K and renaming terms, this specializes to a backward pass for

`O = softmax(scale * (Q @ K.T)) @ V`

when K=V, hdim 512, MQA 128, as in Deepseek v4 Pro sparse attention.

Since with MQA 128 the kernel is embarassingly parallel over the seqlen_q dimension with 2 CTAs per token, it's natural to compare sparse sequence parallel with dense tensor parallel times. For example, with sq = 4k, sk = 32k, topk=2k, we get

```
FWD tflops=1180, ms=1.865
BWD tflops=820, ms=6.70
```

versus for dense causal MHA 16, hdim-hdimv 192-128, sq=32k, sk=32k

```
FWD tflops=1834, ms=3.00
BWD tflops=1095, ms=12.55
```


jayhshah:
For the sparse MLA backward logic, I need to save materialized blockwise P and row max from forward. Thus I have `_flash_attn_fwd` now return `out, lse, p, row_max` and save that. This is unconditional in terms of the signature though these args are Nones if unused -- would this be an acceptable change @tridao ?

Johnsonms (approved):

PR #2622, opened 2026-06-03
fix: add weights_only=True to all torch.load call sites

### LINKED ISSUE #2583: Unsafe torch.load in FlashAttention training / eval checkpoint paths
So apparently someone issued [CVE-2026-31253](https://www.notion.so/CVE-2026-31253-35d1e1393188813f9e77e2038104bc49) because `torch.load` is being used in Flash Attention without `weights_only=True` in [this function](https://github.com/Dao-AILab/flash-attention/blob/4178915405522cb161ece80005e882c7465693b7/training/src/utils/checkpoint.py#L20) and [this function](https://github.com/Dao-AILab/flash-attention/blob/4178915405522cb161ece80005e882c7465693b7/training/src/eval.py#L34). Presumably they may take issue with [this](https://github.com/Dao-AILab/flash-attention/blob/4178915405522cb161ece80005e882c7465693b7/flash_attn/utils/pretrained.py#L62) and [this](https://github.com/Dao-AILab/flash-attention/blob/4178915405522cb161ece80005e882c7465693b7/flash_attn/models/llama.py#L388) as well.

I don't see any PR or issue open for this, not sure why that wasn't done before the CVE was issued, but I figured I'd start one.

aryanputta:
## Summary

Adds `weights_only=True` to all four `torch.load` call sites in the repository, addressing the unsafe deserialization vector reported in #2583.

**Why this matters:** `torch.load` without `weights_only=True` uses Python's `pickle` protocol for deserialization. A malicious checkpoint file can execute arbitrary Python code on the machine that loads it — CWE-502 / CVE-class issue. Passing `weights_only=True` restricts deserialization to tensors, primitive types, dicts, lists, and tuples only.

## Sites fixed

| File | Function |
|---|---|
| `training/src/utils/checkpoint.py` | `load_checkpoint()` |
| `training/src/eval.py` | eval checkpoint loader |
| `flash_attn/utils/pretrained.py` | `partial(torch.load, ...)` loader |
| `flash_attn/models/llama.py` | `state_dicts_from_checkpoint()` |

## Compatibility

`weights_only=True` was introduced in PyTorch 1.13 and is the default in PyTorch ≥ 2.4. FA4's CuTeDSL dependency already requires a modern PyTorch 2.x build, so there is no compatibility regression. All four sites load model state dicts, which are always `weights_only`-safe (tensors + primitive containers only).

Fixes #2583

### DIFF
--- a/flash_attn/models/llama.py
+++ b/flash_attn/models/llama.py
@@ -385,7 +385,7 @@ def state_dicts_from_checkpoint(
 ) -> List[dict]:
     # Need to sort, otherwise we mess up the ordering and the weights are wrong
     return [
-        torch.load(path, map_location="cpu")
+        torch.load(path, map_location="cpu", weights_only=True)
         for path in sorted((Path(checkpoint_path) / model_name).glob("consolidated.*.pth"))
     ]
 
--- a/flash_attn/utils/pretrained.py
+++ b/flash_attn/utils/pretrained.py
@@ -59,7 +59,7 @@ def state_dict_from_pretrained(model_name, device=None, dtype=None):
     if load_safe:
         loader = partial(safe_load_file, device=mapped_device)
     else:
-        loader = partial(torch.load, map_location=mapped_device)
+        loader = partial(torch.load, map_location=mapped_device, weights_only=True)
 
     if is_sharded:
         # resolved_archive_file becomes a list of files that point to the different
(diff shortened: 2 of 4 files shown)


tridao (approved):

tridao:
Thanks!

ppham-nv:
https://github.com/advisories/GHSA-7g5w-pq96-8c5w should be updated to reference this PR

PR #2629, opened 2026-06-05
[Cute,Sm100,Fwd] use correction warps if not tma store; remove outdated packgqa guard

jayhshah:
We remove the pack gqa = False guard for split kv + non-varlen q. This was an outdated guard stemming from the pack gqa address shuffle being broken prior to PR https://github.com/Dao-AILab/flash-attention/pull/2097. We also switch to correction warps for store when tma store O is False in general, rather than only for varlen. Finally, to declutter the interface a bit we remove a second irrelevant guard for pack gqa in the MLA kernel when nheads > 128, since this case is unsupported anyway.

Changes tested via `test_flash_attn.py`

### DIFF
--- a/flash_attn/cute/flash_fwd_sm100.py
+++ b/flash_attn/cute/flash_fwd_sm100.py
@@ -176,10 +176,15 @@ class FlashAttentionForwardSm100:
         self.is_causal = is_causal
         self.is_local = is_local
         self.is_varlen_q = is_varlen_q
-        self.use_correction_warps_for_epi = is_varlen_q
         self.qhead_per_kvhead = qhead_per_kvhead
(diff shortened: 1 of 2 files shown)


drisspg (approved):
Nice; I also found this: https://github.com/Dao-AILab/flash-attention/pull/2537/changes 

this is a better clean up

jayhshah:
I'll merge that and merge this on top for the additional changes.

jayhshah:
@drisspg Whoops, I merged your commit internally. Did you mean to make that PR to main?

I'll just merge this one and we can land the other tests at some point.

drisspg:
No worries, yours supersedes and is better :)

PR #2642, opened 2026-06-10
[Fwd,Sm100] Tune FP8 causal hd128 ex2_emu_freq (8 vs inherited 16)

Johnsonms:
### Background

FP8 forward on Blackwell is **MUFU/exp-bound**, not MMA-bound: once the GEMMs run in FP8 the softmax `exp2` (on the MUFU unit) is the critical path. FA4 mitigates this with *ex2 emulation* — computing a fraction of `exp2` on the ALU instead of the MUFU (`_FP8_TUNING_CONFIG` / `apply_exp2_convert`, gated by `ex2_emu_freq`).

`_FP8_TUNING_CONFIG` only had one hand-tuned entry — `(True,False,128,False)` (hd128 non-causal). The **causal** hd128 key `(False,True,128,False)` had no FP8 entry, so it inherited bf16's `ex2_emu_freq=16`. Since FP8 is more MUFU-bound than bf16, a more aggressive emulation frequency helps.

### Change

Add one entry — causal hd128 FP8 uses `ex2_emu_freq=8` (offloads more `exp` from MUFU):

```python
(False, True, 128, False): {'ex2_emu_freq': 8, 'ex2_emu_start_frg': 1},
```

Scoped to the causal key only: `freq=8` regresses non-causal hd128 (0.94×), so non-causal keeps its existing `freq=10`.

### Results (B200 / sm100)

Thermally-matched back-to-back A/B (same process, hot GPU, `freq=16`↔`freq=8` interleaved, median of 300 iters, `nheads=16` = benchmark default) across the official `benchmark_flash_attention_fp8.py` causal-hd128 shapes:

| b | s | freq=16 TFLOPs | freq=8 TFLOPs | Δ |
|---|---|---:|---:|---:|
| 32 | 512 | 500.6 | 516.3 | +3.1% |
| 16 | 1024 | 796.5 | 832.5 | +4.5% |
| 8 | 2048 | 1124.6 | 1175.1 | +4.5% |
| 4 | 4096 | 1407.9 | 1481.3 | +5.2% |
| 2 | 8192 | 1604.1 | 1661.7 | +3.6% |
| 1 | 16384 | 1683.7 | 1726.0 | +2.5% |

Consistent **+2.5% – +5.2%**, MHA and GQA (q16/kv2 separately validated), accuracy-neutral (FP8-vs-bf16 mean-abs-err unchanged; `benchmark_flash_attention_fp8.py --check` passes 24/24 before and after). `ex2_emu` decode↔prefill consistency test still passes.

> Note: numbers are a matched in-process A/B (freq=16↔8 back-to-back, same thermal state); comparing two *separate* full benchmark runs is unreliable at this 3–5% magnitude due to cross-run thermal/clock drift.

### Microarchitecture (NCU)

Nsight Compute on the fwd kernel (causal hd128, b2 s8192, single launch, base clock) confirms the mechanism — `freq=8` rebalances work off the MUFU/transcendental (XU) pipe onto FMA/ALU:

| metric | freq=16 | freq=8 | Δ | reading |
|---|---:|---:|---:|---|
| Duration | 633 µs | 607 µs | −4.1% | kernel finishes faster (matches the +3–5% app-level win) |
| XU (MUFU) pipe util | 69.1% | 63.2% | −8.6% | pressure comes off the bottleneck pipe |
| FMA pipe util | 15.9% | 22.2% | +39% | emulated `exp2` now runs here instead |
| ALU pipe util | 33.8% | 38.5% | +14% | emulation also consumes ALU |
| warp stall: `mio_throttle` | 0.34 | 0.12 | −66% | far fewer warps stalled waiting on the MUFU queue — the actual bottleneck |
| achieved occupancy | 23.1% | 23.1% | ±0 | unchanged → a pure pipe rebalance, not an occupancy/memory effect |

The kernel was MUFU-bound (XU ~69%, `mio_throttle` the dominant short stall); moving more `exp2` to emulation drops XU pressure and cuts `mio_throttle` stalls by ~66%, so the kernel finishes faster at unchanged occupancy — a pure pipe rebalance, not an occupancy/memory effect.

Depends on #2640 (FP8 fwd fix) for FP8 to run on cutlass-dsl ≥4.5.2.

### DIFF
--- a/flash_attn/cute/flash_fwd_sm100.py
+++ b/flash_attn/cute/flash_fwd_sm100.py
@@ -94,6 +94,11 @@ _TUNING_CONFIG = {
 }
 _FP8_TUNING_CONFIG = {
     (True, False, 128, False): {'ex2_emu_freq': 10, 'ex2_emu_start_frg': 1, 'num_regs_softmax': 160, 'num_regs_correction': 72},
+    # Causal hd128 FP8 previously inherited bf16's freq=16. FP8 fwd is MUFU/ex2-bound, so a more
+    # aggressive emulation freq=8 offloads more exp from MUFU: +3.4%(4k)..+5.5%(16k) on B200,
+    # MHA & GQA, accuracy-neutral (3-run validated, locked clocks). freq=8 would regress non-causal
+    # (0.94x), hence keyed on is_causal=True only.
+    (False, True, 128, False): {'ex2_emu_freq': 8, 'ex2_emu_start_frg': 1},
 }
 _FP8_SMALL_HDIM_REGS = {
     False: {"num_regs_softmax": 168, "num_regs_correction": 96, "num_regs_other": 80},


Johnsonms:
Added @liaojianjin for review and verification

jayhshah (approved):

PR #2645, opened 2026-06-11
[CuTe,Bwd] Include subtile_factor in backward compile keys

leonardHONG:
Since #2411, `subtile_factor` (`sparse_q block size // m_block_size`) varies with the block-sparse configuration and specializes the backward kernels as a `cutlass.Constexpr`, but it was missing from both backward compile keys. The forward key already includes the corresponding `q_subtile_factor`.

As a result, two configs that differ only in sparse Q block size, for example `head_dim=64` with `block_size=(128,128)` followed by `block_size=(256,128)` in the same process, can hash to the same compile key. The second run then silently reuses a kernel compiled for the wrong `subtile_factor`, producing incorrect `dQ/dK/dV` with no runtime error.

This PR adds `subtile_factor` to both backward compile keys, next to the other block-sparse key fields. The SM90 key fixes the observed issue; the SM100/SM110 key entry is defensive, since `sparse_q` is currently only derived on SM90.

Verified on H20 / SM90 with an all-blocks-active sparse pattern, which is mathematically equivalent to dense attention. Dense backward is used as the reference, and both sparse configurations are run in one process to exercise the in-memory compile cache.

| config                  | dq     | dk   | dv   |
| ----------------------- | ------ | ---- | ---- |
| sparse(128,128), before | 4.3e-4 | 0    | 0    |
| sparse(256,128), before | 1.00   | 1.14 | 0.81 |
| sparse(256,128), after  | 8.6e-4 | 0    | 0    |

Added a regression test in `test_mask_mod.py` that runs both block sizes in one process. It fails before this change and passes after.

Existing block-sparse test suite on H20:

```text

test_mask_mod.py -k block_sparse

19 passed, 57 skipped, 0 failed

```


drisspg:
Hey this was fixed here: https://github.com/Dao-AILab/flash-attention/pull/2661

PR #2634, opened 2026-06-08
FA4 consumer Blackwell (sm_120) integration

thad0ctor:
## FA4 consumer Blackwell (sm_120) integration

Adds FlashAttention-4 (CuTeDSL) support for consumer/workstation Blackwell — RTX 50-series / RTX PRO 6000 / DGX Spark (compute capability 12.x). Dispatch and tile selection are auto-tuned for the arch; **no environment variables are required for normal use.**

### Background
Consumer/workstation Blackwell (`sm_120`) ships 5th-gen tensor cores with new fp4/fp8 data types but — unlike datacenter Blackwell (`sm_100`) — keeps the Ampere-style **warp-level `mma.sync`** programming model: **no WGMMA, no tcgen05, no Tensor Memory** (confirmed against the CUDA 13.x toolkit and ptxas SASS — the tensor-memory subsystem is absent in silicon, not just unexposed). So FA4's `sm_100` path does not apply; `sm_120` reuses the SM80-base kernels (cp.async, plus a TMA + warp-specialized forward) and needs its own dispatch, tiling, and correctness work.

This builds on the in-flight upstream FA4 `sm_120` kernels (Dao-AILab #2553 / #2349 / #2389), which landed initial kernels but left public-API routing, correctness, and coverage gaps:

| Area | Upstream (#2553 / #2349 / #2389) | This PR |
|---|---|---|
| Dense forward | kernels exist; public routing hits dispatcher/runtime gaps | works across fixed + varlen `sm_120` paths via the public API, auto-tuned dispatch |
| Backward | not usable end-to-end for the coverage target | supported head dims work through the public API, incl. D256 dense/causal |
| Packed GQA | crashes / disabled on several `sm_120` GQA shapes | works where valid; invalid cases auto-downgrade; **pack_gqa + SplitKV decode now works** |
| Paged KV | SM80-base path takes the args but doesn't implement `sm_120` correctly | real paged-KV forward through `head_dim ≤ 256`, validated |
| `head_dim > head_dim_v` | can route into the `sm_120` TMA hang | routed to the non-TMA path; hang-guard tests included |
| fp8 KV-cache decode | — | new memory-efficient fp8 e4m3/e5m2 decode (≈1.6–1.9× at GQA ratio ≤ 4, half the KV bandwidth) |
| Correctness / hardening | — | 8 real bugs found and fixed (below); strict arch-gating (non-`sm_120` byte-identical to `main`); CodeRabbit follow-ups |

### What's supported
[...]

PR #2655, opened 2026-06-15
[NVIDIA][CuTe,Fwd,sm120] Fix use_tma_O crash on SM120 (issue #2649)

### LINKED ISSUE #2649: FlashAttentionForwardSm120: use_tma_O incorrectly True on SM_121 → AttributeError on tma_atom_O=None
## Summary

`flash_fwd.py:658` sets `self.use_tma_O` with a guard that fires incorrectly on SM_121, causing an `AttributeError` when the kernel tries to access `tma_atom_O._trait` on a `None` object.

## Environment

- flash-attn-4: v4.0.0b16
- nvidia-cutlass-dsl: 4.5.2
- CUDA: 12.8
- GPU: NVIDIA GB10 (sm_121 — consumer Blackwell / DGX Spark)

## Root Cause

`FlashAttentionForwardSm120` defines `arch = 80` as a **class attribute**. However, `BaseDSL.__init__` (called via `super().__init__()`) clobbers `self.arch` with `BaseDSL._get_dsl().get_arch_enum()`, which on SM_121 returns `Arch.sm_121`.

The current guard:
```python
self.use_tma_O = self.arch >= Arch.sm_90
```

On SM_121: `Arch.sm_121 >= Arch.sm_90` → `True`. The kernel then tries `tma_atom_O._trait`, but `tma_atom_O = None` (SM_121 is not in the TMA-O initialization branch) → `AttributeError: 'NoneType' object has no attribute '_trait'`.

## Fix

```python
# flash_fwd.py:658
# Old:
self.use_tma_O = self.arch >= Arch.sm_90
# New:
self.use_tma_O = Arch.sm_90 <= self.arch < Arch.sm_120
```

SM_121 should use the SM_80 code path (which it already does for everything else), not the SM_120 TMA-O path.

## Verification

Applied this patch locally on GB10 (SM_121). Confirmed: `AttributeError` on `tma_atom_O` no longer occurs. Base MHA forward tests pass numerically.

## Related

This was found during SM_121 (GB10 consumer Blackwell) bringup. SM_121 inherits the SM_80 kernel path — most SM_80 code works correctly, but this guard condition was overly broad.

johnnynunez:
## Summary

Fixes #2649. On SM120 (Blackwell GeForce / RTX PRO 6000 / DGX Spark) the CuTe-DSL forward kernel is currently **unusable** — every call crashes with:

```
  in cpasync.tma_partition (flash_fwd.py epilogue O-store)
```

## Root cause

`FlashAttentionForwardSm120` inherits the forward `__call__`, which sets:

```python
self.use_tma_O = self.arch >= Arch.sm_90
```

This enables the TMA-based O-store epilogue for SM120. But SM120 does not build the TMA store atom (`tma_atom_O` is `None`), because it lacks the WGMMA-era TMA-store epilogue path. The first `tma_partition` on the null atom raises the `AttributeError`.

## Fix

Restrict the TMA O-store to `sm_90 .. sm_119`:

```python
self.use_tma_O = Arch.sm_90 <= self.arch < Arch.sm_120
```

SM120 then falls back to the register→gmem O store already used by the SM80 path, which is what the CpAsync SM120 kernel expects.

## Testing

Verified on **RTX PRO 6000 Blackwell** (sm_120, cc 12.0), torch 2.12.0+cu130, nvidia-cutlass-dsl 4.5.2:

- Before: every SM120 forward call raises the `AttributeError` above.
- After: forward runs and matches PyTorch SDPA reference (bf16), max abs error:
  - hdim=64  non-causal: 1.3e-3
  - hdim=96  non-causal: 1.1e-3
  - hdim=128 causal:     8.1e-3

One-line behavioral change; no effect on sm80/sm90/sm100 paths.

### DIFF
--- a/flash_attn/cute/flash_fwd.py
+++ b/flash_attn/cute/flash_fwd.py
@@ -656,7 +656,11 @@ class FlashAttentionForwardSm80(FlashAttentionForwardBase):
         self.num_Q_load_threads = self.num_threads
         self.num_epilogue_threads = self.num_threads
         # self.use_tma_O = self.arch >= 90 and mCuSeqlensQ is None
-        self.use_tma_O = self.arch >= Arch.sm_90
+        # SM120 (Blackwell GeForce / RTX PRO / DGX Spark) lacks the WGMMA-era TMA-store
+        # epilogue path used here; only sm_90..sm_119 use TMA for the O store. On sm_120
+        # tma_atom_O is None, so using it crashes in tma_partition with
+        # 'NoneType' object has no attribute '_trait'. (issue #2649)
+        self.use_tma_O = Arch.sm_90 <= self.arch < Arch.sm_120
         self._setup_attributes()
         SharedStorage = self._get_shared_storage_cls()
         mQ, mK, mV, mO = [assume_tensor_aligned(t) for t in (mQ, mK, mV, mO)]


thad0ctor:
#2634 disables the broken SM80-base TMA-O path for SM120, though with a broader                                                                   self.use_tma_O = False approach in the SM80-base path.

v0i0 (approved):

PR #2656, opened 2026-06-15
[NVIDIA][CuTe,Fwd,sm120] Implement Pack-GQA on SM120 (+ graceful SplitKV fallback)

johnnynunez:
## Summary

Implements **Pack-GQA** end-to-end on SM120 (Blackwell GeForce / RTX PRO 6000 / DGX Spark), and makes SplitKV fall back gracefully.

### Pack-GQA (the real fix)

Pack-GQA was only *half*-wired in the SM80/SM120 CpAsync forward: the epilogue called `PackGQA.store_O`/`store_LSE`, but the **Q-load and head-indexing still used the plain unpacked path**. So `pack_gqa=True` crashed:

```
ValueError: Operation creation failed
  pack_gqa.py compute_ptr -> elem_pointer -> crd2idx   # packed (h_idx, m_idx) vs unpacked mO layout
```

This wires Pack-GQA fully, mirroring the SM90 implementation:

- Reshape `mQ`/`mO` (`head_idx=2`) and `mLSE` (`head_idx=1`) via `pack_gqa_layout`, folding `qhead_per_kvhead` into the seqlen mode.
- Scheduler args use `cute.size(mQ.shape[0])` (packed total rows) and `seqlen_q_static = mQ.shape[0][1]` (logical seqlen) so causal/mask `q_idx` stay correct.
- Kernel head-indexing: under Pack-GQA `num_head` from the scheduler already indexes the **KV** head, so no division.
- Q-load: gather rows via `PackGQA.load_Q` (per-row `(h_idx, m_idx)` gmem pointers) instead of the contiguous `local_tile` path.

### SplitKV

SplitKV (`num_splits > 1`) is an **SM100-only** feature — SM80/SM90 both assert it unsupported, and SM120 has no forward+combine path. Fall back to `num_splits = 1` (numerically correct) instead of crashing in `_check_type` on the fp32 partials.

## Testing

On **RTX PRO 6000 Blackwell** (sm_120), torch 2.12.0+cu130, nvidia-cutlass-dsl 4.5.2:

| config | pack_gqa=True err vs SDPA | packed vs unpacked |
|---|---|---|
| h8/2 d128 noncausal s1024 | 1.1e-3 | **0.0 (bit-identical)** |
| h8/2 d64 causal s512 | 7.7e-3 | 0.0 |
| h6/3 d128 noncausal s2048 | 9e-4 | 0.0 |
| h8/1 (MQA) d64 s1024 | 1.2e-3 | 0.0 |
| h16/2 d128 causal s1024 | 8.4e-3 | 0.0 |

Pack-GQA output is **bit-identical** to the unpacked path (it's a layout optimization), and `num_splits=3` falls back and matches reference (err 6.8e-4).

Stacked on #2655 (SM120 `use_tma_O` crash fix) — SM120 forward does not run without it.

### DIFF
--- a/flash_attn/cute/flash_fwd.py
+++ b/flash_attn/cute/flash_fwd.py
@@ -656,10 +656,6 @@ class FlashAttentionForwardSm80(FlashAttentionForwardBase):
         self.num_Q_load_threads = self.num_threads
         self.num_epilogue_threads = self.num_threads
         # self.use_tma_O = self.arch >= 90 and mCuSeqlensQ is None
-        # SM120 (Blackwell GeForce / RTX PRO / DGX Spark) lacks the WGMMA-era TMA-store
-        # epilogue path used here; only sm_90..sm_119 use TMA for the O store. On sm_120
-        # tma_atom_O is None, so using it crashes in tma_partition with
-        # 'NoneType' object has no attribute '_trait'. (issue #2649)
         self.use_tma_O = Arch.sm_90 <= self.arch < Arch.sm_120
         self._setup_attributes()
         SharedStorage = self._get_shared_storage_cls()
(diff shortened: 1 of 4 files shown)


thad0ctor:
#2634 already implements SM120 Pack-GQA more fully.  #2656 falls back to num_splits=1 for SM120 SplitKV,
while #2634 implements SM120 SplitKV support, so #2656 would be a functional regression

johnnynunez:
> #2634 already implements SM120 Pack-GQA more fully. #2656 falls back to num_splits=1 for SM120 SplitKV, while #2634 implements SM120 SplitKV support, so #2656 would be a functional regression

Good! Ping @drisspg and @tridao to merge them

drisspg (approved):

PR #2658, opened 2026-06-15
[NVIDIA][CuTe,Bwd,sm120] Enable SM120 backward (dQ_single_wg + softmax_scale None fixes)

johnnynunez:
## Summary

The CuTe-DSL **backward pass did not run on SM120** (Blackwell GeForce / RTX PRO 6000 / DGX Spark) at all — every backward call crashed before launch. Two bugs:

### 1. `dQ_single_wg` UnboundLocalError
`interface._flash_attn_bwd` builds the bwd compile key with `dQ_single_wg`, but the SM120 branch (`arch // 10 == 12`) never defined it (only the SM90 `cfg`-based branch does). First backward call → `UnboundLocalError: cannot access local variable 'dQ_single_wg'`. Define `dQ_single_wg = False` for SM120 (single MMA warp-group dQ).

### 2. `None to Float` at kernel launch
`flash_bwd.py __call__` used `utils.compute_softmax_scale_log2()`, which returns `softmax_scale = None` when `score_mod is None`. But the kernel parameter is typed `cutlass.Float32`, and the backward multiplies `acc_dK` by `softmax_scale` (dK rescale when `qhead_per_kvhead == 1`). So launch crashed:
```
DSLRuntimeError: None to Float conversion is not supported
```
Fix: compute `softmax_scale_log2` inline (fold `LOG2_E`) and keep `softmax_scale` a real `Float32` — exactly what the SM100 backward already does.

### 3. `nvvm.atomicrmw` binding robustness
`utils.atomic_add_fp32` (used by the dQ atomic accumulation) hard-coded the `res=/op=/ptr=/a=` kwargs of the cu12.x `nvvm.atomicrmw` binding. The cu13 build takes positional `(op, ptr, a)` with inferred result and rejects `res=`. Try the cu13 form, fall back to cu12.x — decided once at trace time.

## Testing

On **RTX PRO 6000 Blackwell** (sm_120, cc 12.0), torch 2.12.0+cu130, nvidia-cutlass-dsl 4.5.2:

- Before: every SM120 backward raised `UnboundLocalError` / `None to Float`.
- After: full forward+backward runs and gradients match a PyTorch fp32 reference:

| config | max \|dq/dk/dv − ref\| (bf16) |
|---|---|
| hdim 64, non-causal | 3.3e-3 |
| hdim 96, non-causal | 2.6e-3 |
| hdim 128, causal | 1.3e-2 |

## Note

Stacked on #2655 (SM120 `use_tma_O` forward crash fix) — the forward must run for the backward to be exercisable.

### DIFF
--- a/flash_attn/cute/flash_bwd.py
+++ b/flash_attn/cute/flash_bwd.py
@@ -21,6 +21,7 @@ from flash_attn.cute import utils
 from flash_attn.cute.mask import AttentionMask
 from flash_attn.cute.softmax import call_score_mod, call_score_mod_bwd
 from flash_attn.cute.seqlen_info import SeqlenInfoQK
+from flash_attn.cute.block_info import BlockInfo
 from quack.cute_dsl_utils import ParamsBase
 from flash_attn.cute.tile_scheduler import SingleTileScheduler, SingleTileVarlenScheduler, TileSchedulerArguments
 from flash_attn.cute.block_sparsity import BlockSparseTensors
@@ -41,6 +42,7 @@ class FlashAttentionBackwardSm80:
         num_threads: int = 256,
         pack_gqa: bool = False,
         is_causal: bool = False,
+        is_local: bool = False,
         SdP_swapAB: bool = False,
         dKV_swapAB: bool = False,
         dQ_swapAB: bool = False,
@@ -83,6 +85,7 @@ class FlashAttentionBackwardSm80:
         self.num_threads = num_threads
         self.pack_gqa = pack_gqa
         self.is_causal = is_causal
+        self.is_local = is_local
         self.num_stages_Q = num_stages_Q
         self.num_stages_dO = num_stages_dO
         self.SdP_swapAB = SdP_swapAB
@@ -436,7 +439,9 @@ class FlashAttentionBackwardSm80:
         tile_sched_params = TileScheduler.to_underlying_arguments(tile_sched_args)
         grid_dim = TileScheduler.get_grid_shape(tile_sched_params)
 
-        softmax_scale_log2, softmax_scale = utils.compute_softmax_scale_log2(softmax_scale, self.score_mod)
+        softmax_scale_log2, _ = utils.compute_softmax_scale_log2(
+            softmax_scale, self.score_mod
+        )
         self.kernel(
             mQ,
             mK,
@@ -453,6 +458,8 @@ class FlashAttentionBackwardSm80:
             mSeqUsedK,
             softmax_scale,
             softmax_scale_log2,
+            window_size_left,
+            window_size_right,
             self.sQ_layout,
             self.sK_layout,
             self.sV_layout,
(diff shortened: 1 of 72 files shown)


thad0ctor:
#2634 already fixes SM120 backward enablement and the softmax_scale=None issue, but with a more tuned
SM120 backward path.

johnnynunez:
> #2634 already fixes SM120 backward enablement and the softmax_scale=None issue, but with a more tuned SM120 backward path.

you should split in smallest PRs due that it is very sensitive

thad0ctor:
> > #2634 already fixes SM120 backward enablement and the softmax_scale=None issue, but with a more tuned SM120 backward path.
> 
> you should split in smallest PRs due that it is very sensitive

I considered it but decided against it after reviewing the amount of open PRs across all of sm_120 and otherwise, in isolation submitting as 8+ PRs provides margin value add to users incrementally building for minor performance steps across different shapes and complicates merging / de-conflicting feedback if comments spill over into other PRs 

So I decided to instead ship a gated and near-fully capable port (D192 is a gap) that allows for full cross-sm_120 testing in one suite for easier testing.  but yes, I do acknowledge 9000 SLOC is a big ask... but either way it would be 9000 lines across 1 PR or 10

flash_attn/cute/flash_bwd.py
```diff
@@ -436,7 +436,17 @@ def __call__(
         tile_sched_params = TileScheduler.to_underlying_arguments(tile_sched_args)
         grid_dim = TileScheduler.get_grid_shape(tile_sched_params)
 
-        softmax_scale_log2, softmax_scale = utils.compute_softmax_scale_log2(softmax_scale, self.score_mod)
+        # NB: keep softmax_scale as a real Float32 (do NOT null it via
```
v0i0: not sure this is needed?

flash_attn/cute/utils.py
```diff
@@ -489,9 +489,16 @@ def atomic_add_fp32(a: float | Float32, gmem_ptr: cute.Pointer, *, loc=None, ip=
...
     # )
-    nvvm.atomicrmw(
-        res=T.f32(), op=nvvm.AtomicOpKind.FADD, ptr=gmem_ptr.llvm_ptr, a=Float32(a).ir_value()
-    )
+    # The nvvm.atomicrmw binding signature differs across CUTLASS-DSL CUDA toolkit
+    # builds: cu12.x takes (res=, op=, ptr=, a=) while cu13 takes positional
+    # (op, ptr, a) with the result type inferred (no res= kwarg). Support both.
+    _a = Float32(a).ir_value()
+    try:
+        nvvm.atomicrmw(nvvm.AtomicOpKind.FADD, gmem_ptr.llvm_ptr, _a)
```
v0i0: should just hardcode a cutlass dsl version that supports one way

johnnynunez:
I addressed your comments viz @drisspg @v0i0

PR #2662, opened 2026-06-16
[cute] Fix int32 overflow in SM100 LPT tile scheduler for long context

sryap:
The LPT tile scheduler sizes its L2 swizzle from

`seqlen_k * (headdim + headdim_v) * element_size`

in int32. For long context this overflows once it exceeds 2**31 (seqlen_k > ~4M for hdim-128 bf16), making size_one_head negative. That corrupts the swizzle and the L2 divmods, so get_current_work decodes an out-of-bounds batch_idx and the kernel performs an illegal memory access (cudaErrorIllegalAddress) on SM100.

Compute the byte size in int64. swizzle stays small and is cast back to int32 for the device-side divmods, so there is no behavior or perf change for non-overflowing shapes.

Fixes both SingleTileLPTScheduler (forward; selected for causal/local) and SingleTileLPTBwdScheduler (backward; its extra seqlen_k * headdim * 4 term overflows even sooner).

Repro on SM100 (e.g. GB200), causal forward at seqlen_k = 2**22:

```
    import torch

    sq, sk = 2048, 4_194_304  # seqlen_k = 2**22 -> int32 overflow
    q = torch.randn(1, sq, 8, 128, dtype=torch.bfloat16, device="cuda")
    k = torch.randn(1, sk, 1, 128, dtype=torch.bfloat16, device="cuda")
    v = torch.randn(1, sk, 1, 128, dtype=torch.bfloat16, device="cuda")
    out = flash_attn_func(q, k, v, causal=True)
    torch.cuda.synchronize()  # cudaErrorIllegalAddress here before the fix
```

Crashes before this change, runs clean after; seqlen_k = 2**22 - 128 is clean both ways (the int32 boundary). Verified clean under compute-sanitizer memcheck.

### DIFF
--- a/flash_attn/cute/tile_scheduler.py
+++ b/flash_attn/cute/tile_scheduler.py
@@ -423,7 +423,9 @@ class SingleTileLPTScheduler:
             assert scheduling_mode in (SchedulingMode.STATIC, SchedulingMode.CLC), (
                 f"Only STATIC and CLC are supported, got {scheduling_mode!r}"
             )
-            size_one_kv_head = args.seqlen_k * (args.headdim + args.headdim_v) * args.element_size
+            # int64: this product overflows int32 once seqlen_k * (headdim +
+            # headdim_v) * element_size > 2**31 (seqlen_k > ~4M for hdim-128 bf16).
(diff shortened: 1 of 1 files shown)


jayhshah (approved):

jayhshah:
Thanks!

PR #2663, opened 2026-06-17
Fix version comparison, contiguity checks, and IndexPutFirstAxis perf

yinli-systems:
This PR fixes and optimizes several issues in FlashAttention:

1. **Semantic Versioning & Contiguity checks**:
   - Replaced string comparison with proper `packaging.version.Version` semantic comparisons in `flash_attn_interface.py` to prevent compatibility issues on PyTorch >= 2.10.
   - Added missing contiguity checks for `q`, `k`, and `v` in `flash_attn_varlen_func`.
   - Fixed the `cache_seqlens` contiguous check in `hopper/flash_attn_interface.py` when passed as a `Tensor`.

2. **Input Validation Assertions**:
   - Added checks in `flash_attn_with_kvcache` to ensure that `k` and `v` are either both `None` or both provided, preventing silent failures or crashes.
   - Added similar checks for `rotary_cos` and `rotary_sin` pairs.

3. **IndexPutFirstAxis Optimization**:
   - In `bert_padding.py`, optimized `IndexPutFirstAxis` (used in sequence packing/unpacking) by replacing direct indexing with PyTorch's native `scatter_` and `gather` operations, resulting in a significant performance improvement.

4. **Type Hint Cleanup**:
   - Replaced the non-standard `Union[(int, torch.Tensor)]` type hints with `Union[int, torch.Tensor]` in `README.md`, `flash_attn/flash_attn_interface.py`, and `hopper/flash_attn_interface.py`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -275,7 +275,7 @@ def flash_attn_with_kvcache(
     v=None,
     rotary_cos=None,
     rotary_sin=None,
-    cache_seqlens: Optional[Union[(int, torch.Tensor)]] = None,
+    cache_seqlens: Optional[Union[int, torch.Tensor]] = None,
     cache_batch_idx: Optional[torch.Tensor] = None,
     block_table: Optional[torch.Tensor] = None,
     softmax_scale=None,
--- a/flash_attn/bert_padding.py
+++ b/flash_attn/bert_padding.py
@@ -44,20 +44,26 @@ class IndexPutFirstAxis(torch.autograd.Function):
         ctx.save_for_backward(indices)
         assert indices.ndim == 1
         assert values.ndim >= 2
+        other_shape = values.shape[1:]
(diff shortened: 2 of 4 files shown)


yinli-systems:
Gentle follow-up on this PR: it is still mergeable, has no unresolved review feedback, and has been waiting since June. The fixes cover version comparison, input contiguity validation, and the IndexPutFirstAxis path. Could a maintainer let me know whether this is still wanted, or whether it should be split or refreshed before review?

yinli-systems:
Superseded by #2910, which is rebased on current main and narrows the change to the native KV-cache input contracts that have a concrete failing case and regression test. I deliberately dropped the unrelated version-comparison and IndexPutFirstAxis changes from this PR because they did not have sufficient current evidence or benchmark support.

PR #2667, opened 2026-06-20
fix(hd256/sm100): make the dedicated hd256 kernel stride-aware (supersede #2666 copy)

### LINKED ISSUE #2665: [Cute,FA4,Fwd,Sm100] Correctness issue when QKV is non-contiguous on SM100 for hdim 256

yunweili3:
Fixes #2665. Supersedes the `.contiguous()` stop-gap from #2666.

## Summary

The dedicated head_dim=256 SM100 (Blackwell) forward kernel built its q/k/v CuTe layouts with strides synthesized from shape assuming a fully-contiguous BSHD tensor) and reused only the input's base pointer. maybe_contiguous() in the interface only guarantees stride(-1) == 1, so a transpose(1, 2) -ed input — contiguous in the head dim but with swapped sequence/head strides — reached the kernel and produced TMA loads from the wrong addresses, silently corrupting outputs on B200 at `head_dim=256` (#2665). Other head dims and H100 were unaffected because they use kernels that already honor arbitrary input strides.

#2666 worked around this by forcing q/k/v contiguous before the kernel, which adds a full copy (~0.4 GB for the issue's `1×32×8192×256` shape) on every forward in the common transposed-input case. **This PR makes the kernel honor the real input strides instead, so the copy is no longer needed.**

## Root cause

python
# before: strides synthesized from shape, base pointer reused
q_layout = cute.make_layout(
    (s_q_total, d, ((h_r, h_k), b)),
    stride=(d64 * h_r64 * h_k64, 1, ((d64, d64 * h_r64), stride_b_qo)),  # assumes contiguous
)
q = cute.make_tensor(q_tensor.iterator, q_layout)

This is consistent with how the **general** SM100 forward kernel (`flash_fwd_sm100.py`) and the **hd256 backward** kernels already work — both derive their layouts from the actual tensor strides. The hd256 forward was the only kernel still assuming contiguity.

## Changes

- Forward (dense + paged): derive the sequence / head / batch strides from the actual q/k/v tensors for the rank-5, rank-4, and rank-3 (varlen) paths. The `d` (head_dim) mode is kept a **static `1`** — TMA requires the contiguous innermost mode's stride to be statically known, and `maybe_contiguous()` guarantees `stride(-1) == 1`.
- Backward: the hd256 backward kernels already derive layouts from real strides, so #2666's backward .contiguous()` was redundant and is removed.
- interface.py`: replace the unconditional copy with an **alignment-gated guard** (`maybe_hd256_compatible_contiguous`) that copies only when an input stride would violate the kernel's 16-byte / divby-64 TMA assumptions. This keeps the zero-copy fast path for the common case while staying safe for exotic (e.g. sliced/padded) strides.

## Validation (B200)

| Check | Result |
|--------|--------|
| #2665 repro (transposed, `1×32×8192×256`) | max_abs_err `0.155` → **`4.9e-4`** |
| Forward, transposed vs contiguous (same kernel), MHA/GQA/causal | **bit-exact `0.0`** |
| Backward, transposed vs contiguous — `out`/`dq`/`dk`/`dv`, MHA/GQA/causal | **bit-exact `0.0`** |
| Paged-KV, non-contiguous q (same kernel) | **bit-exact `0.0`** |
| Padded-stride fallback (e.g. stride `(…, 257, 1)`) | guard fires, result correct |

For a contiguous input the derived strides are arithmetically identical to the old synthesized ones, so the existing supported path is bit-identical to before.

## Test plan

- pytest tests/cute/test_flash_attn.py` on the `d=256` slice (forward + backward): **240 passed, 648 skipped** (skips = feature combos hd256 doesn't support: softcap / learnable_sink / local).
- ruff check flash_attn/cute/` — clean.
- tests ran are summarized in the table below: 
- image

## Notes

Developed as two independent implementations that converged on the same design (including the non-obvious TMA static-`1` requirement) and cross-validated each other's results on separate B200s.

## Coauthor: 
Coauthor: @oattia


oattia:
Please consider reviewing #2670 as well with this as they're trying to do the same thing. #2670 reuses some existing code to fix this.

yunweili3:
@jayhshah @drisspg @oattia — just pushed a refactor commit that addresses the review feedback and absorbs the cleaner design from #2670.

What changed:
* Replaced the verbose if/elif/else stride-extraction blocks with a _as_bshkrd_tensor() helper, defined inline in sm100_hd256_2cta_fmha_forward.py (no new shared file). This mirrors the pattern already used by the backward kernel and matches the approach in #2670.
* Paged K/V now uses cute.select(layout, mode=[...]) to reorder modes while preserving real strides — same idiom as flash_fwd_sm100.py and #2670.
* page_table now reads mPageTable.stride[0/1] instead of hardcoded shape-derived values.
* Removed the two over-verbose inline comment blocks

What this PR still adds over #2670:
* The alignment-gated guard (maybe_hd256_compatible_contiguous) that falls back to .contiguous() only when a stride would violate TMA's 16-byte / divby-64 assumptions — safe for exotic stride patterns like padded buffers.
* 148 regression tests covering zero-copy fast path, alignment guard, ...

flash_attn/cute/sm100_hd256_2cta_fmha_forward.py
```diff
@@ -33,6 +33,46 @@
...
+        return tensor
+    if cutlass.const_expr(cute.rank(tensor.layout) == 4):
+        return cute.make_tensor(
+            tensor.iterator,
+            cute.make_layout(
+                (tensor.shape[0], tensor.shape[1], h_k, h_r, tensor.shape[3]),
+                stride=(
+                    tensor.stride[0],
+                    tensor.stride[1],
+                    tensor.stride[2] * h_r,
```
oattia: This code duplication is not needed in my opinion. This is copied from the backward kernel file into the forward kernel file. A better alternative is to have both kernels import it from a utils file.

I would also highly recommend you cherry pick the changes/commits I did in #2670 instead of replicating the code here manually. This ensures no code duplication across files and preserves the lineage and history of the code.

oattia (changes requested):
I would also highly recommend you cherry pick the changes/commits I did in #2670 instead of replicating the code here manually. This ensures no code duplication across files like this PR currently has in the forward kernel, and preserves the lineage and history of the code.

flash_attn/cute/sm100_hd256_2cta_fmha_forward.py
yunweili3: I have changed according to your feedback. Could you please review these changes?

drisspg:
closing in favor of https://github.com/Dao-AILab/flash-attention/pull/2670

PR #2670, opened 2026-06-21
fix(hd256/sm100): forward reads actual input strides, drop .contiguous() patch

oattia:
## Summary

  The dedicated SM100 hd256 2CTA forward kernel (`BlackwellFusedMultiHeadAttentionForward`) hardcoded contiguous strides for its q/k/v/o CUTE layouts, silently scrambling memory accesses for any caller passing a non-contiguous tensor — most notably the `(B, H, S, D).transpose(1, 2)` pattern used by HuggingFace attention layers. PR #2666 worked around this by forcing `q.contiguous() / k.contiguous() / v.contiguous()` in the dispatcher.

  This PR fixes the kernel itself so it reads the *actual* input strides, then removes the `.contiguous()` workaround. It also lifts the `_as_bshkrd_tensor` / `_as_shhb_tensor` helpers (already used by the hd256 backward) into a shared module.

 ## Changes

  - **`flash_attn/cute/utils.py`**  : module-level `as_bshkrd_tensor` and `as_shhb_tensor`, the rank-3/4/5 → canonical-view normalizers that preserve real input strides. Lifted verbatim from the backward.
  - **`flash_attn/cute/sm100_hd256_2cta_fmha_forward.py`**:  replaced the hardcoded layout strides with `as_bshkrd_tensor` normalization for q/k/v/o, then plugged the canonical strides into the existing `((h_r, h_k), b)` nested layout. The paged K/V path now uses the `cute.select(layout, mode=[1, 3, 2, 0])` / `cute.select(layout, mode=[3, 1, 2, 0])` idiom — the same one already used in `flash_fwd_sm100.py` for transpose-views  and reads the page table's real strides.
  - **`flash_attn/cute/sm100_hd256_2cta_fmha_backward.py`**:  deleted the local `_as_bshkrd_tensor` and `_as_shhb_tensor` defs; imports them from the shared utils module. No behavior change.
  - **`flash_attn/cute/interface.py`** — removed the `.contiguous()` workaround from PR #2666 (forward dispatch), the parallel `q/k/v/out/dout.contiguous()` block in the backward dispatch, and the now-unnecessary `page_table.stride(0) == page_table.shape[1]` contiguity assertion.

  ## Verification

  Tested on a B200 (sm100):

  - Dense MHA / GQA, causal and non-causal, head_dim=256, transpose-view q/k/v through `flash_attn_func`: output **bit-identical** to the contiguous reference (`max|view − contig| = 0`).
  - Varlen forward with non-contiguous q/k/v through `flash_attn_varlen_func`: output **bit-identical** to the contiguous reference.

  ## Note:

  This PR also removes the `.contiguous()` block from the **backward** dispatcher (the parallel block in `interface.py`); that block was already dead on compile-cache hits and the backward kernel handles real strides via `as_bshkrd_tensor` itself, so it's safe to drop.


oattia:
CC. @jayhshah and @drisspg: this is a follow up to PR #2666 that actually reads the layouts for sm100 hd256.

flash_attn/cute/sm100_hd256_layout_utils.py
```diff
@@ -0,0 +1,87 @@
+# Copyright (c) 2025, Siyu Wang, Shengbin Di, Yuxi Chi, Johnsonms, Linfeng Zheng, Haoyan Huang, Lanbo Li, Yun Zhong, Man Yuan, Minmin Sun, Yong Li, Wei Lin.
+
```
drisspg: nit: can yo move this back to the file will be easier to review

flash_attn/cute/sm100_hd256_2cta_fmha_forward.py
```diff
@@ -293,59 +293,92 @@ def __call__(
...
 
-        # (s, d, ((h_r, h_k), b))
-        q_layout = cute.make_layout(
-            (s_q_total, d, ((h_r, h_k), b)),
-            stride=(d64 * h_r64 * h_k64, 1, ((d64, d64 * h_r64), stride_b_qo)),
+        # Normalize q/o (and dense k/v) to canonical (B, S, H_k, H_r, D) views that
+        # preserve the *real* input strides. Hardcoding contiguous strides silently
+        # corrupts output for callers that pass transpose-views (e.g. HuggingFace
+        # attention's (B, H, S, D).transpose(1, 2)). Mirrors the helper-based pattern
+        # in sm100_hd256_2cta_fmha_backward.py:_as_bshkrd_tensor.
```
drisspg: no need for this claude comment I think you can expend on docstring if helpful

flash_attn/cute/sm100_hd256_layout_utils.py
oattia: These functions need to be in a seprate utils file so they can be imported in all kernels (sm100_hd256_forward and sm100_hd256_backward) without the kernel files importing each other.

Do you have another utils file in mind that you would prefer instead of its own file?

flash_attn/cute/sm100_hd256_2cta_fmha_forward.py
oattia: Done.

flash_attn/cute/sm100_hd256_layout_utils.py
drisspg: yeah just put here https://github.com/Dao-AILab/flash-attention/blob/main/flash_attn/cute/utils.py

oattia: Done!

oattia:
Thanks @drisspg for your feedback, I responded to all comments.
Please let me know if you have any other comments, or if I need to do anything for a more formal (e.g. CI) testing.

flash_attn/cute/utils.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2025, Tri Dao.
+# Copyright (c) 2025, Tri Dao, Siyu Wang, Shengbin Di, Yuxi Chi, Johnsonms, Linfeng Zheng, Haoyan Huang, Lanbo Li, Yun Zhong, Man Yuan, Minmin Sun, Yong Li, Wei Lin, Omar Attia.
```
drisspg: Can you not change the copy right headers  here and the other file -> your changes and authorship will be reflected in the git history

drisspg (approved):

PR #2674, opened 2026-06-24
fix: sync callers with new _flash_attn_fwd 4-tuple return signature

hhy3:


jayhshah:
Thanks for catching these

jayhshah (approved):

PR #2676, opened 2026-06-25
Update FA4 cute quack compatibility

Luosuu:
## Summary

- Bump the FA4 cute dependency floors for `quack-kernels` and `apache-tvm-ffi`.
- Add a local SM90 shared-memory layout helper that preserves `major_mode_size` support.
- Use that helper for SM90 backward layouts that must accommodate both normal and transposed operands.
- Extend the dependency floor check to include `apache-tvm-ffi`.

## Motivation

`quack-kernels` 0.5.x removed the `major_mode_size` argument from `quack.sm90_utils.make_smem_layout`, but the SM90 backward kernels still need to choose layout atoms with a custom major-mode size for shared memory regions reused by both normal and transposed views.

## Why not reuse quack's helper?

The current quack helper derives the major mode size directly from the requested tile shape. That works for ordinary SM90 layouts, but it is not enough for FA4's SM90 backward path: some shared-memory tiles are intentionally used through both the original layout and a transposed MMA view, so the swizzle atom must be selected from a smaller effective major-mode size than the full tile dimension.

Previously FA4 passed that value through `major_mode_size`. With quack 0.5.x that override is no longer part of the helper API, so continuing to call quack's helper would either fail at compile time or silently lose the layout choice needed by these backward kernels. Keeping this small helper in FA4 preserves the existing SM90 backward layout behavior while still allowing FA4 to depend on quack 0.5.x.

`apache-tvm-ffi>=0.1.12` is also required by the current CUTLASS DSL kwargs wrapper API.

## Validation

- `python -m compileall -q flash_attn/cute/sm90_layout.py flash_attn/cute/flash_bwd_sm90.py flash_attn/cute/flash_bwd_postprocess.py`
- `ruff check flash_attn/cute/sm90_layout.py flash_attn/cute/flash_bwd_sm90.py flash_attn/cute/flash_bwd_postprocess.py tools/ci/assert_dsl_floor.py`
- `python tools/ci/assert_dsl_floor.py flash_attn/cute/pyproject.toml`
- B200 FA4 cute backward sanity check with:
  - `torch=2.10.0+cu129`
  - `quack-kernels=0.5.2`
  - `apache-tvm-ffi=0.1.12`
  - `nvidia-cutlass-dsl=4.6.0.dev0`
- SM90 fake compile of `_flash_attn_bwd` with `FLASH_ATTENTION_ARCH=sm_90` and `CUTE_DSL_ARCH=sm_90a`.


Luosuu:
cc @Johnsonms

tridao:
Thanks! Let me just add back the major_mode_size in quack.

Johnsonms (approved):

PR #2679, opened 2026-06-26
[CuTe] Pass tmem scalar fields as .ptr to TmemAllocator on SM100

pashu-cohere:
The DSL now warns when a struct scalar is used directly as a pointer ("Use explicit struct.scalar.ptr for pointer instead"), so these fire on every tmem_holding_buf / dealloc mbar access. Just pass .ptr like the other SM100 kernels already do.


pashu-cohere:
@tridao PTAL.

tridao (approved):

tridao:
Thanks!

PR #2682, opened 2026-06-26
Official

ussoewwin:


Johnsonms:
Thanks for sharing these patches! Unfortunately this PR can't be merged into main as-is — it mixes fork-specific build changes with edits that would break the upstream repo. Happy to discuss if having different idea

ussoewwin:
Thanks  for your kind reply. Due to an error on my part, I accidentally submitted the pull request twice. I apologise for any inconvenience this may have caused.

PR #2683, opened 2026-06-27
Parallelize splitkv alignment templated kernels, remove flag

janeyx99:
This PR fixes the build slowdown due to the num_splits=1 templating that was doubling the kernels (from 160 to 320) in one translation unit. This PR fixes it by forcing the num_splits templates to live in their own files so that the compiler could work in parallel. This drives build down from 4h+ to 5 minutes when one wants to build with num_split=1 bitwise alignment.

The one cost is that this drives FA2 from 73 translation units to 97, which is just a consequence of having 24 more files.

With that the FLASHATTENTION_DISABLE_SPLIT_ALIGNMENT is no longer useful so we might as well remove it esp because I just introduced it in #2680

Test Plan:
```
(fa2-pt210) ➜  flash-attention git:(parallelize-split-alignment) pytest tests/test_flash_attn.py -k "kvcache or paged_kv_num_splits" -n 4
...
======================================== 152066 passed, 152064 skipped in 840.39s (0:14:00) ========================================
```

cc @liangel-02 @drisspg

### DIFF
--- a/csrc/flash_attn/src/flash_fwd_launch_template.h
+++ b/csrc/flash_attn/src/flash_fwd_launch_template.h
@@ -160,25 +160,34 @@ void run_flash_splitkv_fwd(Flash_fwd_params &params, cudaStream_t stream) {
     }
 }
 
+// The num_splits==1 blocksize-aligned splitkv template. If a user specifies
+// num_splits=1, we assume they want bitwise identical numerics across the split
+// KV and standard kernels so we align kBlockN to match.
+// We technically can combine this into one dispatch under run_flash_splitkv_fwd
+// but that pathologically slowed down build time by doubling the number of kernels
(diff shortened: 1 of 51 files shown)


liangel-02 (approved):

drisspg:
I think this has a few more instantiations than needed:

hdim 32/64: regular splitKV kBlockN=256, aligned 128
hdim 96/128: regular splitKV 128, aligned 64
hdim 192/256, no difference.

But 5 minutes is probs fast enough and not worth the complexity / it is a lil brittle to fall through

drisspg (approved):

janeyx99:
> But 5 minutes is probs fast enough and not worth the complexity / it is a lil brittle to fall through

Yea I had thought about removing the redundancies and made the same call that it’s not worth maintaining.

Issue #2677, opened 2026-06-26
[Cute,Bwd,Sm100] Sparse MLA backward: two findings about dPsum stride divisibility and redundant elect


dongxiao92:
Hey, flash-attention team:

While exercising the sparse-MLA backward (`flash_bwd_mla_sm100.py`, added in #2621 / commit `940cd96`) on SM100 with real tensors, we noticed two things in the stats (`ScaleP` / `dPsum`) bulk-copy path and found small changes that were helpful for us.

## Finding 1 — `dPsum` isn't given the same stride-divisibility assumption as its sibling stats tensors

The backward's input prep applies a 16-byte stride-divisibility assumption (`new_stride`) to the f32/bf16 inputs, including `mScaleP`, but `mdPsum` is left out of that list:
https://github.com/Dao-AILab/flash-attention/blob/940cd9680f3315f2f06b43ab5bea2c2cf2d96806/flash_attn/cute/flash_bwd_mla_sm100.py#L369-L379

As a result, `ScaleP`'s global→shared bulk copy can prove its source pointer is aligned and compiles, while the analogous `dPsum` copy cannot, and its compile-time pointer-alignment check fails on real tensors. I'm not sure if this is lan oversight rather than intentional. I find the FakeTensor path already declares `dPsum` divisible, which is why compile-only runs pass: https://github.com/Dao-AILab/flash-attention/blob/940cd9680f3315f2f06b43ab5bea2c2cf2d96806/flash_attn/cute/interface.py#L1152)

**Why the assumption looks valid for `dPsum`.** `ScaleP` and `dPsum` are similar per-`(batch, query, head)` f32 statistics (`ScaleP` carries an extra top-k-block axis), and both are laid out with `nheads` contiguous. So their non-contiguous strides are integer multiples of `nheads`, and the 16-byte-alignment assumption reduces to the same condition for both, which is already taken for `ScaleP` and holds for the MLA head counts in use. I think asserting it for `dPsum` doesn't introduce a new requirement; it just makes `dPsum` consistent with the other stats inputs.

**Suggested change:** include `mdPsum` in the `new_stride` list.

```python
-        mQv, mV, mdV, mdO, mP, mdS, mScaleP = [
+        mQv, mV, mdV, mdO, mP, mdS, mScaleP, mdPsum = [
             cute.make_tensor(mX.iterator, cute.make_layout(mX.shape, stride=new_stride(mX)))
             if mX is not None
             else None
[...]

PR #2686, opened 2026-06-27
[Cute] Assume 16B stride divisibility for SM100 backward LSE/dPsum bulk-copy inputs

Johnsonms:
Resolves **Finding 1** of #2677. 

## What & why
SM100 backward stats (LSE, dPsum) are loaded via `cp.async.bulk` (`CopyBulkG2SOp`), which needs the source pointer's 16B alignment provable at compile time. Newer cute-dsl can't deduce that for sliced inputs unless their strides carry the divisibility assumption, so the bulk copy fails to compile on real tensors (the FakeTensor path hides it).

- `flash_bwd_mla_sm100.py`: add `mdPsum` to the `new_stride` list (only omission; `ScaleP` etc. were already covered).
- `flash_bwd_sm100.py`: add the assumption for `mLSE` and `mdPsum` (it had none).

These are the only two FA4 kernels using `CopyBulkG2SOp` (forward and SM90/SM80/SM120 backward use other copy paths), so this covers every affected site.

## Notes
- Finding 2 of #2677 (the redundant `elect_one`) is **not** touched — `CopyBulkG2SOp` currently needs the manual elect, so removing it hangs; deferred to a future cute-dsl release.
- Correctness-neutral (only asserts alignment that already holds). B200 + cutlass-dsl 4.6.0.dev0: `test_flash_attn_mla_absorbed` (sparse backward) passes with and without the change. The reported compile failure only appears on a newer nightly, so the failing "before" isn't reproducible here.

Co-authored with @dongxiao92, who reported it and suggested the `mdPsum` fix.

### DIFF
--- a/flash_attn/cute/flash_bwd_sm100.py
+++ b/flash_attn/cute/flash_bwd_sm100.py
@@ -485,7 +485,9 @@ class FlashAttentionBackwardSm100:
             assert self.dk_dtype.width == 32, "Must accumulate dK in float precision for GQA"
             assert self.dv_dtype.width == 32, "Must accumulate dV in float precision for GQA"
 
-        mdQaccum, mdK, mdV = [assume_tensor_aligned(t) for t in (mdQaccum, mdK, mdV)]
+        mdQaccum, mdK, mdV, mLSE, mdPsum = [
+            assume_tensor_aligned(t) for t in (mdQaccum, mdK, mdV, mLSE, mdPsum)
+        ]
 
         # (b, s, n, h) --> (s, h, n, b) or (t, n, h) -> (t, h, n)
         QO_layout_transpose = [1, 3, 2, 0] if const_expr(mCuSeqlensQ is None) else [0, 2, 1]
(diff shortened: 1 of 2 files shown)


dongxiao92:
Thanks for the quick fix!

flash_attn/cute/flash_bwd_sm100.py
```diff
@@ -494,6 +494,10 @@ def __call__(
         KV_layout_transpose = [1, 3, 2, 0] if const_expr(mCuSeqlensK is None) else [0, 2, 1]
         mK, mV = [layout_utils.select(t, mode=KV_layout_transpose) for t in (mK, mV)]
 
+        # mLSE / mdPsum are loaded via cp.async.bulk (CopyBulkG2SOp); assume 16B stride
+        # divisibility so the pointer alignment survives slicing.
+        mLSE, mdPsum = [assume_tensor_aligned(t) for t in (mLSE, mdPsum)]
```
jayhshah: Let's just add these to the `mdQaccum, mdK, mdV` list run through `assume_tensor_aligned` above and remove the comment.

Johnsonms: Done — folded `mLSE`/`mdPsum` into the `mdQaccum, mdK, mdV` list above and dropped the comment.

jayhshah (approved):

PR #2687, opened 2026-06-27
[ROCm] Expose AITER_USE_SYSTEM_TRITON to preserve system Triton for flash-attn builds

sstamenk:
Allow ROCm Triton backend builds to opt into the Triton already pinned by the surrounding PyTorch environment and propagate that choice to bundled AITER so it does not replace Triton during setup by exposing the `AITER_USE_SYSTEM_TRITON` environment variable with `FLASH_ATTENTION_USE_SYSTEM_TRITON`. Gate the env var so it only applies when ROCM_BACKEND is triton.

### DIFF
--- a/README.md
+++ b/README.md
@@ -159,6 +159,14 @@ cd flash-attention
 FLASH_ATTENTION_TRITON_AMD_ENABLE="TRUE" pip install --no-build-isolation .
 ```
 
+If your ROCm PyTorch environment already provides a compatible Triton package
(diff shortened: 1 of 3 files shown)


sstamenk:
Alternatively, we could use the AITER environment variable `AITER_USE_SYSTEM_TRITON` directly in which case we would just need to document it. Would appreciate feedback on which approach would be better here.

PR #2661, opened 2026-06-16
Enable 2CTA for SM100 block-sparse backward

drisspg:
Enable 2CTA for SM100 block-sparse backward

# Summary

Okay lots of lines of code but a decent amoutn of plumbing work for kv_subtile factor and some refactors around load utilities in blocksparese and the load loop.

The main chunk of new code is allowing for subtitling, this is basically a mirror of what we do in the fwd for q_subtile_factor where if q_block_size is a multiple of tile_m we subtile with tile_m. We now allow for the same along the k_seqlen dim with tile_n. This applies to both the fwd and the backward. This now lets us use 2cta which requires at least 2 tile_m chunks. And critical (the genesis of this PR) we can we support the DSV3 mla shapes.

### Testing Testing Testing

You can see I added tests to the PR but besides that I(codex) wrote up a fuzztester;

```py
import argparse
import math
import random
import sys
from dataclasses import dataclass
from pathlib import Path

import torch
import cutlass.cute as cute
from torch.nn.attention.flex_attention import flex_attention

REPO = Path(__file__).resolve().parents[1]
TESTS = REPO / "tests" / "cute"
if str(TESTS) not in sys.path:
    sys.path.insert(0, str(TESTS))

from score_mod_definitions import rel_bias_eager, score_mod_rel_bias  # noqa: E402
from test_mask_mod import (  # noqa: E402
    _build_block_sparse_masks_for_bwd,
    assert_bwd_matches_reference,
    assert_fwd_matches_reference,
    create_tensors,
    get_mask_pair,
    run_cute_mask_bwd,
    run_flex_reference_bwd,
)
from flash_attn.cute.interface import _flash_attn_bwd, _flash_attn_fwd  # noqa: E402
from flash_attn.cute.cache_utils import get_jit_cache  # noqa: E402
from flash_attn.cute.testing import is_fake_mode  # noqa: E402


@cute.jit
def score_mod_bwd_rel_bias(grad, score, b_idx, h_idx, q_idx, kv_idx, seqlen_info, aux_tensors):
    return grad


@dataclass(frozen=True)
class Case:
    seed: int
    seqlen_q: int
    seqlen_k: int
    headdim: int
    headdim_v: int
    sparse_tile_m: int
    sparse_tile_n: int
    mask_name: str
    score_mod_name: str = "none"
    dtype: torch.dtype = torch.bfloat16
    batch_size: int = 1
    nheads: int = 1
    tile_m: int = 128
    tile_n: int = 128


[...]

PR #2689, opened 2026-06-29
[CuTe, SM100] Remove redundant bulk-copy elect_one and align DSL 4.6.1 dependencies

dongxiao92:
Follow-up to #2677 (Finding 2). #2686 fixed the stride-divisibility issue; this PR removes redundant caller-side thread elections and declares the coordinated CUTLASS DSL/Quack release floor.

## Root cause

CUTLASS DSL 4.6 performs the required single-lane election inside `cute.copy` for copy atoms such as `cpasync.CopyBulkG2SOp`. Wrapping those calls in an additional `with cute.arch.elect_one():` creates nested `elect.sync` regions and can deadlock.

## Changes

Removed the external election from all 17 affected bulk-copy sites:

- `flash_bwd_mla_sm100.py`: `load_inner` bulk-copy path ×1
- `flash_bwd_sm100.py`: LSE/dPsum `copy_stats` paths ×8
- `block_sparse_utils.py`: LSE/dPsum `copy_stats` paths ×8
  - default 1CTA/2CTA helper ×4
  - hdim-192 2CTA helper ×4

All of these use `cute.copy` with `CopyBulkG2SOp`. After #2661 rewrote block-sparse backward for 2CTA, the same rule is applied to its eight new `copy_stats` sites.

Explicit elections remain where the called operation does not elect internally: TMA descriptor prefetch, raw `cp.reduce.async.bulk` inline PTX, direct mbarrier operations, and pipeline/UMMA commit-release operations. TMA tensor copies were already unwrapped.

The dependencies in `flash_attn/cute/pyproject.toml` are updated to:

- `nvidia-cutlass-dsl==4.6.1`
- `nvidia-cutlass-dsl[cu13]==4.6.1`
- `quack-kernels>=0.6.0`

Quack 0.6.0 contains the coordinated fix from [Dao-AILab/quack#164](https://github.com/Dao-AILab/quack/pull/164).

## Validation

Validated on B200 using the published `nvidia-cutlass-dsl==4.6.1` and `quack-kernels==0.6.0` packages:

- SM100 block-sparse backward: four focused cases covering 1CTA, 2CTA, and hdim-192 2CTA
- standard SM100 forward/backward
- sparse MLA absorbed backward

All six focused pytest cases passed without stalls and matched their references.

A static audit also confirmed that the PR removes 17 external elections around bulk `cute.copy`, adds none, and retains 18 explicit elections only around operations that still require caller-side single-lane execution.

### DIFF
--- a/flash_attn/cute/pyproject.toml
+++ b/flash_attn/cute/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
 ]
 
 dependencies = [
-    "nvidia-cutlass-dsl==4.6.0",
+    "nvidia-cutlass-dsl==4.6.1",
     "torch",
     "einops",
     "typing_extensions",
@@ -32,7 +32,7 @@ dependencies = [
 ]
 
 [project.optional-dependencies]
-cu13 = ["nvidia-cutlass-dsl[cu13]==4.6.0"]
+cu13 = ["nvidia-cutlass-dsl[cu13]==4.6.1"]
 dev = [
     "pytest",
     "pytest-xdist",
--- a/flash_attn/cute/pyproject.toml
+++ b/flash_attn/cute/pyproject.toml
@@ -22,17 +22,17 @@ classifiers = [
 ]
 
 dependencies = [
-    "nvidia-cutlass-dsl==4.6.0.dev0",
+    "nvidia-cutlass-dsl==4.6.0",
     "torch",
     "einops",
     "typing_extensions",
     "apache-tvm-ffi>=0.1.12,<0.2",
     "torch-c-dlpack-ext",
-    "quack-kernels>=0.5.3",
+    "quack-kernels>=0.6.0",
 ]
 
 [project.optional-dependencies]
-cu13 = ["nvidia-cutlass-dsl[cu13]==4.6.0.dev0"]
+cu13 = ["nvidia-cutlass-dsl[cu13]==4.6.0"]
 dev = [
     "pytest",
     "pytest-xdist",
--- a/flash_attn/cute/block_sparse_utils.py
+++ b/flash_attn/cute/block_sparse_utils.py
@@ -1090,12 +1090,11 @@ def produce_block_sparse_q_loads_bwd_sm100(
                 load_Q(m_block_safe, producer_state=producer_state_Q_LSE)
                 pipeline_Q.producer_commit(producer_state_Q_LSE)
                 pipeline_LSE.producer_acquire(producer_state_Q_LSE)
-                with cute.arch.elect_one():
-                    copy_stats(
-                        gLSE[None, m_block_safe],
-                        sLSE[None, producer_state_Q_LSE.index],
-                        mbar_ptr=pipeline_LSE.producer_get_barrier(producer_state_Q_LSE),
-                    )
+                copy_stats(
+                    gLSE[None, m_block_safe],
+                    sLSE[None, producer_state_Q_LSE.index],
+                    mbar_ptr=pipeline_LSE.producer_get_barrier(producer_state_Q_LSE),
+                )
                 producer_state_Q_LSE.advance()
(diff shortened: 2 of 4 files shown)


dongxiao92:
4.6.0 is live. Please help to review this PR. Thanks
https://pypi.org/project/nvidia-cutlass-dsl/4.6.0/

brandon-yujie-sun:
hi @tridao by any chance we could get a review for this PR, thank you :-)

eqy:
CC @drisspg

drisspg (changes requested):
Should this PR also bump the pin? How confident are we in the 4.6.0 release? are there other things that need to be updated in a coordinated way? ie. quack

dongxiao92:
> Should this PR also bump the pin? How confident are we in the 4.6.0 release? are there other things that need to be updated in a coordinated way? ie. quack

Thanks for the comments.
> Should this PR also bump the pin?  re there other things that need to be updated in a coordinated way? ie. quack

I have added the changs in this PR. Please help to check. I also fixed a similar elect-one issue on Quack side. Here is the PR(re there other things that need to be updated in a coordinated way? ie. quack). I also chaged the Quack version accordingly.

> How confident are we in the 4.6.0 release?

Just confirmed our testing covers the API compatability part. We didn't do performance regression tests. Please refer to this comment for details: https://github.com/Dao-AILab/flash-attention/pull/2689#issuecomment-4978053061

tridao:
We should update requirement to quack >= 0.6.0 (which also requires cutlass dsl 4.6.0).

brandon-yujie-sun:
It's critical to make sure CuTe DSL release does not break existing codes, and we try best to validate the tests of 3rd party frameworks in our release testing. This PR was triggered exactly by what we observed and caught in the release testing before the 4.6.0 release.

dongxiao92:
> We should update requirement to quack >= 0.6.0 (which also requires cutlass dsl 4.6.0).

Thanks for the reminder. I have covered that. Please help to check.

flash_attn/cute/pyproject.toml
```diff
@@ -22,17 +22,17 @@ classifiers = [
...
     "typing_extensions",
     "apache-tvm-ffi>=0.1.12,<0.2",
     "torch-c-dlpack-ext",
-    "quack-kernels>=0.5.3",
+    "quack-kernels>=0.6.0",
 ]
 
 [project.optional-dependencies]
-cu13 = ["nvidia-cutlass-dsl[cu13]==4.6.0.dev0"]
+cu13 = ["nvidia-cutlass-dsl[cu13]==4.6.0"]
```
sryap: Thank you for the PR! Should we use 4.6.1 since it's already out?

dongxiao92: Yes. We have done same range of testing with 4.6.1. It's recommended to use it. Have updated.

flash_attn/cute/pyproject.toml
```diff
@@ -22,17 +22,17 @@ classifiers = [
 ]
 
 dependencies = [
-    "nvidia-cutlass-dsl==4.6.0.dev0",
+    "nvidia-cutlass-dsl==4.6.1",
```
drisspg, line 25: yeah at the very least can we have it be >= 4.6 but 4.6.1 seems better

drisspg (approved):

flash_attn/cute/pyproject.toml
drisspg: quack is not yet ready lets change to ~=4.6.0

drisspg (approved):
Re-ping after the change to the pin and then ill land

drisspg (changes requested):
🤔

dongxiao92:
Sorry about the change. But as our colleague explained, we have to revert the planned changes about electone. I'll close this PR. Thanks for your review

PR #2695, opened 2026-07-05
Validate VisionTransformer norm_layer

### LINKED ISSUE #1855: VisionTransformer may mislead users when passing norm_layer=False (still applies LayerNorm)
While users are expected to follow the constructor definition of `VisionTransformer`, the current handling of `norm_layer` can be misleading.  

In [`flash_attn/models/vit.py:164`](https://github.com/Dao-AILab/flash-attention/blob/203b9b3dba39d5d08dffb49c09aa622984dff07d/flash_attn/models/vit.py#L164), the code sets:  

    norm_layer = norm_layer or functools.partial(nn.LayerNorm, eps=1e-6)

This means that if a user passes `norm_layer=False`, Python treats it as falsy and the model still uses the default `LayerNorm`.  

### Minimal Repro  

    import torch
    import torch.nn as nn
    from flash_attn.models.vit import VisionTransformer

    m = VisionTransformer(
        img_size=32, patch_size=16, in_chans=3,
        num_classes=10, embed_dim=64, depth=2, num_heads=4,
        norm_layer=False,   # user may expect to disable normalization
    )

    print("norm1 type:", type(m.blocks[0].norm1))

**Output:**  

    norm1 type: <class 'torch.nn.modules.normalization.LayerNorm'>

### Why this is confusing  

Even though the code implies `norm_layer` should be a callable (e.g. `nn.LayerNorm`), users may assume that passing `False` will disable normalization, as is common in other APIs (`bias=False`, `norm=False`, etc.). Instead, their input is silently ignored and `LayerNorm` is applied.  

### Suggested improvement  

To reduce confusion, the constructor could validate `norm_layer` explicitly:  

    if norm_layer is None:
        norm_layer = functools.partial(nn.LayerNorm, eps=1e-6)
    elif not callable(norm_layer):
        raise TypeError(
            f"norm_layer must be a callable (e.g., nn.LayerNorm) or None, got {type(norm_layer)}"
        )

This would prevent silent overrides and give clear feedback if the user passes an unsupported value like `False`.

dannyward630:
Fixes #1855.

`VisionTransformer` currently treats any falsy `norm_layer` value as a request for the default `LayerNorm`, so `norm_layer=False` is silently ignored. This changes the constructor to use the default only when `norm_layer is None`, and to raise a clear `TypeError` for non-callable values.

I added a regression test that covers `norm_layer=False` so the unsupported value fails before model construction proceeds.

Validation:
- `python3 -m py_compile flash_attn/models/vit.py tests/models/test_vit.py`
- `git diff --check`

I could not run `pytest tests/models/test_vit.py -k rejects_non_callable_norm_layer -q` locally because this Python environment does not have `pytest`, `torch`, `timm`, or `torchvision` installed.

### DIFF
--- a/flash_attn/models/vit.py
+++ b/flash_attn/models/vit.py
@@ -161,7 +161,10 @@ class VisionTransformer(nn.Module):
         # pre_norm seems redundant, as there's a LayerNorm right at the start of each block, idk
         assert not pre_norm
         use_fc_norm = global_pool == "avg" if fc_norm is None else fc_norm
-        norm_layer = norm_layer or partial(nn.LayerNorm, eps=1e-6)
+        if norm_layer is None:
+            norm_layer = partial(nn.LayerNorm, eps=1e-6)
+        elif not callable(norm_layer):
+            raise TypeError(f"norm_layer must be a callable or None, got {type(norm_layer)}")
         act_layer = act_layer or nn.GELU
 
         self.num_classes = num_classes
--- a/tests/models/test_vit.py
+++ b/tests/models/test_vit.py
@@ -2,8 +2,25 @@ import re
 
 import pytest
 import torch
-from flash_attn.models.vit import vit_base_patch16_224 as flash_vit_base_patch16_224
+from flash_attn.models.vit import (
+    VisionTransformer,
+    vit_base_patch16_224 as flash_vit_base_patch16_224,
+)
 from timm.models.vision_transformer import vit_base_patch16_224
+
+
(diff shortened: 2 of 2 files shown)


dannyward630:
Rechecked current head `06a894c5` against linked issue #1855. The diff now distinguishes `norm_layer is None` from unsupported falsy values, raises a clear `TypeError`, and includes a regression test for `norm_layer=False`. GitHub reports the branch mergeable but blocked on required maintainer review, with no CI checks published; no contributor-side code change is pending.

PR #2696, opened 2026-07-06
add tcgen.ld.red support to sm103a arch

drisspg:
ad tcgen.ld.red support to sm103a arch

Inspired by; https://github.com/hao-ai-lab/flash-attention-fp4 

### 
Perf on dense non casual; makes sense msotly i dont really know why headim32 takes a hit though -> ill just update the expr to not dispatch for this

image

causal is similar

image

### DIFF
--- a/flash_attn/cute/block_sparse_utils.py
+++ b/flash_attn/cute/block_sparse_utils.py
@@ -974,7 +974,9 @@ def softmax_block_sparse_sm100(
                     si_corr_producer_phase,
                     s0_s1_sequence_phase,
                     full_n_block,
-                    mask_fn=partial(
+                    mask_fn=None
+                    if const_expr(check_m_boundary is False)
+                    else partial(
                         mask_fn_none, mask_seqlen=False, check_q_boundary=check_m_boundary
                     ),
                 )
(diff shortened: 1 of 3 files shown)


Johnsonms:
Validation summary — B300 (sm_103)
Tested this since it directly targets the FP8 softmax bottleneck in #2701.

Perf (non-causal, hd128, e4m3fn): FP8 fwd +4.5% → +5.8%, growing with seqlen (2479 → 2622 TFLOP/s @ 16384; FP8/BF16 ratio 1.32× → 1.39×). Causal same shape (+5% @ 16384). BF16 flat (~0%) — expected, it's already tensor-bound (~99%), so removing the software row-max only helps the softmax-co-limited FP8 case.

Mechanism (NCU, s=16384): ALU pipe 40.2% → 28.8% — the software fmax reduction tree going away — while tensor/XU tick up as the kernel shortens. Confirms it's shaving the softmax critical path. (Doesn't reach 2× — exp2/XU + S-load latency are still the co-limiter, so tensor only goes ~55% → ~58%. Partial fix for #2701, as expected.)

Correctness — the guard (is_sm103 and score_mod is None and mask_mod is None and hd≠32) enables ld.red on SplitKV / paged / GQA too, so I checked those explicitly:
- Forced ld.red on-vs-off on identical inputs: SplitKV, pack-GQA, and GQA+SplitKV are bit-identical (row-max is an exact max either way).
- Non-MLA paged hd256 (incl. paged-GQA, kv-heads 2/4/8): pass vs reference.
- Full dense FP8 fwd sweep (hd128, causal+non-causal, seqlen 512–16384): 0 failures.
- DeepSeek-MLA paged uses a separate kernel this PR doesn't touch → unaffected.

Johnsonms (approved):

PR #2518, opened 2026-04-29
Fix varlen backward preprocess predicate slicing

hebo1221:
## Summary
- slice the O/dO out-of-bounds predicate to match each per-row copy in backward preprocess
- add a focused regression test for varlen backward preprocess with `head_dim_v=104`
- verify the computed `dpsum` values and padded `dq_accum` initialization, not only successful compilation

## Details
For `head_dim_v` values that are padded to a multiple of 32, backward preprocess builds a predicate for the full O/dO copy tile. The loop copies `tOgO[None, m, None]` and `tOgdO[None, m, None]`, so passing the full predicate leaves the predicate shape wider than the sliced copy operand. Slicing the predicate by the same `m` coordinate makes the predicate shape match the copy shape.

Fixes #2492.

## Validation
Refreshed onto current `main` (`b54df16`) and validated on NVIDIA GB10 / SM121 with PyTorch 2.13.0+cu130 and nvidia-cutlass-dsl 4.6.0.dev0:

- `FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1 pytest -q -x tests/cute/test_flash_attn.py::test_flash_attn_bwd_preprocess_varlen_head_dim_v_not_multiple_of_32` — 2 passed
- exact public repro from #2492 (`head_dim=104`, 16 heads, two 3600-token varlen sequences) — forward and backward passed; q/k/v gradients were all finite
- `ruff check` and `ruff format --check` for `flash_bwd_preprocess.py`
- `python -m compileall -q flash_attn/cute/flash_bwd_preprocess.py tests/cute/test_flash_attn.py`

PR #2698, opened 2026-07-09
[CuTe,Bwd,Sm100] Support head_dim not a multiple of 32 in backward

### LINKED ISSUE #2492: [FA4] Backward error on head_dim not divisible by 32 `'cute.copy' op expects pred to have compatible shape with: (1,(1)) but got actual predShape`
Hello, I was wondering whether the below is expected. On backward pass for FA4 for head_dim 104, it raises a shape mismatch error. This happened when trying to run Pixtral with a head dim non divisible by 32. This issue does NOT occur on FA2.

Repro:

```python
import torch
from flash_attn.cute import flash_attn_varlen_func
hd, nh = 104, 16  # any head_dim % 32 != 0 works  
cu = torch.tensor([0, 3600, 7200], device='cuda', dtype=torch.int32)
q = torch.randn(7200, nh, hd, device='cuda', dtype=torch.bfloat16, requires_grad=True)
k = torch.randn_like(q, requires_grad=True); v = torch.randn_like(q, requires_grad=True)
out, *_ = flash_attn_varlen_func(q, k, v, cu_seqlens_q=cu, cu_seqlens_k=cu, max_seqlen_q=3600, max_seqlen_k=3600, causal=False)
out.sum().backward()
```

Env: B200 , Torch 2.10+cu130. FA4 b10. Python 312

Error: `'cute.copy' op expects pred to have compatible shape with: (1,(1)) but got actual predShape: (1,(8,1))`

```
  File "/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/nvidia_cutlass_dsl/python_packages/cutlass/base_dsl/dsl.py", line 1088, in build_module
    module.operation.verify()
cutlass._mlir._mlir_libs._site_initialize.<locals>.MLIRError: Verification failed:
error: "copy(tOgO[None, m, None], tOrO[None, m, None])"("/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/flash_attn/cute/flash_bwd_preprocess.py":303:16): 'cute.copy' op expects pred to have compatible shape with: (1,(1)) but got actual predShape: (1,(8,1))
 note: "copy(tOgO[None, m, None], tOrO[None, m, None])"("/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/flash_attn/cute/flash_bwd_preprocess.py":303:16): see current operation: "cute.copy"(%1211, %1230, %1245, %1260) <{operandSegmentSizes = array<i32: 1, 1, 1, 1>}> : (!cute_nvgpu.atom.universal_copy<bf16, 128 b>, !cute.memref<bf16, gmem, align<16>, "((8,1),(1)):((1,0),(0))">, !cute.memref<bf16, rmem, align<16>, "((8,1),(1)):((1,0),(0))">, !cute.memref<i8, rmem, align<32>, "(1,(8,1)):(1,(0,1))">) -> ()

During handling of the above exception, another exception occurred:

  File "<stdin>", line 1, in <module>
  File "/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/torch/_tensor.py", line 631, in backward
    torch.autograd.backward(
  File "/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/torch/autograd/__init__.py", line 381, in backward
    _engine_run_backward(
  File "/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/torch/autograd/graph.py", line 869, in _engine_run_backward
    return Variable._execution_engine.run_backward(  # Calls into the C++ engine to run the backward pass
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/torch/autograd/function.py", line 317, in apply
    return user_fn(self, *args)
           ^^^^^^^^^^^^^^^^^^^^
  File "/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/flash_attn/cute/interface.py", line 1935, in backward
    dq, dk, dv = _flash_attn_bwd(
                 ^^^^^^^^^^^^^^^^
  File "/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/flash_attn/cute/interface.py", line 1470, in _flash_attn_bwd
    _bwd_preprocess(
  File "/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/flash_attn/cute/interface.py", line 1091, in _bwd_preprocess
    _bwd_preprocess.compile_cache[compile_key] = _compile_bwd_preprocess(*compile_key)
                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/flash_attn/cute/interface.py", line 1073, in _compile_bwd_preprocess
    return cute.compile(
           ^^^^^^^^^^^^^
  File "/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/nvidia_cutlass_dsl/python_packages/cutlass/base_dsl/dsl.py", line 1090, in build_module
    raise DSLRuntimeError("______ ICE IR Verification Failed ______", cause=e)
cutlass.base_dsl.common.DSLRuntimeError: DSLRuntimeError: ______ ICE IR Verification Failed ______
  Caused exception: Verification failed:
error: "copy(tOgO[None, m, None], tOrO[None, m, None])"("/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/flash_attn/cute/flash_bwd_preprocess.py":303:16): 'cute.copy' op expects pred to have compatible shape with: (1,(1)) but got actual predShape: (1,(8,1))
 note: "copy(tOgO[None, m, None], tOrO[None, m, None])"("/root/miniconda3/envs/py3.12/lib/python3.12/site-packages/flash_attn/cute/flash_bwd_preprocess.py":303:16): see current operation: "cute.copy"(%1211, %1230, %1245, %1260) <{operandSegmentSizes = array<i32: 1, 1, 1, 1>}> : (!cute_nvgpu.atom.universal_copy<bf16, 128 b>, !cute.memref<bf16, gmem, align<16>, "((8,1),(1)):((1,0),(0))">, !cute.memref<bf16, rmem, align<16>, "((8,1),(1)):((1,0),(0))">, !cute.memref<i8, rmem, align<32>, "(1,(8,1)):(1,(0,1))">) -> ()
```

For potential solution, I had Claude help dig into this, so this may be incorrect.

```
Root cause (now confirmed with evidence, not speculation):                                                                                                                                                                                                                  
  - FlashAttentionBackwardPreprocess (in flash_bwd_preprocess.py) pads head_dim to a multiple of_32.                                                                                                                                                                          
  - When head_dim_v != head_dim_v_padded (i.e., head_dim not aligned to_32), it sets check_hdim_v_oob=True.                                                                                                                                                                   
  - That triggers predicate_k() (in utils.py:465) which builds a predicate with shape (1, CPY_M, CPY_K).                                                                                                                                                                      
  - But the universal_copy<bf16, 128 b> atom has sub-mode (8,1), and the cute.copy op's MLIR verifier expects a predicate whose nesting matches the atom. The predicate shape (1,(8,1)) doesn't match the required (1,(1)), and verification fails at compile time.           
  - Any head_dim aligned to 32 skips predicate_k entirely (tOpO = None) and works fine.

Upstream fix location: either flash_attn/cute/utils.py:predicate_k (build a predicate whose shape coalesces with the copy atom's CPY_Atom sub-mode), or flash_attn/cute/flash_bwd_preprocess.py:303 (flatten the copy slices so no predicate inference happens against the  
  nested atom). Most surgical fix is probably in predicate_k.
```

KaijingOfficial:
## Summary

Backward on SM100 (B200/B300) crashes or **silently produces wrong dK/dV** when `head_dim % 32 != 0` (e.g. ViT `head_dim=72`, `104`). Fixes #2492. Three independent root causes:

1. **`flash_bwd_preprocess.py`** — the O/dO load predicate (`predicate_k`) carries a `CPY_M` mode broadcast over `m`, but the copy source is sliced per-`m` (`tOgO[None, m, None]`) without slicing the predicate. The vectorized 128-bit copy atom's predicate-shape verifier then rejects the extra mode (`expects (1,(k)) but got (1,(cpy_m,k))`). Slice the predicate per-`m`, as `flash_fwd.py` / `flash_bwd.py` already do.

2. **`flash_bwd_sm100.py`** — `tile_hdim` padded `head_dim` to a multiple of 16, but the dQ accumulator reduce requires a multiple of 32 (`assert tile_hdim % dQ_reduce_ncol == 0`), and both the preprocess kernel and the interface already round to 32. `head_dim=72 -> 80` tripped the assert. Pad to 32 so all three agree (96); this reuses the already-working `head_dim=96` path.

3. **`interface.py`** — varlen MHA uses the non-TMA dK/dV epilogue store, which writes `tile_hdim` (mult of 32) columns per row with no head-dim OOB predication (its 16-wide vectorization cannot mask a 72 boundary). It spilled past the real `head_dim` and corrupted neighboring dK/dV. Route that store through a head-dim-padded scratch buffer, then slice back. Gated on `arch // 10 in [10, 11]` so SM90/SM120 skip the extra alloc + copy (suggested by @hebo1221).

## Test results

Validated on **B300 (SM103)** against `torch.nn.functional.scaled_dot_product_attention` (fp32 reference):

- **Coverage**: dense + varlen, causal + non-causal, `head_dim ∈ {64, 72, 80, 104, 128}`, MHA and GQA(hd=128).
- **Result**: `dq/dk/dv` max relative error **2e-3 – 7e-3** (bf16 precision), matching the multiple-of-32 baselines. No regression on aligned head dims.
- **End-to-end**: Qwen3.5-VL training (ViT `head_dim=72`, varlen) trains with a decreasing loss and finite grad norm.

## Relation to #2633

This is an alternative to #2633. #2633 keeps V padding at a multiple of 16 and predicates the non-TMA store via a reduced tmem `Repetition` — neater for varlen. However, in our testing that leaves the **dense (TMA-store) dV path numerically wrong** for `head_dim % 32 != 0` (relative error ~1–13 on `dv` for hd 72/80/104), while the varlen path is correct. Padding V to 32 keeps **both dense and varlen** paths correct, at the cost of a little extra dV compute. Happy to combine approaches if maintainers prefer the leaner store.

## Test plan
- [x] Numerical parity vs SDPA (fp32 ref), dense + varlen, causal + non-causal, hd {64,72,80,104,128}
- [x] No regression on multiple-of-32 head dims
- [x] End-to-end VLM training smoke (ViT hd=72)
- [ ] Maintainer CI matrix

## Acknowledgments

Thanks to @hebo1221 for independently verifying all three fixes via FakeTensor compilation on SM121 and for pointing out that the scratch-buffer path should be gated to SM100/SM110 only.

### DIFF
--- a/flash_attn/cute/flash_bwd_preprocess.py
+++ b/flash_attn/cute/flash_bwd_preprocess.py
@@ -13,7 +13,6 @@
 # So the main backward kernel is unchanged; we just replace D with D' = D - dLSE here.
 import math
 import operator
-from functools import partial
 from typing import Callable, Type, Optional
 
 import cuda.bindings.driver as cuda
@@ -365,8 +364,8 @@ class FlashAttentionBackwardPreprocess:
             tOpO = None
             if const_expr(self.check_hdim_v_oob):
                 tOpO = copy_utils.predicate_k(tOcO, limit=headdim_v)
-            # Each copy will use the same predicate
-            copy = partial(copy_utils.copy, pred=tOpO)
+            # Each copy uses the predicate sliced to the current m (see loop below).
+            copy = copy_utils.copy
 
             tOrO = cute.make_rmem_tensor_like(tOgO)
             tOrdO = cute.make_rmem_tensor_like(tOgdO)
@@ -378,8 +377,13 @@ class FlashAttentionBackwardPreprocess:
                 # Instead of using tOcO, we using t0OcO and subtract the offset from the limit.
                 # This is bc the entries of t0OcO are known at compile time.
                 if t0OcO[0, m, 0][0] < seqlen_limit - tOcO[0][0]:
-                    copy(tOgO[None, m, None], tOrO[None, m, None])
-                    copy(tOgdO[None, m, None], tOrdO[None, m, None])
+                    # The predicate carries a CPY_M mode (broadcast over m); slice it to the
+                    # current m so its shape matches the m-sliced copy source. Otherwise the
+                    # vectorized copy atom's predicate-shape verification fails when head_dim_v
+                    # is not a multiple of the copy-atom width (e.g. 72, 104).
+                    tOpO_cur = tOpO[None, m, None] if const_expr(tOpO is not None) else None
+                    copy(tOgO[None, m, None], tOrO[None, m, None], pred=tOpO_cur)
+                    copy(tOgdO[None, m, None], tOrdO[None, m, None], pred=tOpO_cur)
             # O and dO loads are done; signal that the next kernel can start.
             # Correctness is ensured by griddepcontrol_wait() in bwd_sm90 before it reads our outputs.
             if const_expr(self.use_pdl):
--- a/flash_attn/cute/flash_bwd_preprocess.py
+++ b/flash_attn/cute/flash_bwd_preprocess.py
@@ -13,7 +13,6 @@
 # So the main backward kernel is unchanged; we just replace D with D' = D - dLSE here.
 import math
 import operator
-from functools import partial
 from typing import Callable, Type, Optional
 
 import cuda.bindings.driver as cuda
@@ -367,8 +366,8 @@ class FlashAttentionBackwardPreprocess:
             tOpO = None
             if const_expr(self.check_hdim_v_oob):
                 tOpO = copy_utils.predicate_k(tOcO, limit=headdim_v)
-            # Each copy will use the same predicate
-            copy = partial(copy_utils.copy, pred=tOpO)
+            # Each copy uses the predicate sliced to the current m (see loop below).
+            copy = copy_utils.copy
 
             tOrO = cute.make_rmem_tensor_like(tOgO)
             tOrdO = cute.make_rmem_tensor_like(tOgdO)
@@ -380,8 +379,13 @@ class FlashAttentionBackwardPreprocess:
                 # Instead of using tOcO, we using t0OcO and subtract the offset from the limit.
                 # This is bc the entries of t0OcO are known at compile time.
                 if t0OcO[0, m, 0][0] < seqlen_limit - tOcO[0][0]:
-                    copy(tOgO[None, m, None], tOrO[None, m, None])
-                    copy(tOgdO[None, m, None], tOrdO[None, m, None])
+                    # The predicate carries a CPY_M mode (broadcast over m); slice it to the
+                    # current m so its shape matches the m-sliced copy source. Otherwise the
+                    # vectorized copy atom's predicate-shape verification fails when head_dim_v
+                    # is not a multiple of the copy-atom width (e.g. 72, 104).
+                    tOpO_cur = tOpO[None, m, None] if const_expr(tOpO is not None) else None
+                    copy(tOgO[None, m, None], tOrO[None, m, None], pred=tOpO_cur)
+                    copy(tOgdO[None, m, None], tOrdO[None, m, None], pred=tOpO_cur)
             # O and dO loads are done; signal that the next kernel can start.
             # Correctness is ensured by griddepcontrol_wait() in bwd_sm90 before it reads our outputs.
             if const_expr(self.use_pdl):
--- a/flash_attn/cute/flash_bwd_sm100.py
+++ b/flash_attn/cute/flash_bwd_sm100.py
@@ -68,8 +68,10 @@ class FlashAttentionBackwardSm100:
         has_aux_tensors: cutlass.Constexpr = False,
         q_subtile_factor: cutlass.Constexpr[int] = 1,
     ):
-        # padding head_dim to a multiple of 16 as k_block_size
-        hdim_multiple_of = 16
+        # Pad head_dim to a multiple of 32 (k_block_size). Must stay a multiple of 32 so
+        # that the dQ accumulator reduce (dQ_reduce_ncol up to 32) tiles evenly and matches
+        # both the preprocess kernel and the interface's head_dim_rounded (also mult of 32).
+        hdim_multiple_of = 32
         self.tile_hdim = int(math.ceil(head_dim / hdim_multiple_of) * hdim_multiple_of)
         head_dim_v = head_dim_v if head_dim_v is not None else head_dim
         self.same_hdim_kv = head_dim == head_dim_v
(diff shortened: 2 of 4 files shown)


hebo1221:
I tried rebasing this on current main (`a369df7`) while checking whether my older #2518 was enough. It isn't: the predicate change gets past the preprocess error, but SM100 backward still hits `dQ_reduce_ncol` for head dims 72 and 104.

With all three changes from this PR ported over, full varlen fwd+bwd compiled under FakeTensor for head dims 72/104, causal on/off, with `FLASH_ATTENTION_ARCH=sm_100` and `CUTE_DSL_ARCH=sm_100a` (4/4).

One thing I noticed during the port: the padded dK/dV scratch path should probably be limited to `arch // 10 in (10, 11)`. Otherwise SM90 and SM120 allocate and copy the temporary buffers too, even though their store paths don't need them.

I also added a small public API regression test for those four cases. It passes against FP32 SDPA on GB10/SM121. I don't have B200/B300 access, so the SM100 result above is compile-only. I can send the test and guard as a patch if that helps.

KaijingOfficial:
@hebo1221  Thanks for the suggestion — both are incorporated in the latest commit. The scratch-buffer path is now gated to arch // 10 in [10, 11] so SM90/SM120 no longer pay the extra alloc.

We've been running this patch in production (Qwen3.5-VL training on SM103, varlen, head_dim=72) for a while now with no issues. Happy to take your regression test as a follow-up PR against this branch if you'd like to send it over.

KaijingOfficial:
@tridao  Could you take a look when you have a chance? This fixes SM100 backward for non-32-aligned head dims (e.g. ViT head_dim=72)



@WonsukYangTrlbs FYI since this overlaps with #2633 . The main difference is padding V to 32 (rather than 16) keeps both the dense TMA-store and varlen paths correct; happy to discuss if a combined approach makes sense.

KaijingOfficial:
cc @Johnsonms

drisspg (approved):
made a small tweak but thank you!

PR #2702, opened 2026-07-11
hopper/setup.py: harden tarfile extraction against path traversal and symlink escape

### LINKED ISSUE #2637: Bare tarfile.extractall into Predictable Cache During Build in flash-attention
## Summary

flash-attention's `hopper/setup.py` downloads NVIDIA toolchain archives (nvcc, ptxas) from NVIDIA CDN during build and extracts them with bare `tarfile.extractall(path=tmp_path)` into a predictable, user-home-rooted cache directory (`~/.flashattn/nvidia/<name>`) with no tar member filtering, no symlink validation, and a check-then-extract TOCTOU on the cache path.

A local attacker who pre-plants a symlink at `~/.flashattn/nvidia/nvcc` before `pip install flash-attn` (from source) causes `extractall` to write NVIDIA toolchain binaries through the symlink to an attacker-chosen directory — achieving arbitrary file write with the victim's privileges at build time.

## Affected Component

- **Repository**: <https://github.com/Dao-AILab/flash-attention>
- **File**: `hopper/setup.py`
- **Vulnerable function**: `download_and_copy()` (line 405-429)
- **Version**: latest main (as of 2026-06-09), including the FA4/hopper branch

### Realistic scenarios

| Scenario | Feasibility |
|---|---|
| Shared multi-user GPU build server | High — common in ML teams |
| CI/CD with shared home / persistent cache | High — `~/.flashattn` persists between jobs |
| Compromised lower-privilege service on same host | Medium — pre-plants symlink before victim builds |

### Vulnerable function: `download_and_copy()` (line 405-429)

```python
def download_and_copy(name, src_func, dst_path, version, url_func):
    flashattn_cache_path = get_flashattn_cache_path()          # ~/.flashattn
    ...
    tmp_path = os.path.join(flashattn_cache_path, "nvidia", name)  # ~/.flashattn/nvidia/nvcc
    ...
    src_path = os.path.join(tmp_path, src_path)
    download = not os.path.exists(src_path)                    # ← check (follows symlink)
    if download:
        file = tarfile.open(fileobj=open_url(url), mode="r|*")
        file.extractall(path=tmp_path)                         # ← bare extractall, no filter
    os.makedirs(os.path.split(dst_path)[0], exist_ok=True)
    ...
    shutil.copy(src_path, dst_path)                            # also copies from symlink-resolved path
```

### Cache path predictability

```python
def get_flashattn_cache_path():
    user_home = os.getenv("FLASH_ATTENTION_HOME") or os.getenv("HOME")
    return os.path.join(user_home, ".flashattn")   # ~/.flashattn — fully predictable
```

Cache path: `~/.flashattn/nvidia/nvcc` and `~/.flashattn/nvidia/ptxas` — known, stable, no randomization.

### No guards

- No `os.path.islink(tmp_path)` check
- No `os.makedirs(tmp_path, exist_ok=False)` (atomic creation)
- No `tarfile` extraction filter (`filter='data'` or member name validation)
- No integrity verification on extracted files (beyond what HTTPS provides for transport)

### Call site (line 472-480)

```python
download_and_copy(
    "nvcc", nvcc_src, nvcc_dst, NVIDIA_TOOLCHAIN_VERSION["nvcc"],
    lambda system, arch, version:
        f"https://developer.download.nvidia.com/compute/cuda/redist/cuda_nvcc/{system}-{arch}/cuda_nvcc-{system}-{arch}-{version}-archive.tar.xz",
)
download_and_copy(
    "ptxas", ...
)
```

Triggered during `pip install flash-attn` from source (Hopper/Blackwell builds) when CUDA ≥ 12.3 toolchain is needed.

## Attack Chain

1. Attacker creates `~/.flashattn/nvidia/nvcc` → symlink to `/target/dir` (e.g., victim's `~/.local/bin/`)
2. Victim runs `pip install flash-attn` (from source, on CUDA machine)
3. `download_and_copy("nvcc", ...)` checks `os.path.exists(src_path)` → False (target is empty)
4. Downloads legitimate NVIDIA nvcc tar from CDN
5. `file.extractall(path=tmp_path)` — `tmp_path` is symlink → writes nvcc binaries to attacker-chosen location
6. Impact: attacker-controlled location now contains executable binaries (or overwrites existing ones)

## Proof of Concept 

```python
import os, tarfile, io, tempfile
from pathlib import Path
FAKE_HOME = Path(tempfile.mkdtemp())
ATTACKER_TARGET = Path(tempfile.mkdtemp())
flashattn_cache = FAKE_HOME / ".flashattn"
tmp_path = flashattn_cache / "nvidia" / "nvcc"
tmp_path.parent.mkdir(parents=True, exist_ok=True)
os.symlink(str(ATTACKER_TARGET), str(tmp_path))
tar_buf = io.BytesIO()
with tarfile.open(fileobj=tar_buf, mode="w:xz") as tf:
    info = tarfile.TarInfo(name="cuda_nvcc-linux-x86_64-12.6.85-archive/bin/nvcc")
    data = b"#!/bin/bash\necho pwned"
    info.size = len(data)
    info.mode = 0o755
    tf.addfile(info, io.BytesIO(data))
tar_buf.seek(0)
file = tarfile.open(fileobj=tar_buf, mode="r|*")
file.extractall(path=str(tmp_path))  
assert (ATTACKER_TARGET / "cuda_nvcc-linux-x86_64-12.6.85-archive" / "bin" / "nvcc").exists()
```

## Suggested Fix

```python
def download_and_copy(name, src_func, dst_path, version, url_func):
    ...
    tmp_path = os.path.join(flashattn_cache_path, "nvidia", name)

    # Refuse symlinks
    if os.path.islink(tmp_path):
        raise RuntimeError(f"Refusing to extract to symlink: {tmp_path}")

    os.makedirs(tmp_path, exist_ok=True)

    src_path = os.path.join(tmp_path, src_path)
    download = not os.path.exists(src_path)
    if download:
        file = tarfile.open(fileobj=open_url(url), mode="r|*")
        # Use filter on Python 3.12+
        if hasattr(tarfile, 'data_filter'):
            file.extractall(path=tmp_path, filter='data')
        else:
            file.extractall(path=tmp_path)
    ...
```

aryanputta:
Fixes #2637.

`download_and_copy()` extracts NVIDIA toolchain archives with a bare `tarfile.extractall()` into the predictable `~/.flashattn/nvidia/<name>` cache, allowing arbitrary file write at build time via a pre-planted symlink or a malicious archive member.

## Changes
- Add `safe_extractall()`: uses the PEP 706 `data` filter when the interpreter supports it (added in 3.12, backported to 3.10.12 / 3.11.4). On older interpreters the `filter` keyword does not exist, so it validates each member's resolved path stays inside the destination and rejects link members before extracting. The tar is opened in stream mode (`r|*`), so members are validated and extracted in a single pass.
- Refuse extraction when the cache path is a symlink, closing the primary pre-planted-symlink vector on all supported Python versions.

## Verification
Against a crafted tar with a `../` traversal member and a symlink member: the original bare `extractall` writes outside the target directory; the patched path blocks it on both the filter path and the pre-3.10.12 fallback path, and benign archives still extract normally.

### DIFF
--- a/hopper/setup.py
+++ b/hopper/setup.py
@@ -415,12 +415,28 @@ def safe_extractall(tar, path):
     if hasattr(tarfile, "data_filter"):
         tar.extractall(path=dest, filter="data")
         return
+
+    def stays_inside(resolved):
+        return resolved == dest or resolved.startswith(dest + os.sep)
+
     for member in tar:
         target = os.path.realpath(os.path.join(dest, member.name))
-        if target != dest and not target.startswith(dest + os.sep):
+        if not stays_inside(target):
             raise RuntimeError(f"Refusing tar member outside extract dir: {member.name!r}")
+        # The data filter permits links whose target resolves inside the
+        # destination (the cuda_nvcc archives ship such intra-package symlinks,
+        # e.g. libnvvm.so -> libnvvm.so.4). Mirror that instead of rejecting all
+        # links, or real builds regress on interpreters without the data filter.
         if member.issym() or member.islnk():
-            raise RuntimeError(f"Refusing link member in tar archive: {member.name!r}")
+            # A symlink target is relative to the link's own directory; a hard
+            # link's is relative to the archive root.
+            link_base = os.path.dirname(target) if member.issym() else dest
+            link_target = os.path.realpath(os.path.join(link_base, member.linkname))
+            if not stays_inside(link_target):
+                raise RuntimeError(
+                    f"Refusing link member pointing outside extract dir: "
+                    f"{member.name!r} -> {member.linkname!r}"
+                )
         tar.extract(member, path=dest)
 
 
@@ -443,7 +459,13 @@ def download_and_copy(name, src_func, dst_path, version, url_func):
         # Refuse to extract into a pre-planted symlink: an attacker with write
         # access to the predictable cache dir could otherwise redirect the
         # extraction (and the shutil.copy below) to an arbitrary location.
-        if os.path.islink(tmp_path):
+        # islink() only inspects the leaf, so also require the fully resolved
+        # path to stay under the cache root, catching a symlinked parent.
+        cache_root = os.path.realpath(flashattn_cache_path)
+        resolved_tmp = os.path.realpath(tmp_path)
+        if os.path.islink(tmp_path) or not (
+            resolved_tmp == cache_root or resolved_tmp.startswith(cache_root + os.sep)
+        ):
             raise RuntimeError(f"Refusing to extract into symlinked cache path: {tmp_path}")
         os.makedirs(tmp_path, exist_ok=True)
         print(f'downloading and extracting {url} ...')
--- a/hopper/setup.py
+++ b/hopper/setup.py
@@ -402,6 +402,28 @@ def open_url(url):
     return urllib.request.urlopen(request, timeout=300)
 
 
+def safe_extractall(tar, path):
+    """Extract a tar stream, refusing members that would escape ``path``.
+
+    Uses the PEP 706 ``data`` filter when the interpreter supports it (added in
+    3.12, backported to 3.10.12 / 3.11.4). On older interpreters the filter
+    keyword does not exist, so we validate each member's resolved path stays
+    inside ``path`` and reject link members before extracting. The tar is opened
+    in stream mode, so members are validated and extracted in a single pass.
+    """
+    dest = os.path.realpath(path)
+    if hasattr(tarfile, "data_filter"):
+        tar.extractall(path=dest, filter="data")
+        return
+    for member in tar:
+        target = os.path.realpath(os.path.join(dest, member.name))
+        if target != dest and not target.startswith(dest + os.sep):
+            raise RuntimeError(f"Refusing tar member outside extract dir: {member.name!r}")
+        if member.issym() or member.islnk():
+            raise RuntimeError(f"Refusing link member in tar archive: {member.name!r}")
+        tar.extract(member, path=dest)
+
+
 def download_and_copy(name, src_func, dst_path, version, url_func):
     if is_offline_build():
         return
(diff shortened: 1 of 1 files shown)


Johnsonms:
Direction looks right and the data-filter path is correct, but one blocking issue before merge:

1.  Fallback rejects all link members → breaks real builds. The cuda_nvcc archive this downloads legitimately contains symlinks (verified on 12.6.85):

nvvm/lib64/libnvvm.so   -> libnvvm.so.4
nvvm/lib64/libnvvm.so.4 -> libnvvm.so.4.0.0

So on interpreters without tarfile.data_filter (Py 3.9.x, 3.10.0–11, 3.11.0–3), safe_extractall raises RuntimeError("Refusing link member…") and aborts — a regression, since the data filter itself allows these in-destination links. Fix: in the fallback, allow links whose resolved target stays inside dest instead of rejecting all.

2 Minor: os.path.islink(tmp_path) only checks the leaf; a symlinked parent still gets followed via realpath. Consider asserting the resolved path stays under the cache root.

Will approve once the fallback link handling is fixed.

aryanputta:
Thanks for the review, both addressed in 4e86f6a.

1. Fallback no longer rejects all links. It now allows a link whose resolved target stays inside the extract dir (symlink target resolved relative to the link's own directory, hard link relative to the archive root), which is what the data filter permits. The intra-package `libnvvm.so -> libnvvm.so.4` chain from cuda_nvcc extracts cleanly again, while escaping targets and absolute targets are still rejected. Verified against a synthetic tar reproducing the 12.6.85 nvvm symlink layout plus `../../etc/passwd` and `/etc/passwd` link members.

2. Hardened the cache-path check. `os.path.islink` only inspects the leaf, so I also require the fully resolved `tmp_path` to stay under the resolved cache root, which catches a symlinked parent directory.

Johnsonms:
Verified E2E against the real download_and_copy() (only open_url stubbed): both the data-filter and pre-3.10.12 fallback paths block ../ traversal + symlink-escape archives and refuse a pre-planted symlink cache before any network fetch, while benign archives (intra-pkg symlinks included) still extract+copy cleanly — LGTM.

Johnsonms (approved):

PR #2705, opened 2026-07-13
[CuTe, FA4] Preserve first-tile flag during scheduler reconstruction

dongxiao92:
## Summary

Preserve the Python-only `_is_first_block` flag when CuTe reconstructs FlashAttention single-tile schedulers from their MLIR values.

## Root cause

These schedulers omit `_is_first_block` from their MLIR representation, but their constructors initialize it to `True`.

After `advance_to_next_work()` marks the first tile consumed, reconstruction therefore resets the omitted flag and makes the first tile valid again. The kernel can repeat that tile and deadlock its producer/consumer pipeline.

## Change

Copy `_is_first_block` from the original object after reconstructing:

- `SingleTileScheduler`
- `SingleTileVarlenScheduler`
- `Sm100FmhaStaticTileScheduler`

## Validation
Main branch and cutlass-4.7.0 could pass with this change

### DIFF
--- a/flash_attn/cute/tile_scheduler.py
+++ b/flash_attn/cute/tile_scheduler.py
@@ -281,7 +281,10 @@ class SingleTileScheduler:
         for obj, n_items in zip([self.params, self._blk_coord], self._values_pos):
             obj_list.append(cutlass.new_from_mlir_values(obj, values[:n_items]))
             values = values[n_items:]
-        return SingleTileScheduler(*(tuple(obj_list)), loc=self._loc)
+        scheduler = SingleTileScheduler(*(tuple(obj_list)), loc=self._loc)
+        # Preserve the Python-only first-tile flag across CuTe reconstruction.
+        scheduler._is_first_block = self._is_first_block
+        return scheduler
 
 
 class StaticPersistentTileScheduler:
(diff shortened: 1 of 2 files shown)


dongxiao92:
This changes should not be merged before cutlass-dsl 4.7 release. I will update the version once the dsl package is public

dongxiao92:
@tridao  4.7.0 is live. I have also updated the version. Could you help to review? Thanks

flash_attn/cute/tile_scheduler.py
```diff
@@ -342,7 +342,10 @@ def __new_from_mlir_values__(self, values):
         for obj, n_items in zip([self.params, self._blk_coord], self._values_pos):
             obj_list.append(cutlass.new_from_mlir_values(obj, values[:n_items]))
             values = values[n_items:]
-        return SingleTileScheduler(*(tuple(obj_list)), loc=self._loc)
+        scheduler = SingleTileScheduler(*(tuple(obj_list)), loc=self._loc)
+        # Preserve the Python-only first-tile flag across CuTe reconstruction.
```
drisspg: is it safe to say this was a bug in older versions? regardless of 4.7 and this just happend to work; since I imagine we will need to sync the quack upgrade can we just have this change in this or a separate PR and the 4.7 bump isolated?

dongxiao92: >is it safe to say this was a bug in older versions? regardless of 4.7 and this just happend to work

Yes. It happened to work and got exposed with some DSL frontend changes in 4.7.

>can we just have this change in this or a separate PR and the 4.7 bump isolated?

Sounds good to me. I can cancel the version update. Should I add any comments to the code changes then?

drisspg: yeh I think 1 small commetn e.g. # Note python-only attributes; mlir drops the attributes that have no mlir equivalent and are only valid on the python object,   and then for the later call sties you just do `see # Note python-only attributes`

dongxiao92: Done. Please help to review again

drisspg (approved):

PR #2706, opened 2026-07-13
[Cute, bwd, sm90/100/110] Support learnable sink in backward

henrylhtsang:
## Summary
- add FP16, BF16, and FP32 learnable sink gradients to supported CuTe backward kernels on SM90, SM100, and SM110; the sink dtype is independent of Q/K/V and dSink preserves it
- reuse one dQ postprocess CTA per head to reduce dSink from FP32 dPsum and LSE, avoiding atomics and separate zero-initialization
- support fixed-length and varlen autograd paths, including gradients from a returned LSE
- return zero output and LSE equal to the sink when K is empty; learnable sink backward requires non-empty Q and K
- fix SM90 fully masked sink-only rows producing NaN LSE

## Test plan

```bash
CUDA_VISIBLE_DEVICES=0 FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1 pytest -x -q \
  'tests/cute/test_flash_attn.py::test_flash_attn_output[128-128-64-False-0-0.0-False-False-True-mha-dtype0]' \
  'tests/cute/test_flash_attn.py::test_flash_attn_output[128-128-128-True-0-0.0-False-False-True-gqa-dtype0]' \
  'tests/cute/test_flash_attn.py::test_flash_attn_output[128-128-192-False-1-15.0-True-False-True-mqa-dtype0]'
```

```bash
CUDA_VISIBLE_DEVICES=0 FLASH_ATTENTION_FAKE_TENSOR=1 \
  FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1 pytest -x -q \
  'tests/cute/test_flash_attn.py::test_flash_attn_learnable_sink_backward_dtype' \
  'tests/cute/test_flash_attn.py::test_flash_attn_varlen_learnable_sink_backward_with_lse'
```

The first three cases passed on SM100. The focused fake-tensor run passed all four cases: FP16, BF16, and FP32 sink dtypes plus the varlen/LSE regression.


henrylhtsang:
cc @Johnsonms @drisspg @jayhshah would appreciate a review

flash_attn/cute/flash_bwd_postprocess.py
```diff
@@ -293,6 +312,10 @@ def kernel(
         mdQ: cute.Tensor,
         mCuSeqlensQ: Optional[cute.Tensor],
         mSeqUsedQ: Optional[cute.Tensor],
+        mdPsum: Optional[cute.Tensor],
+        mLSE: Optional[cute.Tensor],
```
drisspg: ditto down ehre

flash_attn/cute/interface.py
```diff
@@ -1255,13 +1266,18 @@ def _compile_bwd_postprocess(
     batchp1 = cute.sym_int()
     mCuSeqlensQ = fake_tensor(Int32, (batchp1,), divisibility=1) if has_cuseqlens_q else None
     mSeqUsedQ = fake_tensor(Int32, (batch,), divisibility=1) if has_seqused_q else None
+    mLearnableSink = fake_tensor(cutlass.BFloat16, (mQ.shape[-2],), divisibility=1) if has_learnable_sink else None
```
drisspg: side note; do we not support fp16  for our sinks on purpose here? feels like a good follow up

flash_attn/cute/interface.py
```diff
@@ -1273,18 +1289,21 @@ def _bwd_postprocess_convert(
     arch, dtype, hdim, block_size, num_threads,
     atom_layout, swap_ab,
     use_2cta_instrs=False, cluster_size=1,
+    dpsum=None, lse=None, learnable_sink=None, dsink=None,
```
drisspg: same deal could be a tuple of types since they need to come together

flash_attn/cute/interface.py
```diff
@@ -2706,8 +2764,14 @@ def backward(ctx, dout, dlse):
                 aux_scalars=ctx.aux_scalars,
                 mask_mod=ctx.mask_mod,
                 dlse=dlse,
+                learnable_sink=learnable_sink,
```
drisspg, line 3309: codex: It doesn't look like learnable-sink backward is tested for varlen—the varlen test hardcodes `has_learnable_sink=False` and excludes it from backward. Could you add a numerical varlen regression, ideally with unequal and zero sequence lengths?

tests/cute/test_flash_attn.py
```diff
@@ -396,7 +398,10 @@ def test_flash_attn_output(
                 pytest.xfail("SM90 GQA bwd currently requires headdim == headdim_v")
             g = torch.randn_like(out)
             # do_o = ((g.float() * out.float()).sum(-1)).transpose(1, 2)
-            dq, dk, dv = torch.autograd.grad(out, (q, k, v), g)
+            grad_tensors = (q, k, v, learnable_sink) if has_learnable_sink else (q, k, v)
+            grads = torch.autograd.grad(out, grad_tensors, g)
```
drisspg, line 412: CODEX: 

Can we also add a `return_lse=True` case with `learnable_sink`, using a loss that depends on both `out` and `lse`, and compare `dSink` against the reference? This currently differentiates only `out`, so it tests the `dO` contribution to `dSink` but not the `dLSE` contribution. I want both features exercised together.

flash_attn/cute/flash_bwd_postprocess.py
```diff
@@ -328,6 +351,72 @@ def kernel(
...
+                            lse_val = mLSE[sink_batch, sink_head_idx, sink_row]
+                        else:
+                            dpsum_val = mdPsum[
+                                sink_head_idx, sink_seqlen.padded_offset_q + sink_row
+                            ]
+                            lse_val = mLSE[sink_head_idx, sink_seqlen.offset_q + sink_row]
+                        lse_val = Float32(lse_val)
+                        sink_prob = (
+                            Float32(1.0)
+                            if lse_val == -Float32.inf
```
drisspg, line 403: Codex now: SM90 sink-only rows can reach this code with `LSE = NaN`, so this `-inf` special case does not protect them and `sink_prob`/`dSink` become NaN. `flash_fwd_sm90.py` leaves `row_max=-inf` before adding the finite sink, while SM100 replaces the empty-row max with the sink. Could SM90 use the same handling and add a causal `seqlen_q > seqlen_k` sink-gradient regression? The current backward test excludes that case.

flash_attn/cute/flash_bwd_postprocess.py
```diff
@@ -215,6 +216,10 @@ def __call__(
         scale: cutlass.Float32,
         mCuSeqlensQ: Optional[cute.Tensor],
         mSeqUsedQ: Optional[cute.Tensor],
+        mdPsum: Optional[cute.Tensor],
```
drisspg: Could we bundle these as `sink_tensors: tuple[cute.Tensor, cute.Tensor, cute.Tensor, cute.Tensor] | None`, or preferably a small `NamedTuple`? These four tensors are all-or-none, so grouping them would encode that invariant and avoid threading four optional arguments through both entry points.

flash_attn/cute/interface.py
```diff
@@ -495,7 +504,15 @@ def _flash_attn_fwd(
     if seqlen_k == 0 or total_q == 0:
         out.zero_()
         if lse is not None:
-            lse.fill_(float("-inf"))
+            if learnable_sink is None:
+                lse.fill_(float("-inf"))
+            else:
```
henrylhtsang, line 728: NOTE: this is a behavior change for forward when sink is present.

flash_attn/cute/softmax.py
```diff
@@ -205,12 +205,13 @@ def finalize(
...
 
         for r in cutlass.range(cute.size(row_sum), unroll_full=True):
+            row_max_scaled = row_max[r] * scale_log2
             if cutlass.const_expr(sink_val is not None):
                 sink_val_cur = sink_val if not isinstance(sink_val, cute.Tensor) else sink_val[r]
                 LOG2_E = math.log2(math.e)
-                row_sum[r] += cute.math.exp2(
-                    sink_val_cur * LOG2_E - row_max[r] * scale_log2, fastmath=True
-                )
+                if row_max[r] == -Float32.inf:
```
henrylhtsang, line 212: afaik this is for sm90 only.

henrylhtsang:
addressed feedback

henrylhtsang:
Tested this on an H100 / SM90. We have to relax the `dSink` test tolerance for the SM90 result to pass, but the relaxation is narrow and dtype-aware rather than a blanket threshold.

The kernel accumulates `dSink` in FP32 and casts it to the sink dtype. On SM90, the varlen + LSE case differed from the PyTorch reference by exactly one BF16 ULP (`0.00390625`), while the previous cancellation-based tolerance evaluated to zero. FP16 also needed an output-dtype rounding allowance; FP32 passed the original comparison.

flash_attn/cute/interface.py
drisspg: this makes sense to me ; feels consistent

flash_attn/cute/interface.py
```diff
@@ -1358,6 +1399,16 @@ def _flash_attn_bwd(
...
+        assert arch // 10 in [9, 10, 11], "Learnable sink backward is supported on SM90 and SM100/SM110"
+        assert lse is not None, "learnable_sink backward requires LSE"
+        if q.numel() == 0 or k.numel() == 0:
+            dsink = (
+                dlse.sum(dim=(0, 2) if dlse.ndim == 3 else 1).to(learnable_sink.dtype)
+                if dlse is not None
+                else torch.zeros_like(learnable_sink)
+            )
+            return torch.zeros_like(q), torch.zeros_like(k), torch.zeros_like(v), dsink
     sparse_q = None
```
drisspg, line 1830: Could we preserve the preallocated `dq`/`dk`/`dv` outputs in this early-return path? Other paths in `_flash_attn_bwd` respect these buffers (and `test_flash_attn_bwd_preallocated_outputs` checks their identity). For example:

```suggestion
        if q.numel() == 0 or k.numel() == 0:
            dq = torch.zeros_like(q) if dq is None else dq.zero_()
            dk = torch.zeros_like(k) if dk is None else dk.zero_()
            dv = torch.zeros_like(v) if dv is None else dv.zero_()
            dsink = (
                dlse.sum(dim=(0, 2) if dlse.ndim == 3 else 1).to(learnable_sink.dtype)
                if dlse is not None
                else torch.zeros_like(learnable_sink)
            )
            return dq, dk, dv, dsink
```

drisspg (commented):
Looks pretty good; two comments

flash_attn/cute/flash_bwd_postprocess.py
```diff
@@ -328,6 +355,73 @@ def kernel(
...
+                num_warps = self.num_threads // cute.arch.WARP_SIZE
+                if lane_idx == 0:
+                    sdQaccum_flat[warp_idx] = sink_sum
+                cute.arch.barrier(barrier_id=5, number_of_threads=self.num_threads)
+                if warp_idx == 0:
+                    sink_sum = sdQaccum_flat[lane_idx] if lane_idx < num_warps else Float32(0.0)
+                    sink_sum = utils.warp_reduce(sink_sum, operator.add)
+                    if lane_idx == 0:
+                        mdSink[sink_head_idx] = sink_sum.to(mdSink.element_type)
+                cute.arch.barrier(barrier_id=5, number_of_threads=self.num_threads)
```
drisspg: How is performance here? Do we get enough parallelism?

flash_attn/cute/flash_bwd_postprocess.py
```diff
@@ -328,6 +355,73 @@ def kernel(
...
+                        sink_row += self.num_threads
+                    sink_batch += 1
+
+                sink_sum = utils.warp_reduce(sink_sum, operator.add)
+                lane_idx = cute.arch.lane_idx()
+                warp_idx = tidx // cute.arch.WARP_SIZE
+                num_warps = self.num_threads // cute.arch.WARP_SIZE
+                if lane_idx == 0:
+                    sdQaccum_flat[warp_idx] = sink_sum
+                cute.arch.barrier(barrier_id=5, number_of_threads=self.num_threads)
```
drisspg: Good follow up; we should switch to named barriers

drisspg: wait actually can this just be: cute.arch.sync_threads()?

flash_attn/cute/flash_bwd_postprocess.py
henrylhtsang: good call out. it does have perf impact. Though as tensors get bigger, perf impact gets smaller. 

| Batch | Sequence length | Q/KV heads | Head dimension | No sink | Sink | Overhead (backward e2e) |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 4K | 2/2 | 64 | 0.07174 ms | 0.07380 ms | +2.87% |
| 1 | 4K | 2/2 | 128 | 0.07373 ms | 0.07579 ms | +2.79% |
| 1 | 16K | 2/2 | 64 | 0.49600 ms | 0.50314 ms | +1.44% |
| 1 | 16K | 2/2 | 128 | 0.58469 ms | 0.59039 ms | +0.98% |
| 4 | 4K | 32/4 | 64 | 1.60911 ms | 1.61899 ms | +0.61% |
| 4 | 4K | 32/4 | 128 | 2.51894 ms | 2.51404 ms | Within noise |
| 4 | 16K | 32/4 | 64 | 23.58911 ms | 23.62510 ms | +0.15% |
| 4 | 16K | 32/4 | 128 | 35.71852 ms | 35.69682 ms | Within noise |

alternative approach would require a workspace for partial dSink, or a semaphore or a second kernel to be deterministic.

Alternatively, we can use two different approaches one for deterministic one for non-deterministic. But that feels like an overkill.

henrylhtsang: @drisspg alternative approach is worse than this approach for small tensors

| Batch | Sequence length | Q/KV heads | Head dimension | Current | Alternative | Change |
|---:|---:|---:|---:|---:|---:|---:|
| 1 | 4K | 2/2 | 64 | 0.0738 ms | 0.0931 ms | +26.1% |
| 1 | 4K | 2/2 | 128 | 0.0737 ms | 0.0956 ms | +29.7% |
| 1 | 16K | 2/2 | 64 | 0.5029 ms | 0.5019 ms | -0.2% |
| 1 | 16K | 2/2 | 128 | 0.5944 ms | 0.5867 ms | -1.3% |
| 4 | 4K | 32/4 | 64 | 1.6246 ms | 1.6218 ms | -0.2% |
| 4 | 4K | 32/4 | 128 | 2.5050 ms | 2.5134 ms | +0.3% |
| 4 | 16K | 32/4 | 64 | 23.7041 ms | 23.6599 ms | -0.2% |
| 4 | 16K | 32/4 | 128 | 35.7585 ms | 35.7580 ms | 0.0% |

drisspg:
rebase and then I think we land :)

henrylhtsang:
> rebase and then I think we land :)

thanks!

henrylhtsang:
Rebased this PR onto the latest `main` and fixed the varlen autograd return arity to account for the newly added scheduler arguments.

Validation on GB200 (CUDA 13, CUTLASS DSL 4.6.0.dev0):

```bash
CUDA_VISIBLE_DEVICES=1 FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1 conda run -n flash pytest -q tests/cute/test_flash_attn.py -k learnable_sink
```

Result: `6 passed, 415807 deselected`.

drisspg (approved):

PR #2712, opened 2026-07-16
chore: set stable cutlass 4.6.0 version for FA4

NanoCode012:
There's upstream 4.6.0 for cutlass now. This also makes it compatible with latest sonicmoe which has also integrated 4.6.0. This will also then not require `pre-release` flag if using with `uv`.

Note: I've only done some light test with this (for run with Axolotl downstream) and not full expansive checks across whole FA4.

### DIFF
--- a/flash_attn/cute/pyproject.toml
+++ b/flash_attn/cute/pyproject.toml
@@ -22,7 +22,7 @@ classifiers = [
 ]
 
 dependencies = [
-    "nvidia-cutlass-dsl==4.6.0.dev0",
+    "nvidia-cutlass-dsl==4.6.0",
     "torch",
     "einops",
     "typing_extensions",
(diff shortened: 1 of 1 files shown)


jayhshah:
PR #2689 is handling the cutlass version bump.

Johnsonms:
I verified this today, it won't work unless with #2689

NanoCode012:
Thank you both for linking that PR, wasn't aware it existed. Closing this PR.

NanoCode012:
@Johnsonms , could I ask what you ran to verify/test?

In my case, I just ran a test again with quack 0.6.0 and cutlass 4.6.0, and I was able to use `flash-attn-4 4.0.0b22` with Axolotl. H100 SXM

Johnsonms:
Hi @NanoCode012 I used B200 for this. The key issue is related https://github.com/Dao-AILab/flash-attention/pull/2689, we need a decision for upgrade or postpone. Thanks for your PR

Issue #2577, opened 2026-05-21
[Cute,Fwd,Sm100] FP8 precision degradation at KV tile boundaries due to rescale_threshold


Robin930:
## Summary

FA4 FP8 (e4m3) forward pass on SM100 exhibits severe precision degradation (cosine similarity dropping to ~0.86) at KV sequence length tile boundaries (multiples of 128). The root cause is `rescale_threshold=4.0` interacting with `max_offset=8`, allowing intermediate P values up to 2^12 = 4096 which overflow FP8 e4m3 max (448).

## Environment

- GPU: NVIDIA B300 (SM100f)
- Flash Attention: main branch @ `4178915`
- FP8 support introduced in: `b65ae6b` (#2109)
- Test config: batch=1, q_seqlen=1, kv_seqlen=1~1023, num_heads=6 Q / 1 KV (GQA), head_dim=128, paged KV (page_size=128)

## Problem

When running FP8 paged decode attention, cosine similarity vs BF16 reference drops sharply at every 128-token boundary:
```
FA4 FP8 vs BF16 kv_seqlen=127: max_diff=0.032471  mean_diff=0.005810  cos_sim=0.99866372
FA4 FP8 vs BF16 kv_seqlen=128: max_diff=0.300781  mean_diff=0.047027  cos_sim=0.86276174
FA4 FP8 vs BF16 kv_seqlen=129: max_diff=0.302246  mean_diff=0.024087  cos_sim=0.95049518
FA4 FP8 vs BF16 kv_seqlen=130: max_diff=0.268127  mean_diff=0.038082  cos_sim=0.94660372
```

<img width="1800" height="750" alt="Image" src="https://github.com/user-attachments/assets/1155a325-9503-49ac-96b8-5384eeaf55ad" />

## Root Cause

In `flash_attn/cute/flash_fwd_sm100.py` (L2002-2005, as of `4178915`):

```python
rescale_threshold = (
    8.0 if const_expr(self.q_dtype.width == 16) else
    4.0 if const_expr(self.q_dtype.width == 8) else  # BUG
    0.0
)
```
Combined with max_offset = 8 (L1994), the effective exponent of P before FP8 cast is:
```
P_exp = (S - row_max) * scale_log2 + max_offset
```
When rescale_threshold=4.0, the softmax's update_row_max skips the O-accumulator rescale if the row_max change is within threshold:

```
[...]

PR #2731, opened 2026-07-25
Numeric tweaks to fp8

drisspg:
Numeric tweaks to fp8 rescale and offset

Addressing: https://github.com/Dao-AILab/flash-attention/issues/2577 

This however has some tradeoffs ill show. And this is choice is not yet finalized;

## Summary

The summary in that issue is pretty much spot on and I agree with its fix although there is some interesting perf/ numeric trade offs, and fundamentally the optimal choice is data dependent so we might wanna think about exposing this somehow.

TLDR: With `offset=8` and `rescale_threshold=4`, the max intermediate value in the first `P` block is `2^8 = 256`. A later block can have an unshifted max up to 3.999 log2 units above the tracked max. Because this increase is below the threshold, we keep the old `M` and skip rescaling the accumulator. The second block's max intermediate `P` value can then approach `2^(8+4) = 4096`.

This exceeds E4M3's max value of 448. The denominator `l_i` is accumulated in FP32, but `P` is cast to E4M3 before `P @ V`, causing these large values to saturate. The denominator includes the original values while the numerator includes only their clipped contributions. Final normalization therefore can destroy these values that should contribute to the output

lets say we set rescale threshold to zero,  then any increase in the block max also updates `M`. The  previous accumulator and `l_i` are rescaled by `alpha = 2^(M_old - M_new)`, and the new block is evaluated relative to `M_new`. Its max intermediate `P` value is therefore `2^8 = 256`, rather than 4096.

The current Pr uses a threshold of `0.75`; it allows the maximum to reach `2^8.75 ≈ 431`, which remains below 448, while still avoiding some accumulator rescales, but there are many ways to accomplish the same affect

### Perf Numeric Summary

-- agent notes -- 

#### Pareto frontier

E4M3 offset-threshold Pareto

The full confirmation supports **`max_offset=8, rescale_threshold=0`** as the best global policy measured. Relative to upstream `(4,4)`:

| Path | Median latency | Time-weighted latency | Median relative-L2 / upstream |
| --- | ---: | ---: | ---: |
| Production decode, auto-SplitKV | +0.11% | +0.01% | 0.9945 |
| Decode diagnostic, forced one split | +0.94% | +0.94% | 0.8917 |
| Paged causal prefill | -2.70% | -3.80% | 0.9382 |

Lower is better. Threshold zero produced the stable prefill improvement. Offsets 6–8 had no reproducible performance ordering in the focused recheck, so offset 8 is preferred for its wider retained probability-tail range and slightly better aggregate prefill numerics. The tradeoff is roughly 1% slower forced-one-split decode; production auto-SplitKV decode remained neutral.

> **Implementation note:** the current diff uses `(8, 0)` for FP8 inputs, matching the measured global policy.

#### Why `(8, 0)` vs `(8, 0.75)`

`0.75` does genuinely help decode: it skips accumulator corrections when the running maximum increases by less than the threshold. Relative to `(8,0)`, `(8,0.75)` was **0.12% faster time-weighted** across the broad 72-case auto-SplitKV decode set and **0.61% faster** across the 72 forced-one-split cases. In a focused two-case long-context recheck, the gains were larger—**1.67%** and **4.26%**, respectively—showing that correction cost can be exposed in long unsplit decode.

Threshold zero, however, is a distinct compile-time specialization that removes the positive-threshold compare/select sequence. A codegen isolation test compared `0` with `1e-6`: their BF16 outputs matched bit-for-bit in **8/8** cases, but `1e-6` was **7.66% slower time-weighted** for prefill. `0.75` recovered some work by skipping real corrections, but remained **6.24% slower** than zero for prefill. It also had slightly worse aggregate numerics than zero.

I am going to stick with (8, 0) for now but int eh future this should maybe be user configurable .. 

#### Methodology

- **Discovery:** 38 distinct offset/threshold pairs, 54 matched cases each — **2,052 cells**.
- **Confirmation:** 9 selected pairs, 216 matched cases each — **1,944 cells**.
- **Focused recheck:** 5 finalists across 8 long-context cases with rotated policy order — **40 cells**.
- **Total:** **4,036 matched policy/workload cells**.
- Captured Qwen2.5-1.5B and Qwen2.5-7B attention tensors using C4 and instruction/GSM8K text, layers 0/16/27, and realistic MHA/GQA geometry.
- Paged causal decode used `Q=1`, `SK={129,512,2048,8192}`, both auto-SplitKV and forced one split. Paged causal prefill used `Q=K`, `S={512,2048,4096,8192}`.
- NVIDIA B200 with SM clock locked to 1500 MHz. Warm fixed-pointer CUDA-graph timing; quantization/setup excluded.
- Numerics compare FA4 against FP32 attention over the exact dequantized vLLM-style E4M3 inputs. This measures isolated FA4 attention, not end-to-end vLLM serving.

<details>
<summary>Full 2D discovery heatmaps</summary>

Green is better than upstream `(4,4)`. Unsafe pairs are retained as diagnostic controls and show the expected rapid saturation.

E4M3 offset-threshold heatmaps

</details>

## E5M2 follow-up

E5M2 offset-threshold Pareto

I repeated the performance/numerics sweep with the captured inputs quantized to E5M2. This supports retaining `max_offset=8` and changing E5M2 from `rescale_threshold=4` to **`rescale_threshold=0`**. Relative to inherited `(8,4)`:

| Path | Median latency | Time-weighted latency | Median relative-L2 / `(8,4)` |
| --- | ---: | ---: | ---: |
| Production decode, auto-SplitKV | -0.52% | -0.55% | 1.0000 |
| Decode diagnostic, forced one split | +0.08% | +0.11% | 0.9191 |
| Paged causal prefill | -3.49% | -4.17% | 0.9270 |

Threshold zero again removes the positive-threshold compare/select path and improves prefill while remaining slightly faster for production auto-SplitKV decode. A focused 8K forced-one-split decode recheck exposed the path-specific downside more clearly: `(8,0)` was about **3.1% slower** than `(8,4)` there.

E5M2's extra exponent range is still being used at offset 8: its effective probability floor is `2^-24`, versus `2^-17` for E4M3 at the same offset. Threshold-zero offsets 8, 12, and 15 produced essentially identical numerics, so spending more saturation headroom on a larger offset did not help. The grid did not test offsets below 8; this supports retaining the existing offset, not claiming that 8 is the global E5M2 optimum for every context length.

#### E5M2 methodology

- **Discovery:** 40 pairs (34 safe candidates plus 6 saturation controls) × 54 cases — **2,160 cells**.
- **Confirmation:** 8 selected pairs × 216 cases — **1,728 cells**.
- **Focused recheck:** 6 policies × 8 long-context cases — **48 cells**.
- **Total:** **3,936 matched policy/workload cells**.
- Same Qwen2.5-1.5B/Qwen2.5-7B captures, prefill/decode shapes, B200 timing contract, and FP32-reference methodology as the E4M3 study; only the input quantization format changed to E5M2.

<details>
<summary>Full E5M2 2D discovery heatmaps</summary>

E5M2 offset-threshold heatmaps

</details>

### DIFF
--- a/flash_attn/cute/flash_fwd_sm100.py
+++ b/flash_attn/cute/flash_fwd_sm100.py
@@ -80,6 +80,22 @@ from flash_attn.cute.utils import AuxData
 #   num_regs_correction: int — register count for correction warps (multiple of 8)
 #   num_regs_other is derived: 512 - num_regs_softmax * 2 - num_regs_correction
 #                  (hd256 exception: num_regs_other is fixed at 32, not derived)
+
+# Note [Low Precision Scaling]
+# P is in (0, 1] and is cast to the input dtype before P @ V, so scaling it by 2^max_offset
+# spends the dtype's unused upper code points on the probability tail. A positive
+# rescale_threshold lets the row max lag by that many log2 units, so P can reach
+# 2^(max_offset + rescale_threshold); above the dtype max the top probabilities saturate
+# while the FP32 denominator still counts them in full, shrinking the output (#2716).
+
+# log2 of the largest finite value representable in each supported input dtype.
+_LOG2_DTYPE_MAX = {
+    cutlass.Float8E4M3FN: math.log2(448.0),
+    cutlass.Float8E5M2: math.log2(57344.0),
+    cutlass.Float16: math.log2(65504.0),
+    cutlass.BFloat16: math.log2(3.3895313892515355e38),
+}
+
 _TUNING_CONFIG = {
     (True, False, 128, False): {"ex2_emu_freq": 10, "ex2_emu_start_frg": 1, "num_regs_softmax": 176, "num_regs_correction": 88},
     (False, True, 128, False): {"ex2_emu_freq": 16, "ex2_emu_start_frg": 1, "num_regs_softmax": 192, "num_regs_correction": 72},
@@ -2027,16 +2043,8 @@ class FlashAttentionForwardSm100:
 
             qk_descale, _ = self._load_effective_descales(descale_tensors, batch_idx, kv_head_idx)
 
-            # P is scaled by 2^max_offset before the FP8 conversion. With rescale_threshold > 0
-            # the row max can be stale by up to rescale_threshold (in log2 units), so P can reach
-            # 2^(max_offset + rescale_threshold). max_offset + rescale_threshold must stay within
-            # log2(fp8_max) (448 = 2^8.8 for e4m3fn, 57344 = 2^15.8 for e5m2), otherwise the
-            # largest probabilities saturate and accuracy degrades (#2716).
-            max_offset = (
-                4 if cutlass.const_expr(self.q_dtype is cutlass.Float8E4M3FN) else
-                8 if cutlass.const_expr(self.q_dtype.width == 8) else
-                0
-            )
+            # See Note [Low Precision Scaling]
+            max_offset = 8 if cutlass.const_expr(self.q_dtype.width == 8) else 0
             if const_expr(self.score_mod is None):
                 softmax_scale_log2_eff = softmax_scale_log2 * qk_descale
                 softmax_scale_eff = None
@@ -2044,10 +2052,11 @@ class FlashAttentionForwardSm100:
                 softmax_scale_log2_eff = softmax_scale_log2
                 softmax_scale_eff = softmax_scale * qk_descale
 
-            rescale_threshold = (
-                8.0 if const_expr(self.q_dtype.width == 16) else
-                4.0 if const_expr(self.q_dtype.width == 8) else
-                0.0
+            rescale_threshold = 8.0 if const_expr(self.q_dtype.width == 16) else 0.0
+            # See Note [Low Precision Scaling]
+            assert max_offset + rescale_threshold < _LOG2_DTYPE_MAX[self.q_dtype], (
+                f"max_offset ({max_offset}) + rescale_threshold ({rescale_threshold}) must stay "
+                f"below log2(max {self.q_dtype} value) to avoid saturating P"
             )
             softmax = SoftmaxSm100.create(
                 softmax_scale_log2_eff,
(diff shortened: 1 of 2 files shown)


MatthewBonanni:
0.75 is what I used here and it works well: https://github.com/vllm-project/flash-attention/pull/166

drisspg:
@MatthewBonanni hey thanks for pinging, I did a bunch of tests and came to a slightly different conclusion.. although I think we can re-assess. I am working on a fwd config like system and honestly this feels like it should be user overridable in the limit

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -2046,6 +2040,7 @@ def softmax_loop(
 
             rescale_threshold = (
                 8.0 if const_expr(self.q_dtype.width == 16) else
+                0.0 if const_expr(self.q_dtype is cutlass.Float8E4M3FN) else
```
drisspg: let me clean this up..

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -2027,28 +2027,18 @@ def softmax_loop(
...
             else:
                 softmax_scale_log2_eff = softmax_scale_log2
                 softmax_scale_eff = softmax_scale * qk_descale
 
-            rescale_threshold = (
-                8.0 if const_expr(self.q_dtype.width == 16) else
-                4.0 if const_expr(self.q_dtype.width == 8) else
-                0.0
-            )
+            rescale_threshold = 8.0 if const_expr(self.q_dtype.width == 16) else 0.0
```
v0i0, line 2055: would be nice to assert max_offset+reset_threshold < max_value[dtype], too

v0i0 (approved):

flash_attn/cute/flash_fwd_sm100.py
drisspg: ohh great idea :)

PR #2732, opened 2026-07-25
[CuTe] Add typed forward configurations

drisspg:
Stacked PRs:
 * #2743
 * #2741
 * #2738
 * #2735
 * #2733
 * __->__#2732

--- --- ---

# Human Note

# Agent Note

## Summary

This is the small config layer I want before we tune FA4. A forward launch is now one immutable, fully resolved `FwdConfig`; the selector gets a separate `FwdHeuristicInputs`; and only the main/combine codegen projections key compilation.

That separation is the important bit: runtime shapes can choose a policy without raw shapes becoming compile-cache keys, while an explicit config can still force one exact kernel for A/B testing. Register allocation is now one hashable architecture union inside that config: `FwdSm90RegisterAllocation(mma, producer) | FwdSm100RegisterAllocation(softmax, correction, other)`, rather than unrelated values or hidden kernel tables.

The keys do not carry mirrored facts. Varlen mode is derived once from the existing Q/K cumulative-length and `seqused` presence bits; QV presence is implied by the MLA kernel family; and combine tile geometry is derived from head dimension. Shape changes therefore key policy selection without duplicating equivalent cache dimensions.

Typed FA4 forward config boundary

There is deliberately no scheduling change in this PR. It only makes the existing choices typed, visible, and testable.

## Host representation cost

`FwdConfig` remains a frozen dataclass because it is the user-facing resolved configuration. The three key-only records are tuple-backed `NamedTuple`s: they remain typed and immutable while avoiding frozen-dataclass construction and hashing on every launch.

On exact pre/post stack heads (`69285d1c`→`56161730`), host-only medians improved as follows:

| Operation | Before | After |
|---|---:|---:|
| Construct selector inputs | 4.103 us | 1.135 us |
| Construct inputs + warm selector hit | 5.734 us | 1.378 us |
| Project main kernel spec | 10.245 us | 5.939 us |
| Hash main spec | 0.816 us | 0.237 us |
| Construct combine spec | 1.161 us | 0.227 us |

This changes only Python representation; equality, persistent-cache identity fields, selector policy, and generated kernels are unchanged. A final simplifier pass removed the mirrored varlen/QV/combine-geometry fields and consolidated architecture dispatch. Against the preceding submitted stack, cached selection measured 190.2→187.1 ns and input construction 1.103→1.080 us; selector-miss body cost remained within 1% run-to-run noise.

Successful explicit-config validation is also cached by immutable `(config, inputs)` with a bounded 1024-entry cache: repeated validation measured 1.350→0.354 us. Invalid configurations still execute validation and raise because exceptions are not cached. Combine split bucketing now uses exact integer `bit_length()` rather than floating `log2/ceil`; all split counts 1–256 produce identical buckets.

## Implementation and validation

- `FwdConfig` contains the resolved tile, stage/thread, scheduler, TMA-O, 2CTA, SplitKV, and applicable SM90 or SM100 warp-group register allocation.
- The selector preserves the existing SM90 1/2/3-MMA-warp-group and cp.async register choices, plus the existing BF16, FP16, FP8, paged, SM103, and fixed-HD256 choices. MLA, SM80, and SM120 carry `registers=None`.
- Register validation enforces the legal `setmaxnreg` direction, 8-register granularity, per-role bounds, and weighted 512-register budget.
- `FwdHeuristicInputs` contains host-visible metadata only; selection does not synchronize, inspect tensor contents, or allocate outputs.
- `FwdMainKernelSpec` and `FwdCombineKernelSpec` contain only code-generation-changing fields; register-only changes produce distinct main specializations.
- Main-kernel identity includes static tensor broadcast patterns but never concrete B/Q/K lengths. Combine identity includes the optional dynamic-split and semaphore operands that gate compile-time branches.
- Invalid explicit configs fail instead of being silently normalized.
- Fixed `seqlen_k_per_split` is selector input: automatic selection resolves the exact required split capacity, while explicit configs below it fail before scheduler metadata allocation.
- Dedicated MLA and head-dim-256 paths keep their canonical fixed configs; automatic SplitKV cannot leak into HD256 decode shapes.
- Dynamic-shape behavior: a shape change that maps to the same typed spec reuses the cache; crossing a codegen boundary compiles that specialization once.

Validation lives in `tests/cute/test_fwd_config.py`, including architecture defaults, exact forced configs, SplitKV resolution, dedicated paths, cache projections, JSON/YAML round trips, register-policy parity, and invalid combinations. The standalone #2732 boundary passes 60 host tests. An exhaustive differential audit matched the old SM100 kernel-side register policy across 110,592 dtype/architecture/head-dimension/paging/causality/CTA combinations, and a table-driven audit covers every SM90 MMA-warp-group and TMA/cp.async branch.

### DIFF
--- a/flash_attn/cute/config.py
+++ b/flash_attn/cute/config.py
@@ -1,0 +1,846 @@
+# Copyright (c) 2025, Jay Shah, Ganesh Bikshandi, Ying Zhang, Vijay Thakkar, Pradeep Ramani, Tri Dao.
+
+"""Typed forward configuration, selection, validation, and compile keys.
+
+`FwdHeuristicInputs` contains immutable host-visible metadata. `FwdConfig` is
+fully resolved: one split means the nonsplit path and larger values are exact
+SplitKV counts. Explicit configs are validated without ranking or normalization.
+The main and combine kernel specs independently project only codegen-changing
+values. Selection must not read tensor contents, synchronize CUDA, or allocate.
+"""
+
+import math
+from dataclasses import dataclass, replace
+from functools import lru_cache
+from typing import NamedTuple
+
+
+class FwdSm90RegisterAllocation(NamedTuple):
+    """SM90 per-thread register allocation by warp-group role."""
+
+    mma: int
+    producer: int
+
+
+class FwdSm100RegisterAllocation(NamedTuple):
+    """SM100 per-thread register allocation by warp-group role."""
+
+    softmax: int
+    correction: int
+    other: int
+
+
+FwdRegisterAllocation = FwdSm90RegisterAllocation | FwdSm100RegisterAllocation
+
+
+@dataclass(frozen=True)
+class FwdConfig:
+    """Fully resolved forward launch and algorithm configuration.
+
+    ``is_static_persistent`` controls static persistence; dynamic persistence is
+    derived from scheduler metadata. ``registers`` names the warp-group roles
+    exposed by the selected architecture and is ``None`` for other kernels.
+    """
+
+    device_capacity: int
+    tile_m: int
+    tile_n: int
+    num_stages: int
+    num_threads: int
+    mma_pv_is_rs: bool
+    intra_wg_overlap: bool
+    q_stage: int
+    use_clc_scheduler: bool
+    is_static_persistent: bool
+    use_tma_o: bool
+    num_splits: int
+    use_2cta_instrs: bool
+    registers: FwdRegisterAllocation | None
+
+    def __post_init__(self) -> None:
+        # JSON turns NamedTuple into a list; campaign YAML may use a mapping.
+        if self.registers is None or isinstance(
+            self.registers, (FwdSm90RegisterAllocation, FwdSm100RegisterAllocation)
+        ):
+            return
+        register_type = {
+            9: FwdSm90RegisterAllocation,
+            10: FwdSm100RegisterAllocation,
+            11: FwdSm100RegisterAllocation,
+        }.get(self.device_capacity)
+        if register_type is None:
+            raise ValueError(f"SM{self.device_capacity} does not expose register allocation")
+        registers = (
+            register_type(**self.registers)
+            if isinstance(self.registers, dict)
+            else register_type(*self.registers)
+        )
+        object.__setattr__(self, "registers", registers)
+
+
+class FwdHeuristicInputs(NamedTuple):
+    """Host-visible metadata used to select and validate a forward config."""
+
+    device_arch: int
+    num_sms: int
+    dtype: str
+    head_dim: int
+    head_dim_v: int
+    num_heads: int
+    num_heads_kv: int
+    batch_size: int
+    total_q: int
+    total_k: int
+    max_seqlen_q: int
+    max_seqlen_k: int
+    seqlen_k_per_split: int | None
+    causal: bool
+    local: bool
+    window_size_left: int | None
+    window_size_right: int | None
+    is_varlen_q: bool
+    has_cu_seqlens_q: bool
+    has_cu_seqlens_k: bool
+    has_seqused: bool
+    pack_gqa: bool
+    page_size: int | None
+    use_block_sparsity: bool
+    sparse_q_block_size: int | None
+    has_qv: bool
+    has_gather_kv: bool
+    has_score_mod: bool
+    has_mask_mod: bool
+    has_learnable_sink: bool
+    has_lse: bool
+    requested_tile_m: int | None
+    requested_tile_n: int | None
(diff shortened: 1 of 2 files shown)


flash_attn/cute/config.py
```diff
@@ -0,0 +1,768 @@
...
+    def head_dim_padded(self) -> int:
+        return math.ceil(self.head_dim / 16) * 16
+
+    @property
+    def head_dim_v_padded(self) -> int:
+        return math.ceil(self.head_dim_v / 16) * 16
+
+
+@dataclass(frozen=True)
+class FwdMainKernelConfigSpec:
```
drisspg: hmm why do we need this

flash_attn/cute/config.py
```diff
@@ -0,0 +1,761 @@
...
+values. Selection must not read tensor contents, synchronize CUDA, or allocate.
+"""
+
+import math
+from dataclasses import dataclass, replace
+from functools import lru_cache
+from typing import NamedTuple
+
+
+class FwdRegisterAllocation(NamedTuple):
```
reubenconducts: Should we have register allocation for sm90?

flash_attn/cute/config.py
```diff
@@ -0,0 +1,761 @@
...
+def fwd_config_compile_bucket(config: FwdConfig, head_dim_v: int) -> tuple[FwdConfig, int | None]:
+    """Deduplicate exact split counts that share main and combine codegen."""
+    main = replace(config, num_splits=2 if config.num_splits > 1 else 1)
+    if config.num_splits == 1:
+        return main, None
+    tile_m, _ = fwd_combine_tile(head_dim_v)
+    return main, combine_log_max_splits(config.num_splits, tile_m)
+
+
+def num_splits_heuristic(
```
reubenconducts, line 283: We should add `seqlen_k_per_split` to the heuristic; I left it out by accident in #2559.

flash_attn/cute/config.py
```diff
@@ -0,0 +1,762 @@
...
+    dtype_partial: object
+    head_dim: int
+    tile_m: int
+    k_block_size: int
+    log_max_splits: int
+    has_cu_seqlens: bool
+    has_seqused: bool
+    has_lse: bool
+    has_num_splits_dynamic_ptr: bool
+    has_varlen_batch_idx: bool
```
reubenconducts, line 255: Now named `virtual_batch_idx`

flash_attn/cute/config.py
drisspg: Yes — added this. `FwdConfig.registers` is now the architecture-specific union `FwdSm90RegisterAllocation(mma, producer) | FwdSm100RegisterAllocation(softmax, correction, other)`. The selector preserves the old SM90 1/2/3-MMA-warp-group choices and the 2-WG cp.async override; validation checks `setmaxnreg` direction, granularity, and the weighted register budget. The next PR routes the resolved SM90 value into the kernel and removes the hidden table. I also covered every old SM90 policy branch and fake-compiled both the default and an alternate explicit allocation.

flash_attn/cute/config.py
drisspg: Added across #2732/#2733.

`seqlen_k_per_split` is now part of `FwdHeuristicInputs`. For an automatic request, the selector resolves the exact static capacity required by `ceil(max_seqlen_k / seqlen_k_per_split)`; an explicit `FwdConfig` with fewer splits is rejected before scheduler metadata allocation, and the selected tile still validates the existing divisibility requirement. The standalone `get_scheduler_metadata` path includes `seqlen_k_new` in that bound.

I added one focused host regression (auto resolves 5 splits for K=4224 and 1024/split, explicit 4 is rejected) and changed the existing fixed-K end-to-end test to use `num_splits=0`. The latter passes both causal modes in fake compilation and on GB300.

flash_attn/cute/config.py
drisspg: Updated to `has_virtual_batch_idx` throughout the main/combine specs, compiler projection, interface, and regression names.

flash_attn/cute/config.py
```diff
@@ -0,0 +1,865 @@
...
+    num_splits: int
+    use_2cta_instrs: bool
+    registers: FwdRegisterAllocation | None
+
+    def __post_init__(self) -> None:
+        # JSON turns NamedTuple into a list; campaign YAML may use a mapping.
+        if self.registers is None or isinstance(
+            self.registers, (FwdSm90RegisterAllocation, FwdSm100RegisterAllocation)
+        ):
+            return
```
drisspg, line 65: how do we handle if someone uses sm120 for instance

PR #2745, opened 2026-07-29
[CuTe] Fix forward dynamic-shape correctness

drisspg:
Stacked PRs:
 * #2743
 * #2741
 * #2738
 * #2735
 * #2733
 * #2732
 * __->__#2745

--- --- ---

# Human Note

This is baiscally all small tweaks to our dynamic shape infra; I did a lot of fuzzing of the stack to get a good sense for when recompiles happen and passing in a bunch of configs. These are all the result of that. While I was here I did smallish tweaks to those helpers to remove host overhead where applicable 

# Agent Note

## Summary

This is the small correctness base for the rest of the stack. Dynamic-shape and layout fuzzing found seven pre-existing forward bugs; this PR fixes them without adding configs, tuning APIs, or selector policy.

- Canonicalize unsupported strides and effective pointer alignment before TVM-FFI launch; reject broadcast output/workspace layouts.
- Include static tensor broadcast patterns in the main compile key.
- Key aux tensors by dtype, assumed alignment, and static stride modes.
- Compile SplitKV dynamic-split and semaphore operands exactly as called while retaining the lightweight existing runtime checks.
- Use matching target-SKU SM metadata during fake selection, with an explicit cross-compilation override.
- Size aliased SM100 K/V shared storage for the larger staged layout.
- Ceil-divide non-TMA paged-loader entries so partial row waves receive page pointers.

The cache fixes still keep concrete batch and sequence lengths dynamic.

## Host-path cost

The initial fuzz fix was correct but too defensive on the eager hot path. This revision keeps the same cache identities and kernel behavior while removing work unrelated to the seven root bugs:

- contiguous input/output layouts return before scanning every stride;
- no-broadcast metadata uses a C-level zero-stride fast path;
- the added same-device pass was removed, leaving the pre-existing CUDA-input contract untouched;
- internal combine no longer repeats rank/shape/dtype/device/layout checks after `_flash_attn_fwd` has allocated or validated those tensors;
- target-SM discovery runs only when automatic or rewritten diff-head SplitKV actually needs it;
- fake-mode state is resolved once per runtime call, and fake stream/aux CuTe conversion work runs only on compile misses;
- inference and training retain the original unconditional tensor detach behavior.

Nightly host microbenchmarks on the same GB300 machine:

| Host path | Before | After | Saved |
|---|---:|---:|---:|
| Q/K/V canonicalization | 1.50 us | 0.75 us | 0.75 us |
| Explicit out + LSE layout checks | 0.92 us | 0.75 us | 0.17 us |
| Common forward broadcast metadata | 1.70 us | 1.00 us | 0.70 us |
| Internal combine validation | 3.971 us | 1.168 us | 2.803 us |
| Added same-device pass | +0.94 us | removed | 0.94 us |
| Fixed-split target-SM lookup | 2.28 us | skipped | 2.28 us |

These are host-only measurements; no selector policy or generated kernel changed.

For a whole-wrapper check, I used real preallocated CUDA tensors, a warm hit-only compile cache, and a no-op compiled callable. This measures all Python dispatch through the point where the generated kernel would launch, without mixing in GPU execution:

| Shape | `main` | This PR | Net cost |
|---|---:|---:|---:|
| Decode: B1, Q1, K128, H32/Hkv8, D128 | 16.951 us | 12.922 us | -4.029 us (23.8%) |
| Short: B2, Q64, K128, H16, D64 | 16.608 us | 12.669 us | -3.939 us (23.7%) |
| Prefill: B1, Q257, K2048, H32, D64 | 16.993 us | 12.873 us | -4.120 us (24.2%) |

Despite adding the correctness checks, this PR is 3.9–4.1 us (23.7–24.2%) faster than `main` on the warm host path because it removes unconditional target-SM discovery, reuses fake-mode state, defers compile-only setup, and uses fast common-layout metadata paths. Real no-graph runs with the actual kernel also showed no latency regression.

## Validation

Focused host tests cover target-SM fallback, CuTe aux-key equivalence, external-base pointer alignment, singleton broadcast-output rejection, and explicit-architecture fake SplitKV. The combine regression now exercises one cache in the order neither optional operand → dynamic splits only → semaphore only → both, checking four specializations and numerical results. The existing unaligned-input test now checks public forward and `dq/dk/dv` for both padded-stride and unaligned-offset views.

The targeted GPU regressions pass on GB300/SM103 against independent FP32 references where applicable. After rebasing onto the varlen dynamic-persistent scheduler in `c46b8144`, 11 bottom-boundary GPU cases passed, including metadata reuse, fixed K-per-split, virtual-batch combine, dynamic combine operands, unaligned backward, paged tail, and aux-layout cache separation. Compute Sanitizer previously reported `ERROR SUMMARY: 0 errors` across the wider q-stage, paged-tail, SplitKV, and `DV > D` matrix. Ruff, formatting, `py_compile`, and `git diff --check` pass.

### DIFF
--- a/flash_attn/cute/cute_dsl_utils.py
+++ b/flash_attn/cute/cute_dsl_utils.py
@@ -1,9 +1,11 @@
 # Copyright (c) 2025, Tri Dao.
 
+import os
 from typing import Tuple
 from functools import lru_cache
 
 import torch
+from torch._subclasses.fake_tensor import FakeTensor
 
 try:
     from triton.tools.disasm import extract
@@ -39,6 +41,75 @@ def get_max_active_clusters(cluster_size):
 @lru_cache
 def get_device_capacity(device: torch.device = None) -> Tuple[int, int]:
     return torch.cuda.get_device_capability(device)
+
+
+@lru_cache
+def _get_device_arch_and_num_sms(device_index: int) -> tuple[int, int]:
+    properties = torch.cuda.get_device_properties(device_index)
+    return properties.major * 10 + properties.minor, properties.multi_processor_count
+
+
+def get_num_sms_for_selection(device_index: int, arch: int) -> int:
+    """Return the SM count of the matching local GPU or cross-compilation target."""
+    override = os.getenv("FLASH_ATTENTION_NUM_SMS")
+    if override is not None:
+        num_sms = int(override)
+        if num_sms <= 0:
+            raise ValueError("FLASH_ATTENTION_NUM_SMS must be positive")
+        return num_sms
+    if torch.cuda.is_available():
+        device_arch, num_sms = _get_device_arch_and_num_sms(device_index)
+        if device_arch == arch:
+            return num_sms
+    raise RuntimeError(
+        "Cannot determine the target GPU's SM count; set FLASH_ATTENTION_NUM_SMS "
+        "when cross-compiling without a matching local GPU"
+    )
+
+
+def _has_aligned_pointer(tensor: torch.Tensor, align_bytes: int) -> bool:
+    address = (
+        tensor.storage_offset() * tensor.element_size()
+        if isinstance(tensor, FakeTensor)
+        else tensor.data_ptr()
+    )
+    return address % align_bytes == 0
+
+
+def _is_aligned_layout(tensor: torch.Tensor, align_bytes: int) -> bool:
+    """Return whether a tensor satisfies the pointer and stride ABI kernels assume."""
+    if tensor.stride(-1) != 1 or not _has_aligned_pointer(tensor, align_bytes):
+        return False
+    stride_alignment = max(1, align_bytes // tensor.element_size())
+    return all(stride == 0 or stride % stride_alignment == 0 for stride in tensor.stride()[:-1])
+
+
+def maybe_contiguous(tensor: torch.Tensor | None, align_bytes: int = 16):
+    """Canonicalize inputs to the pointer and stride alignment kernels assume."""
+    if tensor is None:
+        return None
+    if tensor.is_contiguous():
+        return (
+            tensor
+            if _has_aligned_pointer(tensor, align_bytes)
+            else tensor.clone(memory_format=torch.contiguous_format)
+        )
+    if not _has_aligned_pointer(tensor, align_bytes):
+        return tensor.clone(memory_format=torch.contiguous_format)
+    return tensor if _is_aligned_layout(tensor, align_bytes) else tensor.contiguous()
+
+
+def validate_output_layout(tensor: torch.Tensor, name: str, align_bytes: int) -> None:
+    """Validate a caller-provided output or SplitKV workspace."""
+    assert 0 not in tensor.stride(), f"{name} must not have broadcast dimensions"
+    if tensor.is_contiguous():
+        assert _has_aligned_pointer(tensor, align_bytes), (
+            f"{name} must have aligned strides and a contiguous last dimension"
+        )
+        return
+    assert _is_aligned_layout(tensor, align_bytes), (
+        f"{name} must have aligned strides and a contiguous last dimension"
+    )
 
 
 def assume_strides_aligned(t):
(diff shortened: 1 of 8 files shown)


flash_attn/cute/cute_dsl_utils.py
```diff
@@ -84,33 +84,49 @@ def to_cute_tensor(t, assumed_align=16, leading_dim=-1, fully_dynamic=False, ena
...
 
-def to_cute_aux_tensor(t, enable_tvm_ffi=True):
-    """Convert torch tensor to cute tensor for TVM FFI, tailored to FlexAttention aux tensors.
-    This allows the user to specify alignment and leading dimension for aux tensors used in
-    custom score_mod callables.
-    """
-    assumed_align: int = getattr(t, "__assumed_align__", None)
-    leading_dim: int = getattr(t, "__leading_dim__", None)
-    fully_dynamic: bool = leading_dim is None
+def _resolve_aux_leading_dim(tensor: torch.Tensor) -> int | None:
```
drisspg, line 176: this is annoying but I measured locally and appears to be the fastest way to get the semantics we want

flash_attn/cute/cute_dsl_utils.py
```diff
@@ -84,33 +84,49 @@ def to_cute_tensor(t, assumed_align=16, leading_dim=-1, fully_dynamic=False, ena
...
     )
 
 
 def get_aux_tensor_metadata(aux_tensors):
-    return tuple(
-        (
-            getattr(t, "__assumed_align__", 0),
-            getattr(t, "__leading_dim__", -1),
-            hasattr(t, "__leading_dim__"),
+    """Return the static aux-tensor ABI facts that must key the compile cache."""
```
drisspg, line 197: double checking if .__cache_key__ will work here

flash_attn/cute/cute_dsl_utils.py
drisspg: Updated: `to_cute_aux_tensor` now delegates leading-mode inference to CuTe again. The local resolver is used only to form the cache key before conversion.

flash_attn/cute/cute_dsl_utils.py
drisspg: `__cache_key__` is now the regression-test oracle, but not the runtime lookup key: converted aux tensors only exist after a cache miss. Converting before every lookup measured ~9–10 us/tensor versus ~1 us/tensor for the host metadata projection in the nightly environment.

flash_attn/cute/interface.py
```diff
@@ -569,10 +570,16 @@ def _flash_attn_fwd(
         aux_scalars: Runtime scalar captures used by score_mod or mask_mod.
     """
     aux_scalars = tuple(aux_scalars) if aux_scalars else None
+    fake_mode = is_fake_mode()
     q, k, v, qv = [maybe_contiguous(t) for t in (q, k, v, qv)]
```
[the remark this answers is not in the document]
drisspg: Good catch. The existing unaligned-backward regression forced `return_lse=True`, which masked this. Gradient need is now captured before input canonicalization, and the same regression exercises the default `return_lse=False` path for both padded-stride and unaligned-offset views. Both forward/backward cases pass on GB300.

flash_attn/cute/flash_fwd_sm100.py
```diff
@@ -746,6 +746,8 @@ def __call__(
             cute.cosize(sQ_layout) if const_expr(not self.overlap_sO_sQ) else
             cutlass.max(cute.cosize(sQ_layout), cute.cosize(sO_layout) * self.o_dtype.width // self.q_dtype.width)
         )
+        # K and V alias the same physical buffer and may have different extents.
+        sKV_size = cutlass.max(cute.cosize(sK_layout), cute.cosize(sV_layout))
```
v0i0, line 750: can K and V have different dtypes?

flash_attn/cute/cute_dsl_utils.py
```diff
@@ -41,6 +43,73 @@ def get_device_capacity(device: torch.device = None) -> Tuple[int, int]:
...
+        num_sms = int(override)
+        if num_sms <= 0:
+            raise ValueError("FLASH_ATTENTION_NUM_SMS must be positive")
+        return num_sms
+    if torch.cuda.is_available():
+        device_arch, num_sms = _get_device_arch_and_num_sms(device_index)
+        if device_arch == arch:
+            return num_sms
+    # Cross-compilation cannot infer the target SKU without an override.
+    return 132
```
v0i0: no fallback > arbitrart fallback imo?

v0i0 (approved):

flash_attn/cute/flash_fwd_sm100.py
drisspg: nope same dtype: https://github.com/Dao-AILab/flash-attention/blob/d30f0c7a8edb14eb693b420b9baedc72cbe2bc73/flash_attn/cute/flash_fwd_sm100.py#L502-L506 and https://github.com/Dao-AILab/flash-attention/blob/d30f0c7a8edb14eb693b420b9baedc72cbe2bc73/flash_attn/cute/interface.py#L630-L636
althoguh maybe we want to have kv different than q for quantized kv cache

flash_attn/cute/cute_dsl_utils.py
drisspg: Agreed. Removed the 132-SM fallback: when there is no matching local GPU, cross-compilation now requires `FLASH_ATTENTION_NUM_SMS` and raises an actionable error otherwise. The focused test covers both the explicit override and missing-target case.

PR #2748, opened 2026-07-31
feat(HD256 cuteDSL): expand SM100 HD256 2CTA training support and refactor kernels to generic style

wangsiyu:
Refactor the dedicated HD256 kernels onto the current generic HD128/main contracts while preserving the HD256 2CTA QK iteration order.

- Support dense and varlen MHA/MQA/GQA forward and backward, causal and local attention, seqused inputs, zero-length inputs, and Q-longer-than-K shapes.
- Add softcap, learnable sink, dLSE, deterministic backward, FlexAttention score/mask modifiers, block-sparse forward and fixed-length backward, and paged-KV forward.
- Align GQA dK/dV accumulation and postprocessing, q-head parallel reduction, preallocated gradients, and LSE-gradient handling with current main.
- Fix empty-Q and K=1 synchronization, warp participation, and bulk dK/dV reduction behavior in the HD256 path.
- Enable the existing main correctness suites for HD256 without changing their public test structure.

Validation:
- 204429 correctness cases passed with zero failures or errors.
- 24 forward/backward benchmark points stayed within the 2% regression gate.
- 120000 forward/backward stress calls per rank completed without a watchdog stall.

Please review this PR, thanks ! @tridao @Johnsonms  @drisspg  @jayhshah

### DIFF
--- a/flash_attn/cute/block_sparse_utils.py
+++ b/flash_attn/cute/block_sparse_utils.py
@@ -672,6 +672,63 @@ def load_block_list_sm100(
     return kv_producer_state
 
 
+@cute.jit
+def load_block_list_sm100_qk_ahead(
+    block_indices: cute.Tensor,
+    block_count,
+    first_block_preloaded: cutlass.Constexpr,
+    load_q_with_first: cutlass.Constexpr,
+    q_stage: cutlass.Constexpr,
+    kv_producer_state,
+    load_Q,
+    load_K,
+    load_V,
+    pipeline_kv,
+):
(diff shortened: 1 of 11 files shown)


wangsiyu:
In this PR, for the purpose of generalization, the fused mask has been deprecated. We reuse the same mask implementation with the `head_dim=128` kernel style instead. As a result, the mask variantion like swa, flex attention can be easily support. @Johnsonms @drisspg @jayhshah  @tridao

Issue #1895, opened 2025-09-17
Request: Optimized attention backend (fwd/bwd) support for GPT-OSS with learnable sink


CPFLAME:
Hi team,
I’d like to ask whether there is forward and backward support for GPT-OSS attention when using a learnable sink.

In GPT-OSS, the attention module includes a learnable sink parameter. As a result, we currently have to train in the eager attention mode. This makes memory usage scale quadratically with sequence length, which prevents training on long contexts. I found that some kernels provide forward support for a learnable sink, but I couldn’t find full backward support, so we still can’t use an optimized backend end-to-end for training.

### Questions:
- Is there an existing or planned roadmap to support learnable-sink attention in both forward and backward for optimized backends on recent GPUs?
- If yes, could you share an ETA or pointers to the relevant branch/PR?
- A minimal example showing how to enable this with GPT-OSS (model/config flags, attention impl selection) would be very helpful.

### Environment (for reference):
Model: GPT-OSS (20B,120B)
GPU: H20

Thanks a lot!

chengtbf:
+1， we also need

htdung167work0:
+1, i also need

Issue #1810, opened 2025-08-13
FA3 attention sinks blackwell sm120


fernandaspets:
Hi why can;t we use gpt-oss models with 50 series cards and rtx pro 6000? It says it requires attention sinks and FA3 and that FA3 is not supported. These cards have been out for 8 months I wonder why is there so little support? Is there 2 blackwell architectures? sm100 and sm120?

tridao:
sounds great, would you like to contribute a PR?

fernandaspets:
> sounds great, would you like to contribute a PR?

yes I have to learn how though i did some research yesterday. its way beyond me but im interested so will try to learn more. not user if this is something that requires years of training but happy to try and learn and contribute. thanks for responding!

rpcarvalheira:
#1683 tries to solve that. However, flash-attention is not simple, neither an easy solution, so, you'll have to help debugging with @loscrossos as all applications have bugs and normally, forks may be outdated, which means you'll have to be patient, since there will be no simple solution or fast fixes.

Try that solution and if you cannot contribute with code, you can always provide issues and report bugs in a well described manner. Since you're using AI, you can already request for LLM's help on that aspect and provide references, logs, context, process to replicate and any additional information needed; these may help the community as well. If you cannot do any of these, it's a good time to think about sponsoring projects. :)

cmp-nct:
@tridao I also found it the hard way, compiled FA3 for 6 hours and then it didn't work for the 5090 and gpt-oss-* 
Wouldn't it be possible to just use the CUDA 8.9 architecture until V12 is supported ? 
So 50xx cards can run, just without optimized kernels ?

JINO-ROHIT:
hi, id like to take a stab at this issue, is anyone working on this? can this be assigned to me?

fernandaspets:
> hi, id like to take a stab at this issue, is anyone working on this? can this be assigned to me?

yes there is a community now for sm120 on discord: https://discord.gg/jQwPtBKVP and then also the repo here https://github.com/local-inference-lab/rtx6kpro

 vllm fork for sm120 with custom kernels. very active new developments every single day

arbi-dev:
[...]

PR #2752, opened 2026-07-31
[CuTe, SM80/SM120] Support learnable_sink (attention sinks) on Ampere/Ada/consumer-Blackwell

arbi-dev:
## Summary

`FlashAttentionForwardSm80.__call__` in `flash_attn/cute/flash_fwd.py` (the SM80/Ampere/Ada
forward kernel — A100, RTX 3090/4090, and other non-datacenter GPUs) hardcoded:

```python
assert learnable_sink is None, "Learnable sink is not supported in this kernel"
```

So any model using attention sinks (e.g. gpt-oss, StreamingLLM-style architectures) had
no FA4 CuTeDSL forward path on Ampere/Ada, and — since `FlashAttentionForwardSm120`
(`flash_attn/cute/flash_fwd_sm120.py`) subclasses `FlashAttentionForwardSm80` without
overriding `__call__`/`kernel` — on consumer-Blackwell (RTX 50-series) either. Only
Hopper (`flash_fwd_sm90.py`) and datacenter-Blackwell (`flash_fwd_sm100.py`) had a real
sink implementation. That means most people self-hosting sink-attention models on
consumer/prosumer hardware (no H100/B200) had no FA4 path at all for this.

`flash_attn/cute/softmax.py`'s `Softmax.finalize(sink_val=...)` is already arch-agnostic
and folds a sink logit into the softmax denominator (`row_sum += exp(sink - row_max)`);
SM90/SM100 already call it that way. It just was never wired up on SM80.

### Changes (`flash_attn/cute/flash_fwd.py`)

- Removed the `assert learnable_sink is None`.
- Threaded `learnable_sink` from `__call__` through to `kernel()` (it was accepted as a
  parameter but silently dropped before reaching the device kernel).
- At the point where the mainloop finalizes the softmax (previously `row_scale =
  softmax.finalize()`), compute `sink_val` and pass it through, mirroring
  `flash_fwd_sm90.py`'s existing pattern exactly:
  - Non-`pack_gqa`: every row in the CTA shares one Q-head, so a single scalar
    `Float32(learnable_sink[num_head])` broadcasts to all rows.
  - `pack_gqa`: different rows within the same CTA can belong to different Q-heads
    (multiple Q-heads packed per KV-head), so each row's sink value is looked up via an
    identity-tensor partition of `thr_mma_qk` (the same `cS = cute.make_identity_tensor(...)`
    + `reshape_acc_to_mn` pattern already used elsewhere in this file for `score_mod`/mask
    row indexing), giving a per-row `learnable_sink[q_head_idx]`.

`interface.py`'s `_flash_attn_fwd` dispatch already threads `learnable_sink` generically
through `compile_args`/`call_args` for every architecture (SM80 included) — that plumbing
was already correct, so no changes were needed there. The only actual blocker was the
flat assert inside the SM80 kernel itself.

### Tests

Adds `tests/cute/test_sm80_learnable_sink.py`: focused regression coverage at the
gpt-oss-20b production shape (D=64, GQA 8:1, exercising the `pack_gqa` per-row path) and a
plain-MHA D=128 shape (exercising the scalar-broadcast path), both causal and non-causal,
plus a `learnable_sink=None` regression check, comparing against `attention_ref`'s existing
sink-aware reference. The existing `tests/cute/test_flash_attn.py::test_flash_attn_output`
already parametrizes `has_learnable_sink=[False, True]` with no SM80-specific skip, so those
cases exercise this path too — they would previously fail with the assert above on any
Ampere/Ada GPU.

## Test plan

- Confirmed pre-fix: reverting just `flash_fwd.py` reproduces
  `AssertionError: Learnable sink is not supported in this kernel` on real hardware.
- Ran the new test file, `pytest tests/cute/test_sm80_learnable_sink.py`, on 2x RTX 4090
  (sm_89) — 7/7 passed.
- Additionally wrote a standalone script comparing `flash_attn.cute.flash_attn_func(...,
  learnable_sink=...)` against a plain PyTorch SDPA-with-sink reference (softmax with an
  extra sink logit column) across 5 shapes/configs on the same RTX 4090s:
  - gpt-oss-20b shape (D=64, GQA 8:1, causal): cos_sim = 0.99999821
  - same shape, non-causal: cos_sim = 0.99999774
  - GQA 8:1, longer seqlen_k (multi n-block mainloop): cos_sim = 0.99999785
  - MHA D=128, causal: cos_sim = 0.99999797
  - MHA D=128, non-causal: cos_sim = 0.99999785
- Confirmed `learnable_sink=None` still produces identical (regression-free) output.
- `FlashAttentionForwardSm120` inherits this fix structurally (no override of
  `__call__`/`kernel`), which would also close the SM120/RTX-50-series side of #1810, but
  I don't have SM120 hardware and could not verify it directly — flagging this for a
  maintainer/reviewer with access to that hardware. Related: #1895 (gpt-oss learnable-sink
  backward support) is a separate, larger scope (training/backward) not addressed here —
  this PR is forward-only, matching the existing SM90/SM100 forward-only sink support.

### DIFF
--- a/flash_attn/cute/flash_fwd.py
+++ b/flash_attn/cute/flash_fwd.py
@@ -645,8 +645,11 @@ class FlashAttentionForwardSm80(FlashAttentionForwardBase):
 
         mQ/mK/mV/mO has same data types(supports fp16 and bf16) and same layout:
         (batch_size, seqlen_q, num_head, head_dim):(_, _, _, 1)
+
+        learnable_sink: optional per-head fp32 attention-sink logit (shape (num_head,)),
+        folded into the softmax denominator as an extra column, i.e.
+        row_sum += exp(sink[head] - row_max). None reproduces the previous behavior exactly.
         """
-        assert learnable_sink is None, "Learnable sink is not supported in this kernel"
         self._check_type(
             *(t.element_type if t is not None else None for t in (mQ, mK, mV, mO, mLSE, mCuSeqlensQ, mCuSeqlensK, mSeqUsedQ, mSeqUsedK))
         )
@@ -724,6 +727,7 @@ class FlashAttentionForwardSm80(FlashAttentionForwardBase):
             softmax_scale,
             window_size_left,
             window_size_right,
+            learnable_sink,
             self.sQ_layout,
             self.sK_layout,
             self.sV_layout,
@@ -763,6 +767,7 @@ class FlashAttentionForwardSm80(FlashAttentionForwardBase):
         softmax_scale: Optional[Float32],
         window_size_left: Optional[Int32],
         window_size_right: Optional[Int32],
+        learnable_sink: Optional[cute.Tensor],
         sQ_layout: cute.ComposedLayout,
         sK_layout: cute.ComposedLayout,
         sV_layout: cute.ComposedLayout,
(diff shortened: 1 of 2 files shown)


flash_attn/cute/flash_fwd.py
```diff
@@ -1064,7 +1069,21 @@ def preprocess_Q():
...
+                sink_val = Float32(learnable_sink[num_head])
+            else:  # Each thread might have a different sink value due to different q_head
+                sink_val = cute.make_rmem_tensor_like(softmax.row_max, Float32)
+                cS = cute.make_identity_tensor((self.tile_m, self.tile_n))
+                tScS_mn = layout_utils.reshape_acc_to_mn(thr_mma_qk.partition_C(cS))
+                for r in cutlass.range(cute.size(sink_val), unroll_full=True):
+                    row = m_block * self.tile_m + tScS_mn[r][0]
+                    q_head_idx = row % self.qhead_per_kvhead + num_head * self.qhead_per_kvhead
+                    sink_val[r] = Float32(learnable_sink[q_head_idx])
+        row_scale = softmax.finalize(sink_val=sink_val)
```
bvolpato, line 1086: `sink_val` needs to participate in max/rescaling here. On RTX 5070 Ti, a valid call with physical K length 128 and `seqused_k=[0]` returns zero output but all-NaN LSE; expected LSE is sink logit 0.5. Dense attention with finite sink logit 100 returns all-Inf LSE instead of 100. Both come from evaluating `exp2(sink - row_max)` without first folding sink into `row_max`: empty rows use `row_max=-inf`, while a dominant sink overflows. Could you use the stable max/rescaling approach from SM100 and add zero-K plus dominant-sink regression cases?

bvolpato (changes requested):
Tested current merge result on an RTX 5070 Ti (SM120). All seven focused cases pass (`7 passed in 13.14s`), including packed GQA, multi-block GQA, MHA, causal, and non-causal paths. Packed-GQA head mapping also matches the SM90 implementation.

Found one numerical stability issue in sink finalization that affects valid zero-length batches and large finite sink logits. Reproduction details are inline. Once fixed, I can rerun SM120 coverage.

PR #2754, opened 2026-08-03
[CuTe,Fwd,Sm90] add dynamic persistent scheduler to sm90

reubenconducts:
Follow-up to #2559 adding `DynamicPersistentVarlenScheduler` to `flash_fwd_sm90.py`.

### DIFF
--- a/flash_attn/cute/flash_fwd_sm100.py
+++ b/flash_attn/cute/flash_fwd_sm100.py
@@ -466,7 +466,6 @@ class FlashAttentionForwardSm100:
         5. Grid and work scheduling computation
         6. Kernel launch with appropriate parameters
         """
-        assert (tile_count_semaphore is not None) == self.has_tile_count_semaphore
         # setup static attributes before smem/grid/tma computation
         self.q_dtype = mQ.element_type
         self.k_dtype = mK.element_type
--- a/flash_attn/cute/flash_fwd_sm100.py
+++ b/flash_attn/cute/flash_fwd_sm100.py
@@ -466,6 +466,7 @@ class FlashAttentionForwardSm100:
         5. Grid and work scheduling computation
         6. Kernel launch with appropriate parameters
         """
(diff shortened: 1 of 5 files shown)


LiRunGuo:
Some H200 data that may be useful for this PR, from profiling why the SM90 causal forward trails FA3 (main @ edb5c76, cutlass-dsl 4.8.0, bf16, hdim 128, 32k total tokens):

| causal fwd, TFLOPS | FA3 (persistent) | FA4 single-tile | FA4 persistent LPT (static round-robin) | FA4 static persistent (plain order) |
|---|---|---|---|---|
| seqlen 1k | 511 | 414 | 480 | 374 |
| seqlen 4k | 690 | 610 | 590 | 560 |
| seqlen 16k | 695 | 623 | 575 | 520 |

So a persistent scheduler recovers most of the short-sequence gap (1k: 414 -> 480 of FA3's 511), but a *static* assignment loses at long sequence lengths, where the causal tiles are very unequal and the round-robin leaves SMs idle. That matches your choice of a dynamic (atomic tile counter) scheduler here rather than the static one; I'd expect the dynamic version to hold the 1k gain without the 4k/16k regression, but I haven't measured this branch itself.

Two notes from the experiment:

1. I hit the sO/sQ aliasing problem you handle by deferring `consumer_release_w_index(0)` until after the epilogue. With the release left in place, any persistent scheduler gives wrong output (max abs err 5.6 vs SDPA); with the Q release after the O store, results match single-tile bitwise. Just confirming that the deferral is load-bearing, in case it gets simplified during review.

2. The long-sequence part of the FA3 gap turned out to be unrelated to scheduling: the "no masking" mainloop still applied the causal mask on every KV block. That is #2904 (16k causal 613 -> 660 TFLOPS). It touches the same `mma()` loop as this PR but only changes the `mask_fn=` argument of one `mma_one_n_block` call, so the two should rebase onto each other trivially. Happy to re-run the table above on this branch on H200 if that helps.

PR #2755, opened 2026-08-04
[CuTe, SM100] Sparse MLA bwd: don't scatter dK/dV at -1 sentinel indices

abcdabcd987:
In the sparse-MLA (`gather_kv_indices` / DSA top-k) backward on SM100, both scatter-reduce epilogues atomically accumulate at row indices read straight from `gather_kv_indices` with no validity guard, while every load path treats `-1` as invalid and predicates the gather (`topk_gather_kv.py`: `row_valid = topk_idx >= 0 and topk_idx < self.seqlen_k_limit`):

- `flash_bwd_mla_sm100.py` (dV epilogue): `gmem_n_idx = rIdxTopK[j]` used directly in `elem_pointer(mdV_cur, (gmem_n_idx, dv_offset))` + `atomic_add_fp32x4`.
- `flash_bwd_mla_dk_sm100.py` (dK epilogue): `seqlen_k_idx_in_batch = sI_tile[topk_idx]` used directly in `elem_pointer` + `cute.arch.atomic_add`.

`-1` is the documented sentinel for invalid top-k slots (the forward handles it via the kv bitmask — `interface.py`: "always use kv bitmask by default (handles -1 sentinel)"), and any causal top-k index tensor necessarily contains `-1` padding for early tokens, so any causal DSA workload hits this on every backward call.

### Why the damage is hard to attribute

The masked math is correct — `p` and `ds` are exactly 0.0 at sentinel slots — so the scattered *value* is 0.0. But:

1. Index `-1` computes an address one row before the (batch-sliced) buffer base: for batch 0 that is out of bounds into whatever tensor the allocator placed there; for later batches it is the previous batch's last dk/dv row.
2. PTX `red.add.f32` flushes subnormal destinations to (sign-preserving) zero and canonicalizes NaN payloads **even when the addend is 0.0**. If the adjacent tensor holds int32 data reinterpreted as fp32 — e.g. another top-k index tensor, whose small positive values are all subnormal bit patterns — those values are silently zeroed, and `-1` (0xFFFFFFFF, a NaN bit pattern) is canonicalized to 0x7FC00000, a huge positive int.
3. Whether anything visibly breaks depends purely on CUDA caching-allocator block adjacency, so the symptom is deterministic per layout but appears/disappears with unrelated allocation changes — it mimics a stale-cache or race bug and survives clearing every DSL compile cache. Observed symptoms range from bitwise-correct results, to silently wrong grads, to IMA inside `_sparse_mla_dk` (when the corrupted victim is the index tensor of the same call).

### Demonstration

Plant an int32 canary (values 1..N — all subnormal fp32 bit patterns) immediately before a dv/dk buffer and run one sparse fwd+bwd with causal `-1`-padded indices on GB200:

- unpatched: every canary word covered by row `-1` is flushed to 0x0 while `p`/`ds` at all sentinel slots are bitwise 0.0 and all grads still match an fp32 reference (the OOB landed in the canary, i.e. in dead memory);
- in a second layout, the internally allocated `dk` landed right after an unrelated int32 tensor, and the dK epilogue zeroed that tensor's last 256 bytes — the allocator-adjacency lottery described above;
- patched: canaries intact, valid-slot grads unchanged: max abs err vs the fp32 reference matches the unpatched run's, with residual differences <=2e-5 that also appear between reruns of the same binary (atomic fp32 reduction-order noise).

### Fix

Skip the atomic when the index is negative, mirroring the load-side guard. The skipped contribution is mathematically 0.0, so numerics for valid slots are unchanged (verified: max abs err vs an fp32 reference is the same before/after, up to <=2e-5 run-to-run atomic reduction-order noise).

Also fixes `_flash_attn_bwd_sparse_mla` discarding caller-supplied `dq=`/`dk=` buffers (`dq = dk = None` immediately after recording `prealloc_dq/dk`, then the re-allocation is skipped because `prealloc_*` is True → crash). The new test uses preallocated `dk`/`dv` to plant canaries.

### Tests

`tests/cute/test_flash_attn.py::test_flash_attn_mla_sparse_bwd_sentinel` (parametrized over causal × shared_kv × seqlen 512/1024) and a varlen counterpart `..._sentinel_varlen` (causal × shared_kv, packed docs [512, 4, 1024] including one shorter than topk_len, whose index rows are almost entirely sentinels):

1. builds causal top-k indices with `-1` tail padding (what any causal selector produces for early queries),
2. checks out/lse/grads against `attention_ref` through the public autograd path,
3. reruns the backward with preallocated `dk`/`dv` surrounded by int32 canaries and asserts the canaries are untouched.

The varlen kernels are separate compile-time specializations, and the dK epilogue guard must apply to the doc-relative index *before* `seqlen_k_offset` is added — the varlen test pins that down (doc 0's row `-1` is the canary-visible case; later docs' row `-1` would land in the previous doc's last row).

```bash
# fails on main, passes on this branch
pytest tests/cute/test_flash_attn.py -k "mla_sparse_bwd_sentinel"
# unchanged
pytest tests/cute/test_flash_attn.py -k "mla_absorbed"
```

Fails deterministically on `main` (canary check; grads can also go garbage / IMA depending on allocation layout), passes with the fix. Also made `attention_ref`'s top-k mask sentinel-aware (its `scatter_` routed `-1` into key 0 and tripped scatter's bounds check; out-of-range indices now also map to a dummy column).

Existing `test_flash_attn_mla_absorbed` subset (kv_sparsity, both causal, both shared_kv, seqlen 256/1024) passes unchanged with the guard.

### Notes / not addressed here

- Indices `>= seqlen_k` are treated as invalid by the load-side guard but are not guarded in the scatter (only `-1` is the documented sentinel). If such indices are considered legal input, the scatter guard should become `0 <= idx < seqlen_k`.
- Two related issues observed while debugging, left for separate PRs: non-contiguous (`.split()`-view) k/v are accepted by the sparse path and produce wrong results (`maybe_contiguous` only fixes `stride(-1) != 1`); and for invalid rows the gathered `sK`/`sV` smem tiles are left unwritten by the predicated `cp.async` yet still consumed by the MMA — correctness relies on `dS == 0`, so a leftover Inf/NaN in smem would produce NaN (zero-fill or predicating the MMA would make this robust).

Environment: GB200 (SM100), torch 2.13.0+cu130, nvidia-cutlass-dsl 4.6.0.dev0.

### DIFF
--- a/flash_attn/cute/flash_bwd_mla_dk_sm100.py
+++ b/flash_attn/cute/flash_bwd_mla_dk_sm100.py
@@ -1117,13 +1117,15 @@ class dKGemmKernel:
                 else:
                     seqlen_k_idx = (batch, seqlen_k_idx_in_batch)
                 cute.copy(tiled_copy_c, tCsC[(None, None, topk_idx, c_buffer)], tCrC)
-                for j in cutlass.range_constexpr(cute.size(tCrC, mode=[1])):
-                    for i in cutlass.range_constexpr(cute.size(tCrC, mode=[0])):
-                        ptr = elem_pointer(tCgC, (i, j, subtile_idx, seqlen_k_idx))
-                        cute.arch.atomic_add(
-                            ptr=ptr,
-                            val=tCrC[i, j],
-                        )
+                # Skip -1 sentinel slots (invalid top-k entries)
+                if seqlen_k_idx_in_batch >= 0:
+                    for j in cutlass.range_constexpr(cute.size(tCrC, mode=[1])):
+                        for i in cutlass.range_constexpr(cute.size(tCrC, mode=[0])):
+                            ptr = elem_pointer(tCgC, (i, j, subtile_idx, seqlen_k_idx))
+                            cute.arch.atomic_add(
+                                ptr=ptr,
+                                val=tCrC[i, j],
+                            )
             epilog_sync_barrier.arrive_and_wait()
 
         epilog_sync_barrier.arrive_and_wait()
--- a/flash_attn/cute/flash_bwd_mla_sm100.py
+++ b/flash_attn/cute/flash_bwd_mla_sm100.py
@@ -2103,21 +2103,23 @@ class FlashAttentionSparseMLABackwardSm100:
 
                         for j in cutlass.range_constexpr(gmem_rows_per_thread):
                             gmem_n_idx = rIdxTopK[j]
-                            for w in cutlass.range_constexpr(2):
-                                dv_offset = (
-                                    self.hdimv // self.num_hdimv_splits * split  # 256 * split
-                                    + (self.hdimv // self.num_hdimv_splits // 2) * w  # 128 * w
-                                    + 32 * i
-                                )
-                                dv_offset += tdVcdV[0, j, 0][1]
-                                gmem_coord = (gmem_n_idx, dv_offset)
-                                dV_gmem_ptr = elem_pointer(mdV_cur, gmem_coord)
+                            # Skip -1 sentinel slots (invalid top-k entries)
+                            if gmem_n_idx >= 0:
+                                for w in cutlass.range_constexpr(2):
+                                    dv_offset = (
+                                        self.hdimv // self.num_hdimv_splits * split  # 256 * split
+                                        + (self.hdimv // self.num_hdimv_splits // 2) * w  # 128 * w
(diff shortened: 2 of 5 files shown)


jayhshah:
Thanks!

jayhshah (approved):

PR #2756, opened 2026-08-05
[CuTe,Bwd,Sm90] Fix: wait for bwd_preprocess on the block-sparse path, matching the dense path

Fugoes:
I recently hit the same bug as #2643, and I think this is a better fix. It affects anyone running backward with PyTorch's `flex_attention` flash-attention backend (`backend="FLASH"`) and a `mask_mod`. The kernel reads preprocess's outputs before they have been written, so gradients can come out garbage, sometimes NaN (rarely, 0.1% to 1% of backward launches in my runs).

Credit to @codecaution for the diagnosis in #2643. This differs only in where the wait is: I put it in `_load_q_do_block_sm90` after `load_Q` and before `load_LSE`, matching the dense path's order (`load_K; load_Q; wait; load_LSE`). #2643 waits ahead of all loads, which gives up the first block's K/Q TMA overlap: arounds +1.2% in my local measurements on long-sequence shapes with a sparse mask, about a third of what PDL buys for this shape. Happy to close this and fold the placement change into #2643 instead if that's preferred.

CC @drisspg (block-sparse backward / flex_attention `backend="FLASH"`).

### DIFF
--- a/flash_attn/cute/block_sparse_utils.py
+++ b/flash_attn/cute/block_sparse_utils.py
@@ -1636,6 +1636,13 @@ def _load_q_do_block_sm90(
     else:
         pipeline_Q.producer_acquire(producer_state_Q)
     load_Q(m_block, producer_state=producer_state_Q)
+    if load_kv:
+        # Wait for bwd preprocess to finish writing LSE and dPsum, and zeroing dQaccum. Same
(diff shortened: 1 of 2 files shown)


drisspg (approved):
looks good thanks for fixing

PR #2758, opened 2026-08-05
[CuTe, SM120] Support automatic num_splits

bvolpato:
## Summary

SM120 does not support SplitKV, so `num_splits=1` is its only executable configuration. Automatic selection (`num_splits <= 0`) currently asserts instead of choosing that configuration. This surfaced because `benchmarks/benchmark_attn.py` uses `num_splits=0` by default.

- resolve automatic values to `1` on SM120
- keep rejecting SplitKV requests greater than `1`
- cover default, automatic, and unsupported values in the existing SM120 test

## Testing

```bash
PYTHONPATH=$PWD FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1 \
  pytest -q tests/cute/test_flash_attn.py -k sm120_num_splits
```

Result on RTX 5070 Ti (SM120), rebased onto `main` at `145b101`: `1 passed, 415817 deselected`.

```bash
PYTHONPATH=$PWD FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1 \
  python benchmarks/benchmark_attn.py --backend fa2,fa4 --fwd \
  --headdim 128 --seqlen 1k --causal false --warmup 5 --rep 20
```

Stock benchmark completes without the `--num-splits 1` workaround.

### DIFF
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -367,7 +367,8 @@ def _get_fwd_config(
         132 if is_fake_mode() else torch.cuda.get_device_properties(device).multi_processor_count
     )
     if arch // 10 == 12:
-        assert num_splits == 1, "SM120 forward only supports num_splits=1"
+        assert num_splits <= 1, "SM120 forward only supports num_splits=1"
+        num_splits = 1
     elif num_splits < 1:
         num_splits = num_splits_heuristic(total_mblocks, num_SMs, num_n_blocks, 128)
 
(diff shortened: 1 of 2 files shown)


bvolpato:
@drisspg could you take a look at this SM120 fix when you have a chance? It maps automatic `num_splits <= 0` to the only supported SM120 configuration (`1`) while preserving rejection of SplitKV values greater than `1`. Focused test and stock benchmark passed on RTX 5070 Ti. PR is mergeable and ready for review. Thanks!

Issue #2760, opened 2026-08-06
[CuTe, SM100] Deadlock in varlen + block-sparse + SplitKV forward: packed `split_idx` is not unpacked on the block-sparse paths (regression from #2559)


JiaxuanBai:
## Summary

On SM100, the CuTe forward kernel deadlocks (GPU spins at 100%, kernel never returns) whenever **varlen + block sparsity + SplitKV** are combined. This is a regression introduced by #2559 (`c46b8144`, "Varlen Dynamic Persistent scheduler and metadata"): the scheduler now packs the per-batch dynamic `num_splits` into the top 16 bits of `split_idx`, but the SM100 block-sparse code paths don't consistently unpack it, so different warps in the same CTA disagree on the tile's KV block range and the softmax/correction/MMA mbarrier handshake never completes.

Note: #2705 (first-tile flag preservation) does not fix this — it addresses a different scheduler issue and the hang still reproduces with it applied.

## Reproduction

```bash
pytest -x "tests/cute/test_mask_mod_varlen.py::test_varlen_block_sparse_coarse_kv_requires_offsets[packed-3-14]"
```

The test hangs in its first (valid) forward call. All parametrizations of `test_varlen_block_sparse_splitkv_matches_unsplit` hang the same way. With `pytest -n`, this shows up as workers crashing with "node down: Not properly terminated" once the per-test timeout kills them.

- GPU: B200 (SM100)
- Bisect: `c75d019d` passes → `c46b8144` (#2559) hangs; still hangs at current main.
- `test_varlen_block_sparse_coarse_kv_requires_offsets[seqused-1-15]` (num_splits=1) does **not** hang, consistent with the analysis below (no split packing when there is effectively no split).

## Root cause

Since #2559, when scheduler metadata provides `num_splits_dynamic_ptr`, the varlen schedulers pack the per-batch dynamic split count into the work tile's `split_idx`:

- `tile_scheduler.py`, `SingleTileVarlenScheduler._decode_work_tile` / `DynamicPersistentVarlenScheduler.get_current_work`:
  `split_idx = split_idx | (num_splits << 16)`

The dense path handles this correctly — `BlockInfo.get_n_block_min_max` unpacks internally when `pack_split_idx` is set:

```python
if const_expr(self.pack_split_idx):
    num_splits = split_idx >> 16
    split_idx = split_idx & 0xFFFF
```

[...]

PR #2761, opened 2026-08-06
[CuTe, SM100] Fix deadlock in varlen + block-sparse + SplitKV forward

JiaxuanBai:
https://github.com/Dao-AILab/flash-attention/issues/2760

When scheduler metadata provides per-batch dynamic num_splits, the varlen tile schedulers pack num_splits into the top 16 bits of split_idx. The dense path unpacks it inside BlockInfo.get_n_block_min_max, but the SM100 block-sparse paths pass split_idx to the block-sparse helpers as is: the load and MMA warps passed the packed value, for which split_block_range yields an empty block range, while the softmax and correction warps passed the unpacked value, yielding a non-empty range. The warps then disagree on whether a tile has work, the softmax/correction/MMA mbarrier handshake never completes, and the kernel spins forever. Every varlen + block-sparse
+ SplitKV forward hangs this way.

Also pass the dynamic per-batch num_splits (instead of the static maximum) to the block-sparse helpers so the split ranges cover the whole block list, matching the dense path.

### DIFF
--- a/flash_attn/cute/flash_fwd_sm100.py
+++ b/flash_attn/cute/flash_fwd_sm100.py
@@ -1650,6 +1650,12 @@ class FlashAttentionForwardSm100:
                             kv_producer_state.advance()
 
             else:
+                # Match the dense path (get_n_block_min_max): the scheduler packs the
+                # per-batch dynamic num_splits into the top 16 bits of split_idx.
+                num_splits_dyn = num_splits
+                if const_expr(self.is_split_kv and block_info.pack_split_idx):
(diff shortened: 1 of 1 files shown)


JiaxuanBai:
Hi @reubenconducts @jayhshah . Could you please review this PR? Thanks!

jayhshah:
LGTM, we'll merge this in and disable dynamic num splits for block sparse as a follow on since scheduler metadata isn't currently block sparse aware in any case.

jayhshah (approved):

Issue #2765, opened 2026-08-07
[CuTe, SM100] test_clc_fuzz fails after #2559 due to stale scheduler expectations


JiaxuanBai:
## Summary

After #2559, `tests/cute/test_clc_fuzz.py` fails on SM100 because several scheduler-class assertions still reflect the scheduler dispatch behavior from before #2559.

At commit `c68c592fd9da1e40a4fb0b56229caae6754ac5c9`, the test result is:

```text
22 failed, 167 passed
```

## Representative failures:
```text
AssertionError: Expected SingleTileVarlenScheduler, got StaticPersistentTileScheduler

AssertionError: Expected SingleTileVarlenScheduler, got DynamicPersistentVarlenScheduler

AssertionError: Expected SingleTileScheduler fallback, got StaticPersistentTileScheduler
```
Current main (1cc7ff67cbc5685046c75183e8defecca3e35d5c) still contains the same assertions.
## Reproduction
On an SM100 GPU:
```
pytest -n=16 --tb=short tests/cute/test_clc_fuzz.py
```
Environment used for the observed failure:
GPU: GB200/B200, SM100
OS: Ubuntu 24.04, arm64
Python: 3.12.8
CUDA: 13.3
flash-attention: c68c592fd9da1e40a4fb0b56229caae6754ac5c9
nvidia-cutlass-dsl: 4.7

## Root cause
PR #2559 intentionally introduced new varlen scheduler dispatch behavior.
FlashAttentionForwardSm100 can now select:

1. `DynamicPersistentVarlenScheduler` when dynamic-persistent scheduler metadata is used;
2. `StaticPersistentTileScheduler` when the varlen workload is eligible for static persistence;
3. `SingleTileVarlenScheduler` otherwise.

However, the varlen helpers in `test_clc_fuzz.py` still unconditionally require SingleTileVarlenScheduler:
[check_varlen_output](https://github.com/Dao-AILab/flash-attention/blob/1cc7ff67cbc5685046c75183e8defecca3e35d5c/tests/cute/test_clc_fuzz.py#L305-L330)
[check_varlen_output_seqused](https://github.com/Dao-AILab/flash-attention/blob/1cc7ff67cbc5685046c75183e8defecca3e35d5c/tests/cute/test_clc_fuzz.py#L348-L377)

PR #2766, opened 2026-08-07
Fix CLC fuzz scheduler expectations

JiaxuanBai:
https://github.com/Dao-AILab/flash-attention/issues/2765

PR #2559 added static- and dynamic-persistent dispatch for varlen SM100 kernels, but test_clc_fuzz still required the pre-change single-tile scheduler, so the scheduler assertions failed before numerical validation could run.

Validate each scheduler class against its scheduling mode, account for dynamic SplitKV dispatch, and update the two static fallback expectations. STATIC mode accepts both StaticPersistentTileScheduler and SingleTileVarlenScheduler for varlen because the dispatch depends on whether every batch fits in a single m-block.

Verification: the full tests/cute/test_clc_fuzz.py suite passes on SM100 (B200, CC 10.0) — 189 passed — both on this branch's base (1cc7ff6) and cherry-picked onto c68c592. The GQA + SplitKV varlen cases were confirmed to still select the CLC SingleTileVarlenScheduler, matching the assertion precedence.

### DIFF
--- a/tests/cute/test_clc_fuzz.py
+++ b/tests/cute/test_clc_fuzz.py
@@ -18,6 +18,7 @@ from flash_attn.cute.flash_fwd_sm100 import FlashAttentionForwardSm100
 from flash_attn.cute.interface import flash_attn_func, flash_attn_varlen_func
 from flash_attn.cute.testing import attention_ref
 from flash_attn.cute.tile_scheduler import (
+    DynamicPersistentVarlenScheduler,
     SchedulingMode,
     SingleTileLPTScheduler,
     SingleTileVarlenScheduler,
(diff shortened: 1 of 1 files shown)


JiaxuanBai:
Hi @jayhshah @reubenconducts . Could you please help review this PR? Thanks!

reubenconducts:
@drisspg these edits are reasonable, but do we even still need `test_clc_fuzz` at all?

drisspg:
hmm so this was more just `test_schedulers` basically but we are more confident in clc correctness B the coverage is probably spread throughout other files; maybe we land to unblcok and then we can think if we want to just delete

drisspg (approved):

PR #2770, opened 2026-08-09
[CuTe] Pass Fwd/Bwd args as standardized NamedTuples

reubenconducts:
**Do not land until cute dsl is bumped to 4.7.1**

This PR addresses a pain point from the growing number of specialized kernel classes for the main forward and backward kernels, which do not all take the same positional arguments. We propose the following re-organization:
- Each general kernel group (for this PR, just Forward and Backward) has a single catch-all `*KernelArgs` NamedTuple.
- Each individual kernel class has its own local `Args` struct
- At compile time, we check to make sure that the `*KernelArgs` struct passed to a given `kernel_cls`
	- is a strict superset of the local `kernel_cls.Args` and
	- for every entry in `*KernelArgs`, either it exists in `kernel_cls.Args` or is `None`
- Args are unpacked by name, so there is no risk of error stemming from incorrect position.

This ensures that each kernel class is responsible for raising errors regarding incorrect arguments so that, for example, we do not need [these assertions](https://github.com/Dao-AILab/flash-attention/blob/a369df707e1980fb328abcc1733e3457ec10155f/flash_attn/cute/interface.py#L1307) (of course, we can instead choose to raise those assertions from within the kernel class itself, if the point is to keep a placeholder for when such an argument is eventually supported). 

This PR is intended to compose cleanly with #2732 and subsequent follow-ups.

### DIFF
--- a/flash_attn/cute/cute_dsl_utils.py
+++ b/flash_attn/cute/cute_dsl_utils.py
@@ -24,6 +24,10 @@ load_cubin_module_data_og = cutlass.base_dsl.runtime.cuda.load_cubin_module_data
 cute_compile_og = cute.compile
 
 
+# Monkey-patch: as of 2026-08-10, TVM-FFI cannot detect env stream if all
+# tensor arguments to a kernel are contained in a tuple/NamedTuple. This workaround
+# patches TVMFFIFunctionBuilder.find_env_stream to search through nested args
+# when finding the current stream.
 _PATCH_MARKER = "_flash_attn_nested_env_stream_detection"
 
 
--- a/flash_attn/cute/cute_dsl_utils.py
+++ b/flash_attn/cute/cute_dsl_utils.py
@@ -27,20 +27,24 @@ cute_compile_og = cute.compile
 def _enable_nested_env_stream_detection():
     """Let TVM FFI find the implicit stream inside a tuple parameter.
 
-    Kernels take their arguments as a single Args namedtuple, so no tensor appears at the top
+    Some kernels take their arguments as a single Args namedtuple, so no tensor appears at the top
     level of the signature and the stock lookup, which only scans top-level parameters, finds
     nothing. Retry against a flattened parameter list; signatures that do have a top-level
     tensor never reach the fallback.
     """
-    from cutlass.base_dsl.tvm_ffi_builder import spec
-    from cutlass.base_dsl.tvm_ffi_builder.tvm_ffi_builder import TVMFFIFunctionBuilder
+    try:
+        from cutlass.base_dsl.tvm_ffi_builder import spec
+        from cutlass.base_dsl.tvm_ffi_builder.tvm_ffi_builder import TVMFFIFunctionBuilder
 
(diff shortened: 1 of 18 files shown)


flash_attn/cute/cute_dsl_utils.py
```diff
@@ -24,6 +24,38 @@
 cute_compile_og = cute.compile
 
 
+_PATCH_MARKER = "_flash_attn_nested_env_stream_detection"
+
+
+def _enable_nested_env_stream_detection():
```
drisspg: can we document why we need this? also we should open an issue :)

flash_attn/cute/cute_dsl_utils.py
```diff
@@ -33,6 +65,19 @@
 }
 
 
+# Building a cutlass Int32/Float32 from a python value runs a numpy cast that costs ~1us, which is
+# material next to a kernel launch. These wrap the Python value and nothing else, so one instance
+# per value is shared.
+@lru_cache
+def cutlass_int32(value: int) -> cutlass.Int32:
```
drisspg: another good issue; this is uglyyy

flash_attn/cute/flash_bwd.py
```diff
@@ -25,10 +25,35 @@
 from quack.cute_dsl_utils import ParamsBase
 from flash_attn.cute.tile_scheduler import SingleTileScheduler, SingleTileVarlenScheduler, TileSchedulerArguments
 from flash_attn.cute.block_sparsity import BlockSparseTensors
+from flash_attn.cute.kernel_args import normalize_kernel_args
 from flash_attn.cute.utils import AuxData
 
 
 class FlashAttentionBackwardSm80:
+    class Args(NamedTuple):
```
drisspg, line 33: Args or Params ? the latter is my cutlassy right?

flash_attn/cute/flash_fwd_sm90.py
```diff
@@ -157,24 +178,7 @@ class SharedStorageSharedQV:
...
-        mSeqUsedK: Optional[cute.Tensor] = None,
-        mPageTable: Optional[cute.Tensor] = None,  # (b_k, max_num_pages_per_seq)
-        window_size_left: Int32 | int | None = None,
-        window_size_right: Int32 | int | None = None,
-        learnable_sink: Optional[cute.Tensor] = None,
-        blocksparse_tensors: Optional[BlockSparseTensors] = None,
-        aux_data: AuxData = AuxData(),
-        mCuTotalMBlocks: Optional[cute.Tensor] = None,
-        mCuTotalSplitsMBlocks: Optional[cute.Tensor] = None,
+        args,
```
drisspg: should the arg have type: FwdArgs which is typed as a union of all the differnt sub arg types?

reubenconducts: Yes, it should have type `FwdKernelArgs`. Will add.

flash_attn/cute/flash_bwd.py
reubenconducts: Open to suggestions on this as long as it's consistent and meshes with your config work

PR #2773, opened 2026-08-09
[CuTe] Include callable defaults in JIT cache hashes

guoriyue:
## Summary

CuTe injects Python positional and keyword-only defaults into generated code, but `hash_callable()` only included source/bytecode and closure cells. Factory-created score modifiers with the same source and different defaults therefore shared one compile-cache key and could silently execute the wrong specialization.

This focused change adds `__defaults__` and `__kwdefaults__` to the existing callable hash policy. It intentionally reuses the established `repr` treatment used for closure values instead of introducing a general serializer or a second object-identity system.

## RTX 5090 reproduction

Two `@cute.jit` score-mod factories use identical source but defaults `scale=1.0` and `scale=2.0`.

On unmodified main:

- both callables have the same hash;
- the forward cache contains one compiled kernel after both calls;
- the second call reuses the scale-1 specialization;
- 86.7% of output elements differ from its own reference, with maximum absolute error `1.40625`.

With this patch, the two callables produce distinct cache entries and both outputs match their corresponding PyTorch references.

## Validation

- unmodified main: the two focused default-hash unit tests fail;
- patch: `tests/cute/test_utils.py` reports 13 passed;
- unmodified main: the end-to-end score-mod cache test fails numerically;
- patch: the same RTX 5090 test passes and observes two cache entries;
- `py_compile` and `git diff --check` pass.

## Scope and design

The production delta is limited to hashing the two standard Python default-value attributes. Existing fast paths for explicit `__cute_hash__`, closure handling, and `__vec_size__` remain unchanged.

Mutable defaults after hash memoization, global-variable hashing, a bytecode fallback redesign, tensor captures, and persistent-cache protocol changes are intentionally out of scope. No new helper module, serialization taxonomy, or module-level constant table is introduced.

### DIFF
--- a/flash_attn/cute/utils.py
+++ b/flash_attn/cute/utils.py
@@ -100,7 +100,7 @@ def _get_disable_2cta_default(is_fwd: bool = False) -> bool:
 
 
 def _compute_base_hash(func: Callable) -> str:
-    """Compute hash from source code or bytecode and closure values."""
+    """Compute a hash from callable code and captured compile-time values."""
     try:
         data = inspect.getsource(func).encode()
     except (OSError, TypeError):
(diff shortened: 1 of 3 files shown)


drisspg (approved):
Looks good; thanks for the pr

Issue #2566, opened 2026-05-14
FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1 causes cudaErrorInvalidDeviceFunction on cross-process load


eitanporat:
**Title:** `FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1` causes `cudaErrorInvalidDeviceFunction` on cross-process load

## Summary

Enabling the cute DSL persistent JIT cache (`FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1` + `FLASH_ATTENTION_CUTE_DSL_CACHE_DIR=<path>`) causes the next process to crash on its first attention call with:

```
RuntimeError: CUDA Error: cudaErrorInvalidDeviceFunction
  in _flash_attn_fwd.compile_cache[compile_key](...)
```

The cache key in `flash_attn/cute/cache_utils.py` doesn't fully disambiguate compiled artifacts. A `.cubin` cached by one process is loaded into a different process's `compile_cache` dict for a slightly different call site, then handed to CUDA's launch, which rejects it.

Within a single process the in-memory cache works correctly — same process's compiled kernels match their use sites. The bug is specifically in cross-process reuse of the disk-cached artifacts.

## Environment

- flash-attention: `Dao-AILab/flash-attention@<recent HEAD>` (also reproducible with `RandNMR73/flash-attention@fa4-compile`, fork SHA `71677f5f`)
- torch: 2.11.0+cu128
- CUDA: 12.9.1
- Python: 3.12.13
- GPU: NVIDIA GB200 (sm_100)
- nvidia-cutlass-dsl: bundled with torch wheel
- Workload: LTX-2 video diffusion (FastVideo), bidirectional self-attention via `flash_attn.cute`

## Reproduce

1. Set `FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1` and `FLASH_ATTENTION_CUTE_DSL_CACHE_DIR=/persistent/path` in the container env.
2. Start a Python process that exercises `flash_attn.cute` over multiple attention shapes (we hit it on first call after a warmup pass with stage-1 + stage-2 latent shapes).
3. Let the process terminate gracefully so the cache flushes to disk (`/persistent/path` populated with cubin files).
4. Start a fresh process with the same env vars.
5. On the first `flash_attn.cute` call after `compile_cache` has loaded from disk, observe `cudaErrorInvalidDeviceFunction`.

## Logs

```
[...]

PR #2774, opened 2026-08-09
[CuTe] Prevent partial persistent JIT cache artifacts

guoriyue:
## Summary

The persistent CuTe cache exported directly to its final `<hash>.o` path. If `export_to_c()` wrote any bytes and then failed, that partial file looked like a cache hit forever: retries skipped export and fresh processes attempted to load an invalid object.

This change:

- exports to a unique same-directory temporary `.o` created with `O_EXCL`;
- atomically publishes it with `Path.replace()` only after export succeeds;
- removes temporary artifacts on every ordinary failure path while preserving the original export exception;
- retries load failures after releasing the shared lock and acquiring the per-key exclusive lock;
- removes an object only if it still fails exclusive revalidation, allowing normal cache-miss recompilation;
- preserves the prior `0666 & umask` artifact permissions instead of narrowing shared custom caches to `0600`.

## Reproduction on current main

A deterministic exporter wrote `partial-object` and raised. On unpatched `main`:

- the final object still existed;
- the failed exporter was called once;
- a valid retry exporter was called zero times because the final path existed;
- a fresh cache failed with `BinaryExecutionEngine: Failed to add object file: The file was not recognized as a valid object file`.

With this patch, failed export leaves neither a final object nor a temporary artifact, and a retry publishes the complete object once. An actual CuTe load of arbitrary partial bytes is revalidated, removed, and returned as a cache miss.

## Validation

- `PYTHONPATH=. python -m pytest -q tests/cute/test_cache_utils.py` (`4 passed`)
- real RTX 5090 compile -> atomic export -> in-memory clear -> disk reload; one final object, zero temp objects, bitwise-identical output
- real invalid-object load recovery with CuTe runtime
- five-process failed-writer/contention probe; no deadlock or partial publication
- artifact mode comparison: unpatched `0664`, patched `0664`
- `ruff check`, `ruff format --check`, `python -m py_compile`, and `git diff --check`

## Scope

Cache-key/fingerprint changes, launch-time incompatibility such as #2566, cold-compile serialization, `clear()` concurrency redesign, stale temp cleanup after SIGKILL, and fsync/power-loss durability are intentionally out of scope.

### DIFF
--- a/flash_attn/cute/cache_utils.py
+++ b/flash_attn/cute/cache_utils.py
@@ -3,6 +3,7 @@ import fcntl
 import hashlib
 import os
 import pickle
+import secrets
 import sys
 import tempfile
 import time
@@ -203,8 +204,9 @@ class JITPersistentCache(JITCache):
     def _try_load_from_storage(self, key: CompileKeyType) -> bool:
         """
         Try to load a function from persistent storage into in-memory cache.
-        Returns True if loaded successfully, False if not found on disk.
-        Holds a shared lock during loading to prevent concurrent writes.
+        Returns True if loaded successfully, False if missing or invalid on disk.
+        Loads under a shared lock, then revalidates failures under an exclusive
+        lock before removing an invalid artifact.
         """
         sha256_hex = self._key_to_hash(key)
         obj_path = self.cache_path / f"{sha256_hex}.o"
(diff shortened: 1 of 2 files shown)


guoriyue:
Closing this draft after a stricter contribution audit. The cache-write/recovery hardening is technically valid, but it is a reliability-only change with no new model/kernel capability or measured steady-state performance benefit, so it does not belong in the current ten capability/performance contributions. The branch and commit remain available if the project wants to revisit it under a cache-reliability scope.

PR #2775, opened 2026-08-09
[CuTe, Fwd] Fix PackGQA predication for padded head dimensions

guoriyue:
## Summary

Fix the CuTe PackGQA load/store predicates for head dimensions that are 8-byte-vector aligned but not multiples of 16 elements, such as 72, 88, 104, 120, and 136.

PackGQA slices Q loads and O stores into one K group. The sliced copy operand retains its vector layout, while the old predicate contains only one Boolean transaction value. CuTe IR verification therefore rejects the default GQA path.

Representative main failures:

- `D=72, Dv=64`: expected predicate shape `((8),(1))`, got `(1,(1))`;
- `D=64, Dv=72`: expected predicate shape `((8,1),(1))`, got `(1,(1))`.

The patch materializes a Boolean fragment matching the sliced operand and fills it with the existing transaction predicate. The branch is compile-time eliminated when no head-dimension predicate is needed.

## Correctness coverage

RTX 5090 / SM120:

- unmodified main: the first three representative dense cases fail during IR verification;
- patch: 25 focused real-GPU cases pass;
- patch: 24 dense FakeTensor compile cases pass;
- BF16 and FP16;
- causal and non-causal;
- `D/Dv` combinations `(72,64)`, `(64,72)`, `(88,88)`, `(104,104)`, `(120,120)`, and `(136,136)`;
- varlen coverage and FP32-reference numerical checks;
- tests exercise the public default dispatch, where GQA automatically selects PackGQA.

## RTX 5090 performance

BF16/FP16, batch 1, 8 Q heads / 1 KV head, D=72. Each cell uses seven alternating unpacked/PackGQA timing blocks with 50 launches per block. Values are `pack_gqa=False latency / PackGQA latency`; greater than 1 means PackGQA is faster.

| dtype | sequence | non-causal | causal |
|---|---:|---:|---:|
| BF16 | 512 | 1.052x | 0.997x |
| BF16 | 2048 | 1.003x | 0.967x |
| BF16 | 8192 | 1.002x | 1.125x |
| FP16 | 512 | 1.064x | 0.998x |
| FP16 | 2048 | 1.004x | 0.967x |
| FP16 | 8192 | 1.001x | 1.168x |

The 12-cell geometric mean is 1.027x. The optimization is shape-dependent: causal S2048 is about 3.3% slower, while causal S8192 is 12.5%-16.8% faster. This PR's primary capability is making the already-selected default PackGQA path compile and run for these dimensions; it does not claim a universal speedup.

## Related work and scope

#2634 introduced related full-fragment predicates in partial-row PackGQA branches. Its current tile-aligned generic path still retains the compressed predicate and reproduces this padded-dimension failure, so the focused fix remains independently useful.

The global predicate helper, TMA paths, pointer/layout mapping, and backward kernels are intentionally unchanged.

### DIFF
--- a/flash_attn/cute/pack_gqa.py
+++ b/flash_attn/cute/pack_gqa.py
@@ -176,11 +176,17 @@ class PackGQA:
                 mQ_cur_copy = cute.tiled_divide(mQ_cur, (elems_per_load,))
                 for k in cutlass.range_constexpr(cute.size(tQsQ.shape[2])):
                     ki = tQcQ[0, 0, k][1] // elems_per_load
+                    pred = None
+                    if cutlass.const_expr(self.check_hdim_oob):
+                        # Slicing one K group drops tiled-copy predicate broadcasting,
+                        # so expand its transaction predicate to the copy operand.
+                        pred = cute.make_fragment_like(tQsQ[None, m, k], cutlass.Boolean)
+                        pred.fill(tQpQ[0, m, k])
                     cute.copy(
                         gmem_thr_copy,
                         mQ_cur_copy[None, ki],
                         tQsQ[None, m, k],
-                        pred=tQpQ[None, m, k] if cutlass.const_expr(self.check_hdim_oob) else None,
+                        pred=pred,
                     )
             # We don't need to clear the sQ smem tiles since we'll only write out the valid outputs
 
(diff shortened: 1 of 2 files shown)


drisspg (approved):
LGTM; thanks!

PR #2779, opened 2026-08-09
[Generation] Fix speculative sampling axis indexing

guoriyue:
## Summary

- scatter top-p filtering masks along the vocabulary dimension so speculative logits with multiple leading dimensions work
- use paired batch/position indexing when writing resampled tokens so rows with different rejection positions cannot overwrite each other
- add deterministic CPU regressions for 3D top-p filtering and batched speculative sampling

## Problem

`sample_speculative` passes logits shaped `[batch, sequence, vocabulary]` into the top-p helper, but the helper scattered along dimension 1. Vocabulary indices can therefore be out of range for the sequence dimension, making speculative decoding with top-p fail.

The rejection path also indexed `tokens[:, first_rejected_idx]`. With a different rejection position per row, PyTorch interprets this as a Cartesian selection, so each row writes every selected column. This silently corrupts accepted draft tokens and can copy one sample's resampled token into another sample.

The patch keeps `decode_speculative`'s existing batch-size limitation unchanged; it corrects the documented batched contract of the lower-level `sample_speculative` helper.

## Validation

On `origin/main`, the regressions reproduce:

- `RuntimeError: index ... is out of bounds for dimension 1` for 3D top-p logits
- cross-row output `[[1, 0, 3], [1, 2, 3]]` instead of `[[1, 0, 0], [2, 2, 3]]`

Patched checks:

```text
CUDA_VISIBLE_DEVICES='' PYTHONPATH=. python -m pytest -q tests/test_generation.py
2 passed

ruff check tests/test_generation.py
All checks passed!

black --check --config /dev/null --line-length 100 --target-version py39 tests/test_generation.py
1 file would be left unchanged.

git show --check HEAD
# clean
```

### DIFF
--- a/flash_attn/utils/generation.py
+++ b/flash_attn/utils/generation.py
@@ -61,7 +61,7 @@ def modify_logits_for_top_p_filtering(logits, top_p):
     sorted_indices_to_remove = cumulative_probs <= (1 - top_p)
     # scatter sorted tensors to original indexing
     indices_to_remove = sorted_indices_to_remove.scatter(
-        1, sorted_indices, sorted_indices_to_remove
+        -1, sorted_indices, sorted_indices_to_remove
     )
     logits.masked_fill_(indices_to_remove, float("-inf"))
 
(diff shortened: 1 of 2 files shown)


guoriyue:
Closing rather than overclaiming this low-level fix. The public speculative decoder is still batch-1, explicitly WIP, and does not expose the cross-row path fixed here; the tests cover helpers rather than an end-to-end public workflow. This should be revisited only with a deterministic end-to-end speculative-generation regression and model-call/speed evidence. The branch is preserved.

PR #2780, opened 2026-08-09
[CuTe, Bwd, SM100] Fix dV epilogue for asymmetric head dims

### LINKED ISSUE #2552: [Bug] SM100 backward kernel produces incorrect dV for MLA shapes whose `head_dim_v` isn't aligned to `dK_reduce_ncol`
## Summary

On Blackwell (SM100), `FlashAttentionBackwardSm100` computes a single `dK_reduce_ncol` from `head_dim` (Q/K) but reuses it for the dV TMEM-store atoms. When `(tile_hdimv // 2) % dK_reduce_ncol != 0`, the dV gradient is written with misaligned column strides, corrupting a small fraction of the output.

`_validate_head_dims` accepts the shape (it's inside the standard SM100 range), so the bug only surfaces in numerical comparison, not as a hard error.

## Affected versions / configurations

- `flash-attn-4 == 4.0.0b11` (confirmed); likely all prior `4.0.0b*` releases.
- GPU: SM100 (Blackwell, e.g. B200 / GB200). 
- Shapes: any MLA combination where `head_dim_qk >= 128` (triggers 2CTA backward) and `(head_dim_v / 16) * 16 / 2` is not divisible by `gcd(32, head_dim_qk / 16 * 16 / 2)`.
  - **Failing**: `(head_dim_qk, head_dim_v) = (128, 96)` → `gcd(32, 64)=32`, but `48 % 32 = 16 != 0`.
  - Also affected: `(128, 8)`, `(128, 16)`, `(128, 24)`, `(128, 40)`, `(128, 48)`, `(128, 72)`, `(128, 80)`, `(128, 88)`, `(128, 104)`, `(128, 112)`, `(128, 120)` (anything where `tile_hdimv // 2` is not a multiple of 32).
  - **Working** (already aligned): `(128, 64)`, `(128, 128)`, the dedicated `(256, 256)` kernel path, the DeepSeek `(192, 128)` path.

## Minimal reproducer

Standalone PyTorch script — compares FA4 against a PyTorch reference. Run on Blackwell.

```python
# pip install flash-attn-4==4.0.0b11
import torch

torch.manual_seed(0)
device, dtype = "cuda", torch.bfloat16

b, sq, skv = 2, 1024, 1024
h, hg = 16, 16
head_dim_qk = 128
head_dim_v  = 96         # any value where (ceil(d/16)*16)/2 % gcd(32, head_dim_qk/2) != 0

q = torch.randn(b, sq,  h,  head_dim_qk, device=device, dtype=dtype, requires_grad=True)
k = torch.randn(b, skv, hg, head_dim_qk, device=device, dtype=dtype, requires_grad=True)
v = torch.randn(b, skv, hg, head_dim_v,  device=device, dtype=dtype, requires_grad=True)

# FA4
out_fa, _ = flash_attn_func(q, k, v, causal=True)
g = torch.randn_like(out_fa)
out_fa.backward(g)
dq_fa, dk_fa, dv_fa = q.grad.clone(), k.grad.clone(), v.grad.clone()

# Reference (manual SDPA)
q.grad = k.grad = v.grad = None
scale = head_dim_qk ** -0.5
qf, kf, vf = q.float(), k.float(), v.float()
attn = torch.einsum("bshd,bShd->bhsS", qf, kf) * scale
mask = torch.ones(sq, skv, device=device, dtype=torch.bool).tril()
attn = attn.masked_fill(~mask, float("-inf"))
p = attn.softmax(-1)
out_ref = torch.einsum("bhsS,bShd->bshd", p, vf).to(dtype)
out_ref.backward(g)
dq_ref, dk_ref, dv_ref = q.grad.float(), k.grad.float(), v.grad.float()

# Compare
def err(a, b): return (a.float() - b).abs().max().item()
print(f"out diff     : {err(out_fa, out_ref):.4f}")  # ~0      (forward is correct)
print(f"dQ diff      : {err(dq_fa,  dq_ref):.4f}")   # small
print(f"dK diff      : {err(dk_fa,  dk_ref):.4f}")   # small
print(f"dV diff      : {err(dv_fa,  dv_ref):.4f}")   # >> 0.1, FAILS
```

Expected dV abs error: ~`1e-2` (bf16 noise).
Observed: `~3e-1`, with ~1.3% of elements mismatched.

## Root cause

`flash_attn/cute/flash_bwd_sm100.py:262`:

```python
self.dK_reduce_ncol = math.gcd(32, self.tile_hdim // 2)
```

`dK_reduce_ncol` is derived purely from `tile_hdim` (Q/K). The same value is then used to build the TMEM-load atom for dV in `epilogue_dK_or_dV_tma` (~`flash_bwd_sm100.py:3890`):

```python
self.sdV_layout = cute.make_layout((self.tile_n * self.dK_reduce_ncol, 2))
# ... and used identically for both dK and dV TMA atoms.
```

In 2CTA mode (`head_dim_qk >= 128`), each warpgroup owns `tile_hdimv // 2` columns of dV. The TMEM load atom reads `dK_reduce_ncol` columns per step. When `(tile_hdimv // 2) % dK_reduce_ncol != 0`, partitioning fails to cover the dV tile evenly, leading to misaligned stores.

For `(128, 96)`:
- `tile_hdim = 128`, `tile_hdimv = 96`
- `dK_reduce_ncol = gcd(32, 64) = 32`
- per-WG dV columns: `96 // 2 = 48`
- `48 % 32 = 16 ≠ 0` → misalignment → corrupt dV.

The forward and dQ/dK paths are unaffected; only dV is wrong.

## Proposed fix (preferred)

Make `dK_reduce_ncol` account for `tile_hdimv` so it divides both head dims:

```python
# flash_attn/cute/flash_bwd_sm100.py (~L262)
self.dK_reduce_ncol = math.gcd(32, self.tile_hdim // 2, self.tile_hdimv // 2)
```

Effect on supported shapes:

| `(d_qk, d_v)` | tile_hdim/2 | tile_hdimv/2 | old `dK_reduce_ncol` | new `dK_reduce_ncol` |
|---------------|-------------|--------------|---------------------|---------------------|
| (64, 64)      | 32          | 32           | 32                  | 32 (same)           |
| (128, 64)     | 64          | 32           | 32                  | 32 (same)           |
| (128, 128)    | 64          | 64           | 32                  | 32 (same)           |
| (128, 96)     | 64          | 48           | 32 → **misaligned** | **16 (fixes)**      |
| (192, 128)    | 96          | 64           | 32                  | 32 (same)           |
| (64, 512)     | 32          | 256          | 32                  | 32 (same)           |

Note: `dK_reduce_ncol = 16` halves the TMEM transaction size for the (128, 96) path, which costs ~one extra epilogue iteration on dV — a small perf hit on a previously-broken shape. All currently-working shapes are unaffected.

## Alternative fix (less ideal)

Reject the affected shapes in `_validate_head_dims` (`flash_attn/cute/interface.py:79`):

```python
elif compute_capability in [10, 11]:
    # ... existing checks ...
    if head_dim != head_dim_v and head_dim >= 128:
        tile_hdim   = math.ceil(head_dim   / 16) * 16
        tile_hdimv  = math.ceil(head_dim_v / 16) * 16
        assert (tile_hdimv // 2) % math.gcd(32, tile_hdim // 2) == 0, (
            f"(head_dim, head_dim_v)=({head_dim}, {head_dim_v}) is not supported on "
            f"SM100 backward: dV TMEM atom alignment requires "
            f"(tile_hdimv // 2) % gcd(32, tile_hdim // 2) == 0."
        )
```

This shrinks the supported shape set rather than fixing the underlying issue.

guoriyue:
## Summary

- derive independent TMEM reduction widths for dK and dV from their respective padded head dimensions
- use the dV width consistently in its FP32 staging layout, TMEM load atom, and reduce-add byte count
- preserve the existing dK transaction width instead of shrinking both paths to a common divisor
- add configuration regressions across asymmetric SM100 head dimensions

Fixes #2552.

## Problem

The shared SM100 dK/dV epilogue derives `dK_reduce_ncol` from the Q/K head dimension and previously reused it for dV. For `(head_dim, head_dim_v) = (128, 96)`, each dV warpgroup owns 48 columns, but the reused width is 32:

```text
(tile_hdimv / 2) % dK_reduce_ncol = 48 % 32 = 16
```

That width does not partition the dV tile evenly. Issue #2552 reports correct output/dQ/dK but corrupted dV on B200 for this configuration.

The patch specializes the existing shared epilogue at compile time: dK keeps its 32-column loads, while dV uses `gcd(32, tile_hdimv / 2)` (16 columns for the 128/96 case). Already-aligned shapes keep their existing widths.

## Validation

Configuration coverage on the patch:

```text
PYTHONPATH=.:tests/cute python -m pytest -q tests/cute/test_flash_bwd_sm100_config.py
49 passed
```

A fresh no-cache CuTe compile targeting `sm_100a` passed both affected storage paths for BF16 `(Dqk, Dv) = (128, 96)`:

```text
MHA [(1, 256, 8, 128), (1, 256, 8, 128), (1, 256, 8, 96)]
GQA [(1, 256, 8, 128), (1, 256, 4, 128), (1, 256, 4, 96)]
```

This covers direct dV storage and the GQA FP32 reduce-add path. The compile used:

```text
CUTE_DSL_NO_CACHE=1
FLASH_ATTENTION_ARCH=sm_100
CUTE_DSL_ARCH=sm_100a
FLASH_ATTENTION_NUM_SMS=148
```

Additional checks:

```text
ruff check flash_attn/cute/flash_bwd_sm100.py tests/cute/test_flash_bwd_sm100_config.py
All checks passed!

black --check --config tests/pyproject.toml tests/cute/test_flash_bwd_sm100_config.py
1 file would be left unchanged.

git show --check HEAD
# clean
```

I only have an RTX 5090 (SM120), so I could cross-compile but could not execute the SM100 numerical reproducer locally. The B200 numerical failure and expected error levels are documented in #2552; this remains a draft pending SM100 CI/runtime confirmation.

### DIFF
--- a/flash_attn/cute/flash_bwd_sm100.py
+++ b/flash_attn/cute/flash_bwd_sm100.py
@@ -259,8 +259,10 @@ class FlashAttentionBackwardSm100:
         self.dQaccum_reduce_stage = self.tile_hdim // self.dQ_reduce_ncol
         self.dQaccum_reduce_stage_t2r = self.tile_hdim // self.dQ_reduce_ncol_t2r
         self.cluster_reduce_dQ = False and cute.size(self.cluster_shape_mn) > 1
-        # number of tma reduce adds for dKacc and dVacc epilogue (must divide hdim_per_wg)
+        # Column widths for TMA reduce-add in the dKacc and dVacc epilogues. Each
+        # width must divide its own head dimension per warpgroup.
         self.dK_reduce_ncol = math.gcd(32, self.tile_hdim // 2)
+        self.dV_reduce_ncol = math.gcd(32, self.tile_hdimv // 2)
         # CTA group for MMA operations
         self.cta_group = tcgen05.CtaGroup.TWO if self.use_2cta_instrs else tcgen05.CtaGroup.ONE
 
@@ -437,8 +439,7 @@ class FlashAttentionBackwardSm100:
             )
         else:
             self.sdK_layout = cute.make_layout((self.tile_n * self.dK_reduce_ncol, 2))
-            # self.dK_reduce_ncol same for dV
-            self.sdV_layout = cute.make_layout((self.tile_n * self.dK_reduce_ncol, 2))
+            self.sdV_layout = cute.make_layout((self.tile_n * self.dV_reduce_ncol, 2))
 
     @cute.jit
     def __call__(
@@ -699,6 +700,7 @@ class FlashAttentionBackwardSm100:
         self.tma_copy_bytes["dPsum"] = self.tile_m * Float32.width // 8
         self.tma_copy_bytes["dQ"] = self.tile_m * self.dQ_reduce_ncol * Float32.width // 8
         self.tma_copy_bytes["dKacc"] = self.tile_n * self.dK_reduce_ncol * Float32.width // 8
+        self.tma_copy_bytes["dVacc"] = self.tile_n * self.dV_reduce_ncol * Float32.width // 8
         self.tma_copy_bytes["dS"] = cute.size_in_bytes(self.ds_dtype, self.sdS_layout)
         self.tma_copy_bytes["sdS_xchg"] = self.tma_copy_bytes["dS"] // 2  # Half of dS for exchange
 
@@ -4010,6 +4012,12 @@ class FlashAttentionBackwardSm100:
         flat_epi_tile = (
             self.sdK_flat_epi_tile if const_expr(K_or_V == "K") else self.sdV_flat_epi_tile
         )
+        reduce_ncol = self.dK_reduce_ncol if const_expr(K_or_V == "K") else self.dV_reduce_ncol
+        reduce_copy_bytes = (
+            self.tma_copy_bytes["dKacc"]
+            if const_expr(K_or_V == "K")
+            else self.tma_copy_bytes["dVacc"]
+        )
         num_compute_threads = cute.arch.WARP_SIZE * len(self.compute_warp_ids)
         wg_idx = (cute.arch.thread_idx()[0] % num_compute_threads) // 128
         num_wg = num_compute_threads // 128
@@ -4080,7 +4088,7 @@ class FlashAttentionBackwardSm100:
             )
 
         tmem_load_atom = cute.make_copy_atom(
-            tcgen05.copy.Ld32x32bOp(tcgen05.copy.Repetition(self.dK_reduce_ncol)), Float32
+            tcgen05.copy.Ld32x32bOp(tcgen05.copy.Repetition(reduce_ncol)), Float32
         )
 
         read_flag = const_expr(not deterministic_KV)
@@ -4143,7 +4151,7 @@ class FlashAttentionBackwardSm100:
                         copy_utils.cpasync_reduce_bulk_add_f32(
                             sdKV.iterator,
                             gdKV_epi[None, epi_stage].iterator,
-                            self.tma_copy_bytes["dKacc"],
+                            reduce_copy_bytes,
                         )
                 if const_expr(epi_stage < num_epi_stages - 1):
                     cute.arch.cp_async_bulk_commit_group()
--- a/tests/cute/test_flash_bwd_sm100_config.py
+++ b/tests/cute/test_flash_bwd_sm100_config.py
@@ -1,0 +1,36 @@
+import math
+
+import pytest
+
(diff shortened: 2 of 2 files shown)


guoriyue:
Closing until this SM100-specific change can be validated on B200/B300 hardware. Local config/compile tests support the diagnosis, but RTX 5090 cannot execute SM100 kernels and therefore cannot prove the reported dV numerical correction or performance/no-regression behavior. Keeping the branch for a future hardware-backed submission.

PR #2405, opened 2026-03-28
feat: setup_context for FlashAttnFunc (torch.func.grad)

NJX-njx:
## Summary

Implements setup_context on FlashAttnFunc so 	orch.func.grad / functorch transforms work (#2071), per maintainer guidance.

- When gradients are enabled, forward returns extra tensors for context; non-differentiable outputs are marked accordingly.
- The user-visible output is cloned when saving for backward so it is not a view of out_padded (which is marked non-differentiable).
- lash_attn_func unpacks the 5-tuple training path so the public API is unchanged.
- Adds 	ests/test_flash_attn_functorch.py (skipped if lash_attn_2_cuda is not built).

## Testing

- pytest tests/test_flash_attn_functorch.py (skipped here without CUDA extension build).

Fixes #2071

PR #2781, opened 2026-08-09
fix: mark FA2 auxiliary outputs non-differentiable

### LINKED ISSUE #1817: softmax_lse is silently not backpropogated
`softmax_lse` currently does not support backpropogation. This is fine, however the fact that it doesn't trigger any error makes it problematic IMO.
Suggested fix: Add something like
```
for d in args:
    torch._assert_async((d == 0).all())
```
to the backward functions in `flash_attn_interface.py` (what I don't like about this specific solution is that the error messages are not very informative).
Minimal example:
```
import torch
from flash_attn.flash_attn_interface import flash_attn_func

B, SEQ_LEN, H, D = 4, 2048, 16, 64
dtype = torch.bfloat16
device = 'cuda'

q = torch.randn(B, SEQ_LEN, H, D, device=device, dtype=dtype, requires_grad=True)
k = torch.randn(B, SEQ_LEN, H, D, device=device, dtype=dtype, requires_grad=True)
v = torch.randn(B, SEQ_LEN, H, D, device=device, dtype=dtype, requires_grad=True)

attn_output, softmax_lse_flash, _ = flash_attn_func(q=q, k=k, v=v, return_attn_probs=True)

loss = softmax_lse_flash.mean()
# loss = attn_output.mean() # this case is fine
loss.backward()

# prints zeroes, which is incorrect
print(q.grad.abs().max().item())
print(k.grad.abs().max().item())
print(v.grad.abs().max().item())
```

zupengwang:
## Summary

- mark `softmax_lse` and `S_dmask` as non-differentiable in all six FA2 autograd wrappers when `return_attn_probs=True`
- preserve the existing gradient path through the primary `out` tensor
- cover fixed-length and varlen APIs in QKV-packed, KV-packed, and separate Q/K/V forms

Fixes #1817.

## Motivation

The auxiliary outputs are intended for testing and their gradients are not implemented. Previously, a loss built only from `softmax_lse` silently materialized zero gradients for Q, K, and V. Marking both auxiliary outputs as non-differentiable makes that unsupported use fail explicitly while leaving normal output backpropagation unchanged.

Related work: #2405 independently marks the auxiliary outputs in `FlashAttnFunc` while adding `setup_context` support. This change applies the #1817 behavior consistently to all six current FA2 wrappers.

## Testing

- source build: `FLASH_ATTENTION_FORCE_BUILD=TRUE FLASH_ATTN_CUDA_ARCHS=80 MAX_JOBS=16 NVCC_THREADS=2 python -m pip install -e . --no-build-isolation --no-deps`
- regression: `python -m pytest -q -s tests/test_flash_attn.py::test_flash_attn_aux_outputs_are_non_differentiable` (`6 passed`)
- upstream PyTorch-reference helpers: 7 forward/backward cases across fixed/varlen, FP16/BF16, causal/non-causal, dropout `0.0/0.17`, and head dimensions `32/64/128/192/256`
- syntax and whitespace: `python -m py_compile flash_attn/flash_attn_interface.py tests/test_flash_attn.py` and `git diff --check`

Validated on an NVIDIA GeForce RTX 3090 (SM86) with PyTorch `2.13.0+cu130` and CUDA 13.0. Only one GPU was used because the change has no distributed semantics.

### DIFF
--- a/flash_attn/flash_attn_interface.py
+++ b/flash_attn/flash_attn_interface.py
@@ -505,7 +505,10 @@ class FlashAttnQKVPackedFunc(torch.autograd.Function):
             ctx.alibi_slopes = alibi_slopes
             ctx.deterministic = deterministic
         out = out_padded[..., :head_size_og]
-        return out if not return_softmax else (out, softmax_lse, S_dmask)
+        if not return_softmax:
+            return out
+        ctx.mark_non_differentiable(softmax_lse, S_dmask)
+        return out, softmax_lse, S_dmask
 
     @staticmethod
     def backward(ctx, dout, *args):
@@ -595,7 +598,10 @@ class FlashAttnVarlenQKVPackedFunc(torch.autograd.Function):
             ctx.alibi_slopes = alibi_slopes
             ctx.deterministic = deterministic
         out = out_padded[..., :head_size_og]
-        return out if not return_softmax else (out, softmax_lse, S_dmask)
+        if not return_softmax:
+            return out
+        ctx.mark_non_differentiable(softmax_lse, S_dmask)
+        return out, softmax_lse, S_dmask
 
     @staticmethod
     def backward(ctx, dout, *args):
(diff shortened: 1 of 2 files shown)


bvolpato (approved):
LGTM. Nice focused fix, consistent across all six wrappers, with primary `out` gradient path explicitly covered.

I ran all three fixed-length regression variants on CUDA, plus an existing zero-dropout forward/backward reference case. `py_compile` and `git diff --check` pass. I also reviewed varlen paths statically; local varlen execution was blocked by an older extension ABI, while the PR source-build run reports all six passing.

PR #2788, opened 2026-08-11
publish.yml: trigger the release workflow on tag pushes

### LINKED ISSUE #2769: publish.yml: create trigger has no tag filter, so branch pushes fail the workflow and new matrix entries never build
## Summary

`.github/workflows/publish.yml` triggers on:

```yaml
on:
  create:
    tags:
      - v*
```

The `create` event does not support filters, so the `tags:` key is silently ignored and the workflow also fires on **branch** creation. `setup_release` then runs:

```bash
gh release create ${GITHUB_REF#refs/tags/} --repo $GITHUB_REPOSITORY --title ... --generate-notes
```

For a branch, `refs/heads/...` does not match the `refs/tags/` prefix, so the ref is passed through unstripped and the step fails:

```
HTTP 422: Validation Failed (https://api.github.com/repos/Dao-AILab/flash-attention/releases)
pre_receive Sorry, branch or tag names starting with 'refs/' are not allowed.
Published releases must have a valid tag
```

`build_wheels` and `publish_package` are `needs:`-dependent, so both are skipped.

## Impact

Almost every run of "Build wheels and deploy" is a failure triggered by an ordinary branch push — e.g. [run 31060582771](https://github.com/Dao-AILab/flash-attention/actions/runs/31060582771) on `refs/heads/drisspg/stack/60`. Of the last 100 runs, 99 are branch refs (95 failed, 4 cancelled) and the single success is the tag run `v2.8.3.post1`. The real signal is buried in that noise.

Worth noting the compute cost is near zero — every one of those runs dies in `setup_release` within seconds and the matrix never starts. The cost is signal: the workflow reads as permanently broken, and a genuine release failure would be indistinguishable from the background.

## A separate problem: the current matrix has never run

This is **not** caused by the trigger bug, and I want to be precise about that since the two are easy to conflate. The tag path works correctly under `create` — [run 27279915131](https://github.com/Dao-AILab/flash-attention/actions/runs/27279915131) is the `v2.8.3.post1` run and it succeeded and published.

A tag-triggered run uses the workflow *as it exists at that tag*, and `v2.8.3.post1` points at a commit off a line that diverged from `main` well before the matrix was updated:

```
v2.8.3.post1...main      status: diverged, ahead_by 558, behind_by 2
v2.8.3.post1...c04a808   status: diverged, ahead_by 357, behind_by 2
```

So c04a808 (adds torch 2.9.1 / 2.10.0 and CUDA 13) is not in the tag's tree. Reading `publish.yml` at the tag confirms it still carries `python-version: ['3.9' … '3.13']` and `torch-version: ['2.4.0' … '2.8.0']` — which is why the newest tag has *older* torch coverage than `main` declares.

The practical consequence: fixing the trigger does not by itself build any wheels, and a tag cut from current `main` would run the current matrix with or without the trigger fix. Both changes are worth making; neither substitutes for the other.

Net result today: no cp310 / cp311 / cp313 wheels exist for torch >= 2.9. The only ones published are cp312 — five on `v2.8.3` (uploaded 2025-12-17, 2026-06-02, 2026-06-11, 2026-06-12) and two on `v2.8.3.post1` (2026-06-11) — and those scattered individual timestamps suggest they were added by hand rather than by a matrix run.

## Suggested fix

Move to a trigger that supports filtering:

```yaml
on:
  push:
    tags:
      - 'v*'
```

`publish-fa4.yml` already uses exactly this form (with `fa4-v*`), and the globs do not collide since the pattern is anchored at the start.

Gating the first job instead would also stop the failures:

```yaml
jobs:
  setup_release:
    if: startsWith(github.ref, 'refs/tags/')
```

but it is the weaker option — it still creates a workflow run per branch, just one where every job is skipped. The `push` trigger creates no run at all.

One behavioural difference to be aware of with `push`: `create` fires only on ref creation, while `push` also fires on ref *update*, so force-moving an existing `v*` tag would re-run the workflow and `gh release create` would fail on the already-existing release. `if: github.event.created` on `setup_release` guards it if that matters.

## Question

Is a release cut from current `main` the intended path to torch 2.9.1 / 2.10.0 wheels for python 3.10 / 3.11 / 3.13, or is the cp312-only set deliberate? We're on cp311 with torch 2.9, and the answer decides whether we build the wheel ourselves — happy to do that, just would rather not duplicate work you already intend to publish.

---

*Edit 2026-08-11: corrected the framing of the matrix section. It originally called this a "second-order effect" of the trigger bug, which implies a causal link that does not exist — the tag path works fine under `create`, and the matrix gap is caused solely by the tag pointing at older code. Also corrected the run statistics to the full 100-run sample and fixed asset timestamps that were misattributed between `v2.8.3` and `v2.8.3.post1`.*

aryanputta:
Fixes #2769.

## Problem

`publish.yml` triggers on:

```yaml
on:
  create:
    tags:
      - v*
```

The `create` event does not support filters, so the `tags` key is ignored and the workflow runs on every branch creation as well. `setup_release` then runs:

```bash
gh release create ${GITHUB_REF#refs/tags/} ...
```

On a branch the ref is `refs/heads/...`, which does not match the stripped prefix, so the unstripped ref reaches the API and the step fails:

```
pre_receive Sorry, branch or tag names starting with 'refs/' are not allowed.
Published releases must have a valid tag
```

`build_wheels` and `publish_package` both list `setup_release` in `needs`, so they skip.

Checking the workflow's own history, of the last 40 runs **39 failed and one succeeded**, and the single success is the `v2.8.3.post1` tag. Every other run was a branch push.

## Fix

Switch to the `push` trigger, which honors the tag filter.

`publish-fa4.yml` in this repo already uses exactly this form:

```yaml
on:
  push:
    tags:
      - 'fa4-v*'
```

so this brings the older workflow in line with the newer one. `publish.yml` is the only workflow here still on a `create` trigger.

The `${GITHUB_REF#refs/tags/}` expression in `setup_release` is unchanged and continues to work, since a tag push sets `GITHUB_REF` to `refs/tags/<tag>`.

## Note on the matrix

The issue also observes that a tag-triggered run uses the workflow as it existed at that tag, so the matrix added in c04a808 has never run, and no cp310 / cp311 / cp313 wheels exist for torch >= 2.9. This change is what makes the next tag actually run the current matrix. Which versions to publish is a maintainer decision and is not touched here.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -8,9 +8,9 @@
 name: Build wheels and deploy
 
 on:
-  create:
+  push:
     tags:
-      - v*
+      - 'v*'
 
 jobs:
   setup_release:


joanfabregat:
Author of #2769 here. Thanks for picking this up — the trigger change is correct, and it is the better of the two options I listed. `if: startsWith(github.ref, 'refs/tags/')` would still create a workflow run per branch, just one where every job is skipped; `push` with a `tags:` filter creates no run at all.

Worth stating explicitly for whoever reviews this, since it is the thing that makes the change safe to merge: `GITHUB_REF` is `refs/tags/<tag>` and `GITHUB_SHA` is the tagged commit under *both* events, so `${GITHUB_REF#refs/tags/}` is genuinely unaffected and the tag path cannot regress. The diff really is as small as it looks.

Two notes.

**1. The matrix claim is overstated — and that is my fault, not yours.**

The PR says "This change is what makes the next tag actually run the current matrix." The tag path already works under `create`: run 27279915131 is the `v2.8.3.post1` run, and it succeeded and published. The current matrix has never executed for an unrelated reason:

```
v2.8.3.post1...main      status: diverged, ahead_by 558, behind_by 2
v2.8.3.post1...c04a808   status: diverged, ahead_by 357, behind_by 2
```

`v2.8.3.post1` points at a commit off a line that diverged from `main` 558 commits back, so c04a808 is not in its tree — reading `publish.yml` at the tag confirms it still carries `python-version: ['3.9' … '3.13']` and `torch-version: ['2.4.0' … '2.8.0']`.

So the two are independent: merging this produces zero wheels, and a tag cut from current `main` runs the current matrix (56 build jobs by my count) with or without it. My issue described the matrix gap as a "second-order effect" of the trigger bug, which reads as a causal link that isn't there, and I think that is where the framing came from. I have amended the issue.

This is not a reason to hold the PR — just worth not merging it under the expectation that wheels appear afterwards. They will not; that needs a tag.

**2. `push` also fires on tag *updates*.**

`create` fires only on ref creation. `push` fires on ref update too, so force-moving an existing `v*` tag would re-trigger the workflow and `gh release create` would fail with a 422 against the already-existing release. Arguably a rebuild is what you would want there, so this may be fine as-is; `if: github.event.created` on `setup_release` covers it if not.

The noise fix stands on its own regardless: of the last 100 runs of this workflow, 99 are branch refs (95 failed, 4 cancelled) and exactly one is the tag.

PR #2792, opened 2026-08-12
feat: support per-head rotary frequencies

### LINKED ISSUE #578: Feature request: allow rotary cos/sin to vary per head
I have noticed that rotary does better on my model, especially for small heads which have few rotary frequencies, when I apply a different set of frequencies per head (it is able to use the difference between the heads to extract finer position information). I did this using my own pure PyTorch apply_rotary_emb(), but I currently use apply_rotary_emb_qkv_() for speed and memory savings and that does not support a sin/cos that is different for each head. Could this be added?

Thank you,
Katherine Crowson

zupengwang:
## Summary

- accept per-head rotary cosine and sine tables with shape `(seqlen, nheads, rotary_dim / 2)`
- preserve the existing shared-table Q/K fast path and support distinct Q/K head counts for GQA
- add FP16/BF16 forward and backward coverage for standard QKV and GQA layouts

Fixes #578.

## Testing

- `CUDA_VISIBLE_DEVICES=0 python -m pytest -q tests/test_rotary.py -k per_head -x` — 8 passed
- `CUDA_VISIBLE_DEVICES=1 python -m pytest -q tests/test_rotary.py -k per_head -x` — 8 passed
- `CUDA_VISIBLE_DEVICES=0 python -m pytest -q tests/test_rotary.py` — 224 passed, 1 failed; the failure is the pre-existing `test_compilation_count` incompatibility with Triton 3.7.1 and reproduces on the clean base
- `python -m py_compile flash_attn/ops/triton/rotary.py flash_attn/layers/rotary.py tests/test_rotary.py`
- `git diff --check`

### DIFF
--- a/flash_attn/layers/rotary.py
+++ b/flash_attn/layers/rotary.py
@@ -104,7 +104,8 @@ def apply_rotary_emb(
     Arguments:
         x: (batch_size, seqlen, nheads, headdim) if cu_seqlens is None
             else (total_seqlen, nheads, headdim)
-        cos, sin: (seqlen_rotary, rotary_dim / 2)
+        cos, sin: (seqlen_rotary, rotary_dim / 2) or
+            (seqlen_rotary, nheads, rotary_dim / 2)
         interleaved: if True, rotate pairs of even and odd dimensions (GPT-J style) instead
             of 1st half and 2nd half (GPT-NeoX style).
         inplace: if True, apply rotary embedding in-place.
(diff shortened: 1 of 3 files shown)


bvolpato (approved):
LGTM. I like that this preserves the shared-table single-kernel fast path while adding separate per-head Q/K handling for distinct GQA head counts.

Tested on an RTX 5070 Ti:
- Per-head FP16/BF16 forward and backward cases: 8 passed
- Full rotary suite: 224 passed
- `git diff --check`: passed

Only `test_compilation_count` failed because Triton 3.7.1 removed `JITFunction.cache_hook`; same known base incompatibility.

PR #2802, opened 2026-08-17
[FA4] Normalize scalar tensor max sequence lengths before compilation

### LINKED ISSUE #2801: [FA4] max_seqlen_q tensor causes an argument-annotation mismatch in varlen forward
## Summary

The FA4 varlen forward path can pass a scalar `torch.Tensor` for `max_seqlen_q` into `cute.compile()`, while the compiled function declares that parameter as `Int32 | int | None`. This makes compilation fail with an argument-annotation mismatch.

## Reproduction

From a checkout of the current default branch, run:

```bash
pytest tests/cute/test_flash_attn_fast.py -k test_flash_attn_varlen_tensor_max_seqlen_reuses_fwd_cache
```

## Actual behavior

The test fails with an error equivalent to:

```
expects argument #26 (max_seqlen_q) to be one of (Int32, int, NoneType), but got torch.Tensor
```

## Expected behavior

The cache-stabilization path should retain its intended behavior while passing a value consistent with the compiled function's type contract.

## Possible direction

Normalize the scalar before invoking `cute.compile()` (for example, construct the appropriate CUTLASS DSL scalar type), or update the relevant API contract if a tensor is the intended representation.

TorinLi:
## Summary
- normalize scalar tensor `max_seqlen_q` and `max_seqlen_k` values before forward-kernel compilation
- preserve the compiled kernel's `Int32 | int | None` scalar contract
- make the existing cache-reuse regression test pass for fresh scalar tensor inputs

Fixes #2801.

## Validation
- `python3 -m py_compile flash_attn/cute/interface.py`
- The targeted CUDA test was not run locally because this environment has no CUDA device.

### DIFF
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -78,6 +78,15 @@ BIN_BATCH_SEARCH_THRESH = 256  # above this batch size SingleTileVarlenScheduler
 USE_BLOCKS_TO_BATCH: bool = True
 
 
+def _normalize_max_seqlen(max_seqlen: Optional[int]) -> Optional[int]:
+    """Return a host scalar for compiled forward-kernel scalar arguments."""
+    if isinstance(max_seqlen, torch.Tensor):
+        if max_seqlen.numel() != 1:
+            raise ValueError("max_seqlen must be a scalar")
+        return int(max_seqlen.item())
+    return max_seqlen
+
+
 def _parse_arch_str(arch_str):
     """Parse arch string (e.g. 'sm_80', 'sm_90a', '80', '100') to int (e.g. 80, 90, 100)."""
     import re
@@ -574,6 +583,8 @@ def _flash_attn_fwd(
         aux_tensors: Some score_mods will want to read from global aux_tensors. This is how we thread them through to the inner kernel.
         aux_scalars: Runtime scalar captures used by score_mod or mask_mod.
     """
+    max_seqlen_q = _normalize_max_seqlen(max_seqlen_q)
+    max_seqlen_k = _normalize_max_seqlen(max_seqlen_k)
     aux_scalars = tuple(aux_scalars) if aux_scalars else None
     requires_grad = any(
         t is not None and t.requires_grad for t in (q, k, v, qv, learnable_sink)
--- a/flash_attn/cute/interface.py
+++ b/flash_attn/cute/interface.py
@@ -583,8 +583,6 @@ def _flash_attn_fwd(
         aux_tensors: Some score_mods will want to read from global aux_tensors. This is how we thread them through to the inner kernel.
         aux_scalars: Runtime scalar captures used by score_mod or mask_mod.
     """
-    max_seqlen_q = _normalize_max_seqlen(max_seqlen_q)
-    max_seqlen_k = _normalize_max_seqlen(max_seqlen_k)
     aux_scalars = tuple(aux_scalars) if aux_scalars else None
     requires_grad = any(
         t is not None and t.requires_grad for t in (q, k, v, qv, learnable_sink)
@@ -793,6 +791,11 @@ def _flash_attn_fwd(
         max_seqlen_k = seqlen_k
     if cu_seqlens_k is None and seqused_k is None:
         min_seqlen_k = seqlen_k
+
+    # SM100/SM110 pass max_seqlen_q as a typed kernel argument.  On other
+    # architectures, preserve a tensor value to avoid an unnecessary host sync.
+    if arch // 10 in [10, 11]:
+        max_seqlen_q = _normalize_max_seqlen(max_seqlen_q)
 
     fwd_cfg = _get_fwd_config(
         arch=arch,


TorinLi:
Hi @eamonn-zh @tridao @jayhshah , can you help review this PR? Thanks!

eilamc14:
Verified this on a B200 (SM100) the arch where #2801 reproduces, since `max_seqlen_q` is
only passed as a kernel argument on the `arch // 10 in [10, 11]` path.

**The test it targets**, on `main` @ `0251105` before and after applying this PR:

```
before:  1 failed, 240 deselected in 3.71s
after:   1 passed, 240 deselected in 6.43s
```

**Cache reuse with a fresh 0-dim CUDA `max_seqlen` every call**, counting
`cutlass.cute.compile` invocations (`max_seqlen = (cu[1:] - cu[:-1]).max()` rebuilt each
iteration):

| | call 1 | call 2 | call 3 | call 4 |
|---|---|---|---|---|
| forward only | 1 | 0 | 0 | 0 |
| forward + backward | 4 | 0 | 0 | 0 |

So the key is stable across calls, not just within the one the test covers.

**Regression check:** the varlen selection of `tests/cute/test_flash_attn_fast.py` -
`-k varlen`, 133 tests passes with the PR applied (537 s). I did not run the full
`tests/cute` suite.

Environment: NVIDIA B200, compute capability 10.0, torch 2.11.0+cu130 / CUDA 13.0,
nvidia-cutlass-dsl 4.6.0.dev0, transformers 5.12.1.

For context on why this matters beyond the test: on 4.0.0b19 (which predates both the DSL
signature check and #2507) the same tensor `max_seqlen` silently recompiled the backward on
every call - 276 backward objects and climbing at ~2 min each on a 27B multimodal training
step (Qwen3.x-27B), presenting as a hang with the GPU at 0% util. With this PR the same run completes,
and FA4 measures at parity with FA2 for me on that model.

---

One observation while testing, offered as optional hardening rather than a request. The
backward still receives a tensor: normalization happens inside `_flash_attn_fwd`, which
rebinds a local, while `FlashAttnVarlenFunc.forward` sets `ctx.max_seqlen_q` from its own
parameter. I instrumented `_flash_attn_bwd` and it does get `torch.Tensor`:

```
max_seqlen_q reaching _flash_attn_bwd: Tensor | is tensor: True
```

That is harmless today, for two reasons that both have to keep holding: #2507 guards the
backward compile key with `not torch.is_tensor(seqlen_q_rounded)`, and the backward does not
pass `max_seqlen` as a kernel argument the way the SM100 forward does. Normalizing in
`FlashAttnVarlenFunc.forward` before the `ctx` assignment (or additionally at the top of
`_flash_attn_bwd`) would make the invariant local instead of dependent on those two, at no
extra sync - the value has already been brought to the host by then.

the PR fixes the reported bug as it stands.


---

One tradeoff worth naming: `_normalize_max_seqlen` calls `.item()`, so the tensor path now syncs on
every forward. #2762 avoided that deliberately. The sync looks hard to avoid where the SM100 path
passes `max_seqlen_q` as a kernel argument against `Int32 | int | None`, and one sync clearly beats a
recompile per call, but it is per attention call, and it undoes the reason HF keeps `max_seqlen` on
device in padding-free attention.

The case I'd flag as a question rather than a finding, since I haven't tested it: `.item()` isn't
capturable and forces a graph break under Dynamo, so padding-free training combining `torch.compile`
with cudagraphs may go from "recompiles" to "can't capture".

Two optional ways to narrow the cost:

- Normalize in `FlashAttnVarlenFunc.forward` before the `ctx.max_seqlen_q` assignment: one sync per
  call instead of a local rebind, and it folds in the backward invariant above.
- Only normalize where the value actually reaches the kernel signature. On the non-SM100 paths it
  doesn't, so #2762's compile-key guard already suffices and those callers needn't sync at all.

TorinLi:
@eamonn-zh，can you help review this? 
Many thanks.

eamonn-zh (commented):
Thanks for the fix. This addresses #2801 on SM100/SM110, where max_seqlen_q is passed directly as a typed kernel argument.

One concern is that the current implementation normalizes both max_seqlen_q and max_seqlen_k unconditionally at the beginning of _flash_attn_fwd. For CUDA scalar inputs, .item() introduces a device-to-host synchronization on every forward, including paths where these host scalars are not required. #2762 intentionally avoided this sync while fixing the compile-cache issue.

Would it make sense to only normalize max_seqlen_q on the SM100/SM110 path where it is actually passed as a kernel argument? This should preserve the fix for #2801 without adding unnecessary synchronization on other paths.

TorinLi:
Thanks @eamonn-zh — agreed. I updated the PR to normalize only `max_seqlen_q`, and only on the SM100/SM110 path where it is passed as a typed forward-kernel argument. `max_seqlen_k` and non-SM100/SM110 paths now retain their tensor values, avoiding the unnecessary `.item()` host synchronization while preserving the #2801 fix.

Updated in 6588ac0.

eamonn-zh (approved):
LGTM, thanks for the update!

TorinLi:
@jayhshah Could you please take a final look and merge this PR if it looks good?

PR #2804, opened 2026-08-17
[CuTe] Speed up SM100 hdim-64 bwd: dedicated P/dS TMEM slots + per-warp signaling

jack-carlisle-2025:
Two coupled changes to the hdim-64 SM100 backward that together remove every compute-wide barrier from its main loop. Gradients are bit-identical to the current kernel under `deterministic=True`; every other shape and architecture is untouched.

**Results** (B200, bf16, 64k tokens per call, 32 query heads, sustained-load medians, shipped default vs current main):

| shape | mode | before, TFLOP/s | after, TFLOP/s | speed-up |
|---|---|--:|--:|--:|
| b2 s32k h32:32 | dense | 731 | 841 | 1.150x |
| b2 s32k h32:32 | causal | 705 | 808 | 1.147x |
| b4 s16k h32:8 | dense | 840 | 903 | 1.075x |
| b4 s16k h32:8 | causal | 750 | 842 | 1.123x |
| b8 s8k h32:32 fp16 | dense | 692 | 782 | 1.131x |
| b4 s16k h32:32 | dense, deterministic | 709 | 774 | 1.091x |
| b4 s16k h32:32 | causal, deterministic | 664 | 725 | 1.093x |

Every measured hdim-64 shape improves by 6-15%, and the gain grows with sequence length. A 124-configuration sweep (seqlen 512 to 32k, MHA/GQA/MQA, dense, causal, local, varlen, bf16/fp16, deterministic and not) gives a geometric-mean speed-up of **1.129x non-deterministic** (worst cell 1.023x) and **1.050x deterministic** (worst cell 0.995x). hdim 128 is unchanged.

## Why the hdim-64 backward is slow

At hdim 64 the 1-CTA backward only uses TMEM columns [0, 384), and the bf16 P and dS tiles the compute warps write are aliased over the fp32 S and dP accumulators. Because each warp may only touch its own 32-lane sector, and the eight compute warps span two warpgroups, warp w's P store lands under warp w+4's S lanes. The kernel therefore guards each overwrite with a compute-wide named barrier (an "alias guard"): one for P over S, two for dS over dP. Two more compute-wide barriers sit in front of the signals the compute warps send after their stores. Five barriers per tile hold all eight warps to the pace of the slowest, and the aliasing also forces the MMA warp to wait for PV_t before it can issue QK_{t+1}, so the tensor core idles through most of the softmax.

## What changes

**1. Dedicated TMEM slots for P and dS (`split_P_dS`).** P moves to columns [384, 448) and dS to [448, 512). With nothing aliased:
- the alias guards have nothing to guard and are removed; the per-warp `tcgen05.wait::ld` in front of each stays, since each warp still has to wait for its own asynchronous TMEM loads;
- the fused "S read and P written" handshake splits in two: `pipeline_S_P` now carries only "S read", signalled as soon as S is in registers, and a new one-stage `pipeline_P` carries "P written" / "P consumed";
- the MMA warp issues QK_{t+1} on "S read" alone, ahead of PV_t. Its issue order becomes `QK_{t+1}, PV_t, dK_t, dQ_t, dP_{t+1}`, so the next S tile is computed under the current softmax.

**2. Per-warp signaling (`warp_sync`).** The three remaining compute-wide barriers each sit in front of an mbarrier that already expects one arrival per compute warp, so the barrier was redundant convergence. Each is replaced by `sync_warp`; the mbarrier is the cross-warp join. This is only legal once the alias guards are gone, which is why the two ship together.

Both knobs are in the JIT cache key and default to on wherever legal (1-CTA, 128x128 tiles, hdim 64). `split_P_dS=False, warp_sync=False` reproduces the original control flow exactly, which is what makes the differential tests below possible. `warp_sync` is asserted off at 2-CTA: there the dS/dPsum site's barrier orders the compute warps' `sdS_xchg` writes before a single-thread DSMEM copy to the peer CTA.

**Ablation** (each change alone vs both, sustained load): dedicated slots alone are worth about 2% on dense and 3-4% on causal shapes; per-warp signaling alone is worth 8-9% on dense and roughly nothing on causal, because the alias guards inside the dS stage loop have to stay; together they are worth 13-15% on dense and 11-15% on causal. Most of the win is only reachable once the alias guards are gone.

## Correctness

Neither change alters arithmetic or reduction order, so with `deterministic=True` dq/dk/dv are bit-identical to the current kernel. This was checked with a differential suite against the baseline knobs (`split_P_dS=False, warp_sync=False`), which reproduce the original control flow exactly: bit-exact comparison across dense, causal, GQA, local-window, varlen and odd/ragged sequence lengths, both iteration parities and `main_loop_iters == 0`; the same oracle under SM contention, an occupancy sweep and interleaved launches; a subprocess liveness watchdog; and differential `compute-sanitizer` runs with no new report attributable to the change. The suite's sensitivity was measured rather than assumed: sabotage mutants that re-alias the TMEM regions with the barriers removed are each caught within seconds. Static checks confirm the dealiasing introduces no new TMEM overlap and that every barrier weakened to `sync_warp` fronts an mbarrier expecting one arrival per compute warp. On the non-deterministic path, max error against an fp32 reference matches the baseline everywhere.

The hdim-64 subsets of `test_flash_attn.py` and `test_flash_attn_varlen.py` pass on a B200 with no new failures.

**Measurement.** Per-iteration CUDA events over back-to-back call trains, arms interleaved with rotating order, median of kept iterations; clocks unlocked, so the trains ride the power cap, which is the training-representative regime.

### DIFF
--- a/flash_attn/cute/flash_bwd_sm100.py
+++ b/flash_attn/cute/flash_bwd_sm100.py
@@ -68,6 +68,8 @@ class FlashAttentionBackwardSm100:
         has_aux_tensors: cutlass.Constexpr = False,
         q_subtile_factor: cutlass.Constexpr[int] = 1,
         kv_subtile_factor: cutlass.Constexpr[int] = 1,
+        split_P_dS: Optional[bool] = None,
+        warp_sync: Optional[bool] = None,
     ):
         # padding head_dim to a multiple of 16 as k_block_size
         hdim_multiple_of = 16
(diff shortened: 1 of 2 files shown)


drisspg (approved):
Hey pushed a few tweaks but going to land this now :)

Issue #2729, opened 2026-07-25
[FA4/B200] return_lse=True makes D64 forward about 19% faster


drisspg:
I had my agent create this reproducer while doing some heuristic tuning. This is quite strange, and I wanted to track it.

On B200, the FA4 CuTe forward kernel for dense BF16 attention with `(B, S, H, D) = (2, 16384, 16, 64)` is consistently faster when `return_lse=True`, even though that specialization calculates and writes the additional LSE output.

## Environment

- GPU: NVIDIA B200 (SM100)
- PyTorch: `2.14.0.dev20260722+cu132`
- PyTorch CUDA runtime: 13.2
- NVIDIA CuTe DSL: `4.6.0.dev0`
- FlashAttention commit: `00756db9d921da0846453283ddfbeb7457abd09b`
- Dtype: BF16
- Causal: false
- Measurement: fixed-pointer CUDA graph replay, 25 warmups and 100 measured replays per round, five alternating A/B rounds

## Result

```text
GPU: NVIDIA B200
Torch: 2.14.0.dev20260722+cu132 (CUDA 13.2)
CuTe DSL: 4.6.0.dev0
FA4 commit: 00756db9d921da0846453283ddfbeb7457abd09b
no-LSE: median=3.0018 ms rounds=[3.007457275390625, 3.0017739868164064, 3.001456604003906, 3.0028912353515627, 3.0008636474609376]
with-LSE: median=2.5281 ms rounds=[2.5091484069824217, 2.545796203613281, 2.512978210449219, 2.5281126403808596, 2.532301177978516]
with-LSE speedup: 1.187x
```

Both paths pass a correctness check against PyTorch SDPA.

A focused Nsight Compute comparison showed the same launch topology and resource limits for both specializations (148 blocks, 512 threads, 128 registers/thread, and approximately 232 KB dynamic shared memory). The no-LSE specialization had substantially fewer eligible/issued warps and more samples around the pipeline synchronization path. This looks more like a compiler scheduling/code-generation cliff than an inherent benefit from calculating LSE.

I have only established this behavior on B200; this report does not claim the same result on GB300.

## Minimal reproducer

Run with:

```bash
python repro_fa4_lse_b200.py /path/to/flash-attention
```

```python
"""Reproduce the FA4 D64 B200 performance inversion caused by returning LSE."""

import argparse
import importlib
import importlib.metadata
import statistics
import subprocess
import sys
import time
from pathlib import Path
from types import ModuleType
[...]

PR #2809, opened 2026-08-19
[CuTe, SM100] writeback LSE to shared memory when `return_lse=False` to avoid compiler-induced 14% perf diff

ajuneja23:
Addresses https://github.com/Dao-AILab/flash-attention/issues/2729

@drisspg observed toggling the `return_lse` flag to True to write back log-sum-exp to gmem resulted in 19% speedup on B200 at size `(B, S, H, D) = (2, 16384, 16, 64)`. I was able to repro the performance gap (14% though, not 19%) for these sizes using his script. this table is from running off main on a single B200.

| Shape | return_lse=False | return_lse=True | Speedup |
| --- | ---: | ---: | ---: |
| B2 S2048 H16 D64 | 0.0613 ms | 0.0492 ms | 1.245× |
| B2 S8192 H16 D64 | 0.7624 ms | 0.6420 ms | 1.188× |
| B2 S16384 H16 D64 | 3.0153 ms | 2.6366 ms | 1.144× |

This PR updates the output correction loop so even if `mLSE` is none we will still construct the log-sum-exp per-query and at least writeback to the shared memory state storing row-sum and row-max and this seems to resolve the compiler cliff.

Benchmarks on the above sizes from my branch on a B200.

| Shape | return_lse=False | return_lse=True | Speedup |
| --- | ---: | ---: | ---: |
| B2 S2048 H16 D64 | 0.0492 ms | 0.0492 ms | 1.000× |
| B2 S8192 H16 D64 | 0.6211 ms | 0.6353 ms | 0.978× |
| B2 S16384 H16 D64 | 2.5670 ms | 2.5687 ms | 0.999× |

and the perf gap seems to be closed with this change.

For the `return_lse=True` path this should have no behavior change and we can confirm this by seeing the compiled ptx is byte identical for `B2 S16384 H16 D64` and `return_lse=True` by setting the `CUTE_DSL_KEEP_PTX` environment variable and inspecting the dumped ptx and compare the dumped ptx from this branch and from main.

ptx when running from main with `B2 S16384 H16 D64` with `return_lse=True`
https://gist.github.com/ajuneja23/0a7b2acda995734db3e367a332556ae8

ptx when running from my branch with `B2 S16384 H16 D64` with `return_lse=True`
https://gist.github.com/ajuneja23/25d5535ccf4c9719ee5ec974a521a569

### DIFF
--- a/flash_attn/cute/block_sparse_utils.py
+++ b/flash_attn/cute/block_sparse_utils.py
@@ -883,9 +883,7 @@ def handle_block_sparse_empty_tile_correction_sm100(
 
     for stage in cutlass.range_constexpr(q_stage):
         row_sum_value = Float32(1.0)
-        row_max_value = (
-            -Float32.inf if const_expr(mLSE is not None or learnable_sink is not None) else None
-        )
+        row_max_value = -Float32.inf
         if const_expr(learnable_sink is not None):
             sink_val = -Float32.inf
             if const_expr(not pack_gqa):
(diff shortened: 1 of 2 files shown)


ajuneja23:
@drisspg @tridao Would you be able to take a look at this PR when you get a chance? Thanks!🙏

PR #2768, opened 2026-08-07
Add an Attention Sink to Flash MLA for SM100

floatingtrees:
Currently, attention sinks are not supported for Flash MLA. This PR adds support for a learnable attention sink and associated tests. Please let me know if there's anything I should add or change.

PR #2816, opened 2026-08-24
[CuTe, SM100] Sparse MLA bwd: in-kernel recompute-P + token-chunked backward

abcdabcd987:
Two composable, opt-in memory features for the sparse top-k MLA (DSA) path. Kwargs-only API; the default path is untouched (same binaries, speed, and memory — sentinel tests pass with and without the flags).

## `gather_bwd_recompute_p=True`

The forward saves only `out` + `lse` — no `p`/`row_max` at all (**508 MiB/1k tokens → 0** of saved activations; train forward 4.43 → 3.77 ms at 8k since the `p` store leaves the softmax critical path; the training forward reuses the inference binary). The backward main kernel reconstructs `P = exp2(scale·log2e·S − lse·log2e)` in-kernel with an extra Sᵀ UMMA over the already-gathered V (+ gathered K_rope for the rope chunk), masking invalid top-k slots on the exponent via the gather bitmask; smem stays at the exact 232448 B SM100 cap (dS merges into sP, sScaleP→sLse, stationary-QvB role swap).

- Forward `out`/`lse`: **bitwise-identical** to the default path.
- Grads: not bitwise (one fewer rounding in P) and measure **more** accurate vs an fp32 reference.
- `retain_graph` is safe: nothing saved is consumed.
- `recompute_P` requires the KV bitmask (asserted — it is the only masking in this mode).

## `gather_bwd_token_chunk=N`

The backward loops its three kernels over token chunks, bounding the p-sized dS transient (~520 KiB/token at H=128, W=2048) by the chunk instead of the sequence. With causal masking the chunked main kernel shrinks its K extent to `seqlen_k − seqlen_q + tok1` per chunk (non-varlen: sliced views; varlen: clamped `cu_seqlens_k` ends) so the recomputed bitmask reproduces the forward's key limit exactly; fully-masked chunks (`k_end ≤ 0`) skip all three kernels and zero their dq/dqv directly. Requires varlen or batch 1; the kwarg is validated at the entry points (positive int; sparse-MLA calls only).

- Chunked dq/dqv: **bitwise-identical** to unchunked (including adversarial non-causal indices under `causal=True`, varlen docs split across chunk boundaries, shared-kv, and rectangular `seqlen_q > seqlen_k`).
- dk/dv: differ only by fp32-atomic accumulation order (≤ the unchunked kernel's own run-to-run noise).

## Results (GB200, B=1, W=2048, bf16, square T=S, train step, median time / peak memory)

All columns H=128 except tilelang finetune, whose backward exceeds SM100 smem at H=128 and runs at H=64 (**half** the head-FLOPs).

| T=S  | FA4 main (H=128)         | tilelang dsv32 (H=128)     | tilelang finetune (H=64)   | this PR (`recompute_p` + `chunk=4096`) | vs main              |
|------|--------------------------|----------------------------|----------------------------|----------------------------------------|----------------------|
| 4k   | 12.75 ms / 5.14 GiB      | 64.08 ms / 1.08 GiB        | 70.04 ms / 0.55 GiB        | 13.44 ms / 3.08 GiB                    | +5.4% time, −40% mem |
| 8k   | 25.90 ms / 10.28 GiB     | 105.50 ms / 2.16 GiB       | 137.29 ms / 1.09 GiB       | 27.46 ms / 4.15 GiB                    | +6.0%, −60%          |
| 16k  | 53.61 ms / 20.55 GiB     | 188.53 ms / 4.32 GiB       | 272.47 ms / 2.19 GiB       | 56.40 ms / 6.31 GiB                    | +5.2%, −69%          |
| 32k  | 115.58 ms / 41.10 GiB    | 368.55 ms / 8.64 GiB       | 543.30 ms / 4.37 GiB       | 119.68 ms / 10.62 GiB                  | +3.5%, −74%          |
| 64k  | 260.62 ms / 82.20 GiB    | 730.25 ms / 17.27 GiB      | 1094.00 ms / 8.74 GiB      | 262.75 ms / 19.23 GiB                  | +0.8%, −77%          |
| 128k | OOM (186 GiB)            | 1480.45 ms / 34.55 GiB     | 2213.11 ms / 17.48 GiB     | 575.43 ms / 36.47 GiB                  | —                    |

tilelang reference: `tilelang` @ 0.1.13, same machine and methodology, `-1` index tails remapped to 0 (its gathers/scatters are unguarded; identical traffic). dsv32 = `examples/deepseek_v32/sparse_mla_{fwd,bwd}.py` with the head-split backward — same head-FLOPs, the apples-to-apples column: this PR is 4.8× (4k) to 2.6× (128k) faster. finetune = `examples/dsa_sparse_finetune/` at H=64: even at half the head-FLOPs it is ~4× slower than this PR at H=128. Before this PR the tilelang kernels used ~5× less memory than FA4 while being 2.6–5× slower; after it, FA4 matches their memory scaling (285–300 MiB/1k tokens vs dsv32's 270) at the same speed advantage — the remaining ~2 GiB gap at 64k is the chunk-sized dS transient. Full methodology, smem map, and design notes in `AI/SPARSE_MLA_RECOMPUTE_P.md`.

## Robustness hardening included in this PR

- The sV gather stages are zero-filled once at kernel launch: masked top-k slots skip their `cp.async`, so launch-time smem garbage could otherwise reach the dP GEMM as `0 × NaN` (a pre-existing load-p hazard; recompute mode additionally hardens dS against non-finite dP).
- The dV epilogue scatter is bounds-guarded (`0 ≤ idx < seqlen_k`), so chunked-causal runs never write past the sliced dv view (and skip the wasted zero atomics).
- `pipeline_bitmask` gets its `producer_tail` (same invariant the forward documents); the QvB/dO/dOt TMA tx-count invariants are asserted as actually paired; all warps (including the mma warp) use the same static top-k group count.
- One `SharedStorage` struct for all three compile modes via zero-size fields — per-mode layouts byte-identical (verified at 232448 B each).
- The bwd-preprocess compile key distinguishes `lse_log2` presence, so cached binaries can't be reused across mismatched signatures.

## Tests

`tests/cute/test_flash_attn.py -k mla_sparse` (34 cases): recompute-P vs fp32 reference + bitwise-forward checks; chunked-vs-unchunked (recompute and load-p, non-varlen and varlen with docs split across chunks, for both the rope and shared-KV kernel specializations, checked against per-doc fp32 references); fully-masked rows (all `-1` slots: `out = 0`, `lse = −inf`, NaN-free zero grads, unaffected rows bitwise); rectangular `seqlen_q > seqlen_k` chunking (exercises the skip-chunk path); learnable sink + sparse gather in every mode (out/lse/dsink bitwise-identical across default, recompute-P and chunked; all grads incl. dsink vs the fp32 reference — the sink lives inside `lse`, and dsink depends only on the full-size `dpsum`/`lse`, so neither feature needs sink-specific handling); plus the pre-existing sentinel suites with and without the new flags.


abcdabcd987:
@jayhshah Would you mind reviewing this one?

jayhshah:
Yes, aiming for this weekend -- thanks for the great contribution!

jayhshah:
@abcdabcd987  Sorry for the long delay, I reviewed the PR and it looks good to merge after rebasing. I had Claude write something about this but the rebase should be straightforward:

> `main` picked up the attention-sink change (#2768), which made `_flash_attn_bwd_sparse_mla` return a fifth value (`dsink`) and rewrote the sparse-branch `return` tuples in `FlashAttnFunc.backward` / `FlashAttnVarlenFunc.backward` with fixed trailing-`None` counts. This PR adds two forward arguments (`gather_bwd_recompute_p`, `gather_bwd_token_chunk`) to both Functions, so those counts come up two short. PyTorch tolerates *extra* trailing `None`s but not missing ones, so the merged code raises `returned an incorrect number of gradients` on the first sparse backward.
> 
> **Fix after rebasing onto `main`:**
> 
> 1. Resolve the conflict at the `_flash_attn_bwd_sparse_mla` signature by keeping both sides — the two new kwargs from this PR and `main`'s 5-tuple return annotation:
>    ```
>        recompute_p: bool = False,
>        token_chunk: Optional[int] = None,
>    ) -> Tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor, Optional[torch.Tensor]]:
> 
> 2. In FlashAttnFunc.backward, sparse branch (if qv is not None:), change 12 → 14:
> 
>     ``` 
>     if ctx.shared_kv:
>         return dqv, dv, None, None, None, None, None, None, dsink, *((None,) * 14)
>     else:
>         return dq, dk, dv, dqv, None, None, None, None, dsink, *((None,) * 14)
> 
> 3. In FlashAttnVarlenFunc.backward, sparse branch, change 14 → 16:
> 
>     ```
>     if ctx.shared_kv:
>         return dqv, dv, None, None, *((None,) * 12), dsink, *((None,) * 16)
>     else:
>         return dq, dk, dv, dqv, *((None,) * 12), dsink, *((None,) * 16)
> 
> Everything else merges cleanly and stays consistent: the _bwd_preprocess positional call lines up after main's appended hdim_multiple_of, the sparse compile key / tensor tuple combine both edits, the chunk loop hands the full-size dpsum and lse to main's _bwd_dsink_reduce, and the kernel file only had DSL API renames on main.
> 
> Verified on the resolved merge (GB300): the 12 new tests in this PR pass, the 18 sparse-MLA sentinel + MLA sink tests pass, and an extra probe combining learnable_sink with gather_bwd_recompute_p and gather_bwd_token_chunk (shared-KV and rope forms, causal and non-causal) gives bitwise-identical forward output and dsink across all modes, bitwise chunked-vs-unchunked dq/dqv, and grads within the standard tolerance of the fp32 reference. Recompute-P is compatible with the sink for free since the sink term lives inside lse, and dsink depends only on dpsum and lse, which the chunked path keeps full-size.
> 
> Would be worth adding shared_kv as a parameter to test_flash_attn_mla_sparse_bwd_token_chunk_varlen so the unmerged-dS recompute path gets varlen coverage, and a sparse-gather + sink backward test since main currently has none.

abcdabcd987:
@jayhshah Thanks for the review! Rebased onto `main` (0dc2cb4) and pushed.

**Rebase resolution** — exactly as you described:
- `_flash_attn_bwd_sparse_mla` keeps both the two new kwargs and `main`'s 5-tuple (`dsink`) return.
- Sparse-branch trailing `None`s bumped 12 → 14 in `FlashAttnFunc.backward` and 14 → 16 in `FlashAttnVarlenFunc.backward` (verified by counting forward inputs: 23 and 33 respectively; every backward `return` now matches).
- `flash_bwd_mla_sm100.py` only picked up `main`'s DSL renames (`cute.nvgpu.OperandMajorMode`, `make_trivial_tiled_mma(dtype_a, dtype_b, ...)`); the chunk loop hands the full-size `dpsum`/`lse` to `_bwd_dsink_reduce` after the loop.

**Tests added** (second commit, 4bb5062, so the delta since your review is easy to see; happy to squash if you prefer):
- `test_flash_attn_mla_sparse_bwd_token_chunk_varlen` now parametrizes `shared_kv`, so the no-rope recompute-P layout (dS in its own buffer rather than merged into sP) gets varlen chunked coverage.
- New `test_flash_attn_mla_sparse_bwd_learnable_sink` (varlen × shared_kv × causal): sparse gather with `-1` sentinels + learnable sink, run in default / `recompute_p` / `recompute_p`+`token_chunk` / load-p+`token_chunk`. Asserts `out`/`lse`/`dsink` bitwise-identical across the four modes, chunked `dq`/`dqv` bitwise vs unchunked, and every grad including `dsink` within the standard tolerance of the fp32 reference. (Same result as your probe: recompute-P is sink-agnostic because the sink lives inside `lse`, and `dsink` only depends on the full-size `dpsum`/`lse`. Also added a short note on this to `AI/SPARSE_MLA_RECOMPUTE_P.md`.)

**Validation on the rebased branch** (GB200, CUTLASS DSL 4.7.1): the 30 existing `mla_sparse`/`mla_sink` cases, the 12 new/modified cases (real and `FLASH_ATTENTION_FAKE_TENSOR=1` compile-only mode), and a 326-case regression subset (`test_flash_attn_mla_absorbed[_varlen]` at several shapes including `kv_sparsity` + `has_learnable_sink`, plus the dense `learnable_sink_backward` tests) all pass.

jayhshah (approved):

Issue #2456, opened 2026-04-13
[RFC] FA4 — head_dim=256 & head_dim=512


Johnsonms:
## Overview

This document tracks the roadmap for large head dimension support (`head_dim=256`,`head_dim=512`) in FA4 (the CuTe DSL-based implementation in `flash_attn/cute/`). The work targets Blackwell (SM100/SM110) GPUs using 2CTA instructions and is being developed alongside the main FA4 branch.

Current tracking PR:
  - [Dao-AILab/flash-attention#2412](https://github.com/Dao-AILab/flash-attention/pull/2412) —            
  \[FA4\]\[CuTe DSL\] Add head\_dim=256 support (forward + backward, SM100)                               
  - [Dao-AILab/flash-attention#2454](https://github.com/Dao-AILab/flash-attention/pull/2454) —            
  \[FA4\]\[CuTe DSL\] Add head\_dim=256 support with exp2 FMA emulation optimization (forward + backward, 
  SM100) 

---

## Status Legend

| Symbol | Meaning |
|--------|---------|
| ✅ | Done / merged |
| 🚧 | In code review |
| 🔨 | Code complete, pending PR |
| 📋 | Planned |
| 💡 | Exploratory |

---

## Phase 1 — Core head_dim=256 (SM100 Blackwell)

> All items in this phase are part of PR #2412. The 🔨 follow-up items will be
> submitted as separate PRs roughly half a day after #2412 merges.

| Feature | Status | Notes |
|---------|--------|-------|
| Forward kernel (2CTA, SM100) | 🚧 | In code review — PR #2412 |
| Backward kernel (2CTA, SM100) | 🚧 | In code review — PR #2412 |
| Interface integration (`flash_attn_func`) | 🚧 | Auto-dispatch on SM100, PR #2412 |
| exp2 emulation optimization | 🚧 | ~2–5% FWD speedup at long seqlens, PR #2454 |
| Paged KV table support (inference) | 🔨 | Code complete, accuracy/perf verified in deployment |
| `pack_gqa` support | 🔨 | Code complete |
| FP8 input (FP8 tensor core MMA) | 🔨 | Implementation ready, perf to be improved |

---

## Phase 2 — Optimization & Coverage (Near Term)

| Feature | Status | Notes |
|---------|--------|-------|
| Short sequence performance (vs TRT-LLM kernels) | 📋 | Benchmark-driven tuning |
| Persistent kernel | 📋 | |
[...]

PR #2843, opened 2026-09-02
[CuTe, SM103] Add BF16 head-dim 512 paged SplitKV decode

kzos:
## Summary

This PR adds a narrow FA4 CuTeDSL decode adapter for SM103 with the following
contract:

- BF16 Q/K/V and output;
- Q1, 32 query heads, 4 KV heads, head dimension 512;
- packed HND page-64 KV cache;
- causal full attention;
- up to 32 KV splits.

The producer partitions the KV range across CTAs and writes FP32 output/LSE
partials. A warp-cooperative combine kernel performs the stable log-sum-exp
reduction and writes BF16 output. The adapter validates device, dtype, shape,
stride, page-table, and workspace contracts before compilation. It caches the
compiled CuTe specialization and rejects cold compilation during CUDA Graph
capture.

This does not change the existing public `flash_attn_func` dispatch and does
not claim support for the class's internal page-128, NHD, or dense branches.
The intended first consumer is an inference integration that imports the
narrow adapter directly and retains its existing fallback for all unmatched
calls.

Related: #2456.

## Correctness and graph validation

Tested on one NVIDIA B300 SXM6 AC, compute capability 10.3, against commit
`792115567cf00f3b96d3697954899f9e63996702` on main
`ce088ab9ce0fc0434dcd8afa0a791da9fcc3a820`.

```text
5 passed, 415818 deselected, 40 warnings in 35.32s
```

The focused cells cover K11, K129, K1024, K4096, and heterogeneous B2
sequence lengths `(11, 129)`. A separate regression matrix passed B1/B4/B8
ragged eager execution and CUDA Graph replay with K mutations from 63 through
8192. The modules also import successfully from
`vllm.vllm_flash_attn.cute.*`, validating the relative combine import used by
vendoring integrations.

Standard focused command:

```bash
pytest -vv -x tests/cute/test_flash_attn.py \
  -k sm103_hd512_splitkv_packed_hnd_adapter
```

The recorded archive run used a small package bootstrap because the archive
did not contain the separately built FA2 `flash_attn_2_cuda` extension that
the repository's top-level initializer imports. The bootstrap changed only
package loading; it passed the same test file and selection shown above.

## Performance

Same-input CUDA Graph replay, BF16, B1/Q1, page 64, three rounds of 500
replays; medians are microseconds per attention call:

| KV length | This PR | FlashInfer TRTLLM-gen | Speedup |
| ---: | ---: | ---: | ---: |
| 65 | 8.202 | 8.200 | 1.000x |
| 512 | 8.201 | 12.290 | 1.499x |
| 1024 | 9.599 | 14.346 | 1.495x |
| 2048 | 10.243 | 14.337 | 1.400x |
| 4096 | 10.260 | 14.350 | 1.399x |
| 8192 | 12.831 | 16.388 | 1.277x |

These numbers measure only the maintained attention-call contract; they are
not whole-model tokens-per-second claims.

## Scope and follow-up

- The adapter is intentionally strict and fail closed.
- General D512 prefill, NHD, page-128, other head counts, and other dtypes are
  outside this PR.
- The consumer-side dispatch and fallback integration will be submitted
  separately after this kernel is available in its vendored dependency. That
  integration is tracked in
  [vllm-project/vllm#55017](https://github.com/vllm-project/vllm/issues/55017).

### DIFF
--- a/flash_attn/cute/sm103_hd512_splitkv_combine.py
+++ b/flash_attn/cute/sm103_hd512_splitkv_combine.py
@@ -1,0 +1,234 @@
+# Copyright (c) 2025 - 2026 NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: BSD-3-Clause
+
+"""Specialized combine for SM103 head-dimension-512 SplitKV decode."""
+
+import math
+from functools import partial
+
+import cuda.bindings.driver as cuda
+import torch
+
+import cutlass
+import cutlass.cute as cute
+import cutlass.torch as cutlass_torch
+from cutlass.cute.runtime import from_dlpack
+from cutlass.cute.typing import BFloat16, Float32, Int32
+
+
+log2_e = math.log2(math.e)
+exp2 = partial(cute.math.exp2, fastmath=True)
+
+
+class BlackwellHd512SplitKVCombine:
+    """Warp-cooperative combine for normalized FP32 SplitKV partials.
+
+    This specialization consumes a fixed capacity of 32 splits for Q32/D512.
+    The per-request ``mNumSplitsDynamic`` value may be any value in [1, 32].
+    """
+
+    def __init__(self, head_dim: int = 512, num_heads: int = 32):
+        if head_dim != 512 or num_heads != 32:
+            raise ValueError("SM103 hd512 combine requires Q32/D512")
+        self.head_dim = head_dim
+        self.num_heads = num_heads
+        self.dimensions_per_cta = 128
+        self.num_threads = 128
+        self.max_splits = 32
+
+    @cute.jit
+    def __call__(
+        self,
+        mOPartial: cute.Tensor,
+        mLSEPartial: cute.Tensor,
+        mO: cute.Tensor,
+        mNumSplitsDynamic: cute.Tensor,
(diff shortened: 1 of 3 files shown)


kzos:
Hi @Johnsonms, following up on the large-head-dimension roadmap in #2456: would this narrow BF16 Q1/D512 paged-decode adapter be a suitable addition to FA4? It leaves the existing public dispatch unchanged, and the B300 correctness and CUDA Graph validation are documented in the PR. Would you be willing to review the scope or point me to the appropriate reviewer? Thanks.

kzos:
Hi @Johnsonms, following up on the September 9 scope/review request. Is this narrow SM103 BF16 Q1/D512 paged-decode adapter a suitable addition to FA4, or would another reviewer be better placed to assess it? The PR records B300 correctness and CUDA Graph validation, leaves public dispatch unchanged, and keeps the vLLM consumer integration separate until the dependency is available. Happy to adjust the scope based on feedback. Thanks!

Johnsonms:
@kzos just coming back from medical leave and catching up, will update by end of this week

kzos:
Thanks for the update, @Johnsonms. I hope you're doing well. Please take the time you need to settle back in—there's no pressure on the timeline for this PR. Take care!

PR #2851, opened 2026-09-03
Fix FA2 varlen SplitKV early-exit LSE offset using get_lse_tile and f…

jagane-memwize:
…ix get_lse_tile


jagane-memwize:
This is a cleaner follow up to PR#2573. I noticed that get_lse_tile code did not match the comment. The code ignored params.unpadded_lse.

This PR makes it possible to undo a transpose performed in vLLM's fork of flash-attention in order to compensate for this issue

PR #2872, opened 2026-09-09
Fix Windows builds with CUDA 13 and PyTorch 2.15

### LINKED ISSUE #2832: Windows build fails with CUDA 13.4: missing /Zc:preprocessor, /std:c++17 pinned (plus no sm_121 gencode)
Hi! I tried building 2.8.4 from source on Windows and hit some failures, all in setup.py's Windows branch. They're build configuration only - no kernel, CUTLASS or host-code changes required. I have them all fixed locally and can open a PR. The README says Windows "might work starting v2.3.2 ... but Windows compilation still requires more testing" and invites people to open an issue, so here it is.

Environment: Windows 11 ARM64, MSVC 14.51 (ARM64 host and target), CUDA 13.4, torch 2.15, CPython 3.13 and 3.14, sm_121 (GB10).

The three things that go wrong, all in build configuration rather than the kernels:

**1. There's no `sm_121` gencode.** `cuda_archs()` defaults to `"80;90;100;110;120"` and
`add_cuda_gencodes()` has no `121` branch, so nothing gets emitted for this GPU.

**2. CUDA 13.x's CCCL headers refuse MSVC's traditional preprocessor.** Every `.cu` stops
here:

```
cccl/cuda/std/__cccl/preprocessor.h(20): fatal error C1189: #error: MSVC/cl.exe with
traditional preprocessor is used ... pass `/Zc:preprocessor` to cl.exe
```

The Windows branch passes `/Zc:__cplusplus` already, just not `/Zc:preprocessor`.

**3. `/std:c++17` is pinned, and torch 2.15's headers need C++20:**

```
c10/util/StringUtil.h(171): error C7555: use of designated initializers requires at
  least '/std:c++20'
c10/util/intrusive_ptr.h(775): error C2039: 'strong_ordering': is not a member of 'std'
```

(There's also `M_LOG2E` in `src/softmax.h(76)` coming up undefined, which MSVC only
provides with `_USE_MATH_DEFINES`.)

One thing that might matter more than the ARM64 angle: only (1) is actually
architecture-specific. (2) and (3) look like they'd hit Windows x64 with the same CUDA and
torch versions just as hard.

I'll open a PR with these unless you see a problem with the approach. The build passes correctness checks against an FP32 SDPA reference, dense and varlen, forward and backward.

tuananhlfc:
## Summary

Fix the Windows CUDA build failures reported in #2832.

- Add sm_121 to the default CUDA architectures.
- Generate native sm_121 code with compute_121f on CUDA 12.9+.
- Emit forward-compatible PTX only for architectures supported by the active toolkit.
- Use C++20 with PyTorch 2.15+, while retaining C++17 for older releases.
- Enable MSVC’s conforming preprocessor with /Zc:preprocessor.
- Define _USE_MATH_DEFINES so M_LOG2E is available under MSVC.

These are build-configuration changes only; no attention kernel logic is modified.

## Motivation

CUDA 13.x CCCL headers reject MSVC’s traditional preprocessor. Additionally, PyTorch 2.15 headers use C++20 features, including designated initializers and std::strong_ordering.

On an SM 12.1 GPU, the existing architecture defaults also produce no native sm_121 cubin. Finally, MSVC does not expose M_LOG2E unless _USE_MATH_DEFINES is set.

Before this change, these resulted in:

  - CCCL C1189, requesting /Zc:preprocessor
  - C7555 for C++20 designated initializers
  - C2039 for std::strong_ordering
  - Undefined M_LOG2E
  - No native sm_121 code

## Compatibility

PyTorch releases before 2.15 continue to use C++17. The gencode helper now tracks architectures supported by the detected CUDA version, preventing unsupported architectures from being selected for the forward-compatible PTX target.

## Validation

  Tested on:

  - Windows 11 ARM64
  - MSVC 14.51
  - CUDA 13.4.46
  - PyTorch 2.15 Windows ARM64 headers
  - NVIDIA GPU with compute capability 12.1

Validation performed:

  - Reproduced all reported failures using the previous flags.
  - Compiled CCCL, PyTorch 2.15, and <cmath> headers together using the new flags.
  - Verified native sm_121 output and compute-121 PTX with cuobjdump.
  - Verified host compilation with the new MSVC flags.
  - Checked gencode output for CUDA 11.7, 12.8, 12.9, and 13.4.
  - Ran python -m py_compile setup.py and git diff --check.

  Fixes #2832


setup.py
```diff
@@ -309,32 +320,43 @@ def get_ck_tile_bfloat16_supported_modes(ck_dir):
     if FORCE_CXX11_ABI:
         torch._C._GLIBCXX_USE_CXX11_ABI = True
 
+    cxx_standard = "c++20" if (TORCH_MAJOR, TORCH_MINOR) >= (2, 15) else "c++17"
```
Johnsonms: I think it is supposed to >= (2, 14) as this https://github.com/pytorch/pytorch/pull/178150/changes

tuananhlfc:
@Johnsonms FYI, I just rebased the PR as you requested in #2899 . Is there any blocker for this PR?

Johnsonms:
Could you please paste some screenshot for the setup and verification? as I have no Windows-GPU node to verify

PR #2807, opened 2026-08-18
[CuTe, SM100, HD256] optimize SM100 HD256 varlen scheduling and epilogues

catwinee:
## Summary

- Bound dedicated SM100/SM110 HD256 varlen forward, dQ, and dK/dV grids by runtime maximum sequence lengths instead of packed total-token extents.
- Preserve packed output alignment for vector stores.
- Use TMA stores for complete 128-row dQ tiles while retaining predicate stores for ragged tails.
- Add targeted MQA correctness coverage for complete and ragged tiles under causal and noncausal masking.

This closes #2521. Thanks to @horakka5 for the original root-cause analysis and reproducer.

## Why

The dedicated HD256 kernels used the packed `total_q` / `total_k` extents to size each batch's grid. For a packed input with `B` equal-length sequences this launches approximately `B` times more M tiles than necessary, with most CTAs exiting after reading `cu_seqlens`.

After fixing the grid, packed-varlen epilogues become visible in short-sequence profiles:

- Forward used `domain_offset` for the runtime packed-row offset, which discarded the alignment proof needed by `autovec_copy`.
- Backward constructed a dQ TMA destination only for fixed-length inputs. Every varlen dQ tile therefore used the predicate global-store path, including complete tiles that cannot cross a sequence boundary.

The change keeps the physical packed tensor extents unchanged. Runtime maximum sequence lengths are used only for grid sizing. Within a sequence, complete dQ tiles use TMA; the final partial tile keeps the bounds-checked path.

## Benchmark protocol

- GPU: NVIDIA GB200, SM100
- PyTorch: `2.11.0+cu130`
- CUDA: 13.0
- CUTLASS DSL: 4.5.2
- dtype: BF16
- packed-varlen layout: `Q=[B*S,Hq,256]`, `K/V=[B*S,1,256]`
- `Hkv=1`, dropout disabled, softcap disabled
- seed: `20260814`
- `CUDA_DEVICE_MAX_CONNECTIONS=16`
- 20 warmups per phase
- 5 retained trials; reported latency is the median
- 50 timed iterations per trial for `S <= 8K`, otherwise 20
- FA4 HD256 backward uses `deterministic=False` because the current upstream kernel rejects deterministic mode

The baseline is upstream main before this PR. Commit `0251105` only changes dependency requirements; its HD256 kernels are identical to the measured `145b101` baseline.

MFU uses the repository's historical GB200 roofline:

- compute peak: 2.5 PFLOP/s
- HBM peak: 8 TB/s
[...]

PR #2877, opened 2026-09-10
[CuTe, SM100] Add symmetric D512 forward and backward

lingkong-q:
Add symmetric head dimension 512 (`d_qk = d_v = 512`) to the SM100 CuTe attention path, including FP16/BF16 forward and backward.

### Implementation and scope

- Fused two-CTA dense forward shares QK and online softmax across all 512 output columns, reusing the SM100 MLA pipeline. Generic forward uses two V256 slices for varlen, paged KV, FP8, local masks, softcap and SplitKV; each slice uses the complete D512 QK dot product.
- Separate TCGEN05 dQ/dK/dV kernels keep the working set within shared-memory/TMEM limits. Accumulation and grouped-query partial reduction use FP32, with head grouping and long noncausal tile ordering for reuse.
- Compact layouts keep batch/token extents dynamic, with runtime maximum lengths and scheduling-aware cache keys. Noncompact layouts retain static specialization. Forward, backward and GQA reduction have disk-cache reload regressions.
- Large GQA reduction counts and indices use Int64 from the first multiplication, avoiding overflow at 2^31 output elements.
- Signed window bounds retain their literal meaning. The latest fix also initializes gradients when a negative bound can exclude an entire output tile: for square sequences, those tiles previously skipped their stores and could leave stale gradients. Added tests use nonzero supplied output buffers so allocator reuse cannot hide missing writes.

FP16/BF16: dense/varlen forward and backward, MHA/GQA/MQA, causal/local masks, softcap, LSE gradients and training with SplitKV forward. FP8 E4M3/E5M2 and paged KV are **forward only**. Deterministic backward, learnable sinks, custom score/mask callbacks, block sparsity and sparse gather/scheduler metadata are outside this implementation. Callers must translate external `-1`-means-unbounded conventions to FA4's `None` convention. Validation is on B200; full-graph `torch.compile` integration is not claimed.

Latest follow-up: `71ff336`, retaining the dynamic-cache contribution in `313d646` and the subsequent fixes in `f502aea` and `bed49db`. The diff contains implementation, regression tests and necessary CI wiring; no reports or binary artifacts.

### Validation at `71ff336`

B200, PyTorch 2.11.0+cu130, CUTLASS DSL 4.6.2, quack-kernels 0.6.4:

- A separately installed wheel passed **402 tests: all 394 D512 regressions plus 8 upstream CI-selected smoke cases**, with no skips. All **54 Python modules** match committed source byte-for-byte; interpreter and imported module paths were checked in the controller and workers.
- **56 signed-window cases** cover FP16/BF16, square/rectangular sequences, MHA/GQA, entirely empty and partially empty output tiles, autograd and supplied nonzero gradient buffers. The previous installed wheel failed three sentinel reproductions; the fix also passed 24 independent cases including ordinary-window controls.
- The suite includes cache reuse across batch/token changes, disk reload, output layouts, FP8/paged/SplitKV, independently packed Q/K plus used lengths and LSE gradients, and four checkpoint/autocast/AdamW regressions. The latter exercise a small attention block, not a complete training stack.
- All four large-index reduction cases executed at or above 2^31 output elements, both across and within batches. These require about 40 GiB of buffers and explicitly skip below 48 GiB free memory.
- Compute Sanitizer **memcheck passed four square signed-window MHA/GQA cases with zero errors**. This is targeted sanitizer validation, not a full-suite sanitizer claim.
- The CI driver now selects the D512 file. Local validation used the workflow's filter and the driver's two-stage plan: fake pass **8 passed / 394 skipped**, execution pass **402 passed**. D512 tests require real values and intentionally skip the fake pass. Local execution used four workers; the workflow's default worker count and triggers are unchanged. This is local validation, not a claim that the upstream GitHub Actions workflow has run.
- Ruff on the added kernels/cache helper and `git diff --check` pass. Existing numerical tolerances were not relaxed; comparisons use FP64 and either fixed tolerances or the upstream input-precision error criterion, with finite-value checks.

```bash
python -m pip install -e 'flash_attn/cute[dev]'
pytest -x tests/cute/test_flash_attn_dim512.py
```

### Performance

Fresh measurements from this source on one B200: BF16, B1, Hq8/Hkv2, fixed sequence/mask, changing only D. D128/D256 are the upstream implementations at base `1bda8f9`. Changing D changes the workload; these are not identical-workload regression ratios.

Three interleaved rounds, four independent inputs rotated in a CUDA Graph, median of round medians; each pattern receives at least 150 ms warmup followed by 10 groups of 10 replays. Forward uses `num_splits=1` and `return_lse=True`; complete backward includes preprocessing and required reduction. Compilation and input creation are excluded. No clock settings were changed. Values are milliseconds:

| S / mask | D128 fwd | D256 fwd | D512 fwd | D128 bwd | D256 bwd | D512 bwd |
| --- | ---: | ---: | ---: | ---: | ---: | ---: |
| 512 / causal | 0.01084 | 0.01441 | 0.01235 | 0.03193 | 0.06617 | 0.07547 |
| 512 / full | 0.01133 | 0.01340 | 0.01298 | 0.03191 | 0.06261 | 0.07533 |
| 2048 / causal | 0.02704 | 0.02928 | 0.05181 | 0.05802 | 0.20128 | 0.22549 |
| 2048 / full | 0.02826 | 0.03288 | 0.07498 | 0.05900 | 0.18427 | 0.35949 |
| 8192 / causal | 0.11830 | 0.25866 | 0.51048 | 0.39751 | 1.03853 | 2.54343 |
| 8192 / full | 0.23055 | 0.43531 | 0.93605 | 0.66360 | 1.78877 | 4.93006 |

The same run interleaved the preceding D512 implementation at `bed49db`: 72 records total (54 current cross-dimension records, 18 preceding-D512 records). Observed complete-backward median differences across the six shapes were **-0.35% to +0.23%**. These measurements are not precise causal estimates of sub-percent overhead. No claim is made that optimization has reached a hardware limit.

The timing harness checks forward against FP64 (all queries through S2048, sampled boundary/interior queries at S8192, against all keys). Backward numerical validation comes from the separate regression suite; timing records are not a full 8K backward numerical proof.

### DIFF
--- a/flash_attn/cute/flash_bwd_sm100_hd512.py
+++ b/flash_attn/cute/flash_bwd_sm100_hd512.py
@@ -3,7 +3,7 @@
 The dedicated D256 backward supplies the producer/consumer design: TMA sends
 transaction completion directly to the leader CTA, MMA releases input buffers,
 and elementwise warps publish the probability or dS operand. DQ/DK compute
-probabilities while dP runs; DV uses a wider sequence tile. Two input stages
+probabilities while dP runs; DV selects its query tile by mask type. Two input stages
 split the 512-wide reduction into 256-wide chunks. FP32 gradient accumulators
 remain in TMEM until the final write, with an FP32 GQA reduction when needed.
 """
@@ -1009,6 +1009,9 @@ class NativeD512Dv(NativeD512DqDk):
         self.maxsq = maxsq
         self.maxsk = maxsk
         self.threads = 256
+        # A smaller query tile reduces resource pressure for noncausal DV.
+        # Causal attention retains the wider tile to amortize traversal costs.
+        self.query_tile = 256 if causal else 128
 
     @cute.jit
     def __call__(
@@ -1036,7 +1039,7 @@ class NativeD512Dv(NativeD512DqDk):
             tcgen05.OperandMajorMode.K,
             Float32,
             tcgen05.CtaGroup.TWO,
-            (128, 256),
+            (128, self.query_tile),
         )
         mg = sm100.make_trivial_tiled_mma(
             self.dtype,
@@ -1046,21 +1049,21 @@ class NativeD512Dv(NativeD512DqDk):
             tcgen05.CtaGroup.TWO,
             (128, 256),
         )
-        la = sm100.make_smem_layout_a(ms, (128, 256, 512), self.dtype, 1)
-        lb = sm100.make_smem_layout_b(ms, (128, 256, 256), self.dtype, 2)
-        lp = sm100.make_smem_layout_a(mg, (128, 256, 256), self.dtype, 1)
-        lbg = sm100.make_smem_layout_b(mg, (128, 256, 256), self.dtype, 1)
+        la = sm100.make_smem_layout_a(ms, (128, self.query_tile, 512), self.dtype, 1)
+        lb = sm100.make_smem_layout_b(ms, (128, self.query_tile, 256), self.dtype, 2)
+        lp = sm100.make_smem_layout_a(mg, (128, 256, self.query_tile), self.dtype, 1)
+        lbg = sm100.make_smem_layout_b(mg, (128, 256, self.query_tile), self.dtype, 1)
         # TMA uses the per-CTA physical operand layouts; MMA coordinates remain cluster-wide.
         ls = cute.composition(
             cute.select(la, mode=[0, 1, 2]), cute.make_layout((64, 512), stride=(1, 64))
         )
         lg = cute.composition(
             cute.select(lbg, mode=[0, 1, 2]),
-            cute.make_layout((128, 256), stride=(1, 128)),
+            cute.make_layout((128, self.query_tile), stride=(1, 128)),
         )
         lbs = cute.composition(
             cute.select(lb, mode=[0, 1, 2]),
-            cute.make_layout((128, 256), stride=(1, 128)),
+            cute.make_layout((self.query_tile // 2, 256), stride=(1, self.query_tile // 2)),
         )
         op = cpasync.CopyBulkTensorTileG2SOp(tcgen05.CtaGroup.ONE)
         qa, qt = cpasync.make_tiled_tma_atom(op, self.view(q), ls, (64, 512))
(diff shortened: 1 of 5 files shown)


lingkong-q:
@drisspg Would you be willing to review this SM100 symmetric-D512 extension, or suggest the appropriate reviewer? I saw that you reviewed the recent SM100 HD256 update in #2807.

The branch now includes upstream `1bda8f9`; the merge conflict is resolved, and the D512 backward MMA construction calls follow upstream's CUTLASS API migration. On the updated head (`0cd061b`), all 307 focused D512 regression tests pass on B200. The PR description separates this source-checkout validation from the earlier wheel, sanitizer, and performance measurements.

There is also downstream integration evidence: the author of NVIDIA/TransformerEngine#3528 reports that, using this PR's earlier `346aa9e` revision with TE's D512 gate lifted, D512 context-parallel tests passed for `p2p`, `all_gather`, and `a2a` after normalizing TE's unbounded-window sentinel for FA4. That report retracts the initial suspicion of a separate D512 defect; the TE-side normalization fix is proposed in NVIDIA/TransformerEngine#3532. This is contributor-reported validation of those configurations, not a claim of TE maintainer approval or validation of the latest merge.

Feedback on the kernel decomposition, dispatch boundaries, and any additional coverage needed for upstream acceptance would be appreciated. Thank you!

dawnop:
Both D512 paths convert their tensors with a plain `from_dlpack`, so every extent is baked into the compiled kernel and the cache key has to carry the full `(shape, stride, dtype)`: `_forward_cache` in `flash_fwd_sm100_hd512.py`, `_native_cache` in `flash_bwd_sm100_hd512.py`, which keys on `max_seqlen` as well. Fixed shapes pay that once. Packed varlen training hands over a new token count on nearly every micro-batch and recompiles each time — about 1.7 s per new forward shape, 3.9 s per new dq/dk/dv trio. It cost 20% of step time in an 8xB200 SFT run, and per-kernel timings show nothing, because a shape that has been seen once runs in 42 ms.

313d646 marks the batch and sequence modes dynamic where the layout is compact, passes `max_seqlen` to the backward as a runtime `Int32` instead of constructor state, and keys both caches on the head shape alone. In the kernel `maxsq` was only used for `tile = ceil_div(maxsq, 128) - 1 - tile`, which takes a dynamic value fine. Non-compact layouts keep the static path, and the GQA reduction judges its own tensors, since the caller's `dk`/`dv` arrive over-strided in `test_large_gradient_stride`.

One fwd+bwd per shape, one B200, bf16, 16q/2kv, d512, causal, varlen (ms; the first call compiles in both):

```
packed tokens   32768    32000    32100    32300   32768 (repeat)
before         7093.5   3931.7   3945.0   3931.6     41.9
after          7298.3     36.7     40.4     49.5     42.1
```

Dense forward only, batch 1-4 by seqlen 1024-32768, eight shapes: 7 compiles before, 1 after.

`tests/cute/test_flash_attn_dim512.py`: 307 passed, warm timings unchanged, gradients at bf16 rounding against an SDPA reference for varlen, dense, and the non-compact fallback.

Caveat: my `nvidia-cutlass-dsl` predates the `make_trivial_tiled_mma` signature this branch moved to, so I ran all of this with the same three files based at 346aa9e; the backward differs from what I tested only by that MMA rename, which does not touch the edited regions.

PR #2879, opened 2026-09-11
Build CUDA extension with C++20 on PyTorch 2.13+

### LINKED ISSUE #2863: CUDA build uses C++17 but PyTorch 2.13+ requires C++20
## Summary

The NVIDIA CUDA build in the current FlashAttention source explicitly passes
`-std=c++17` to both the host compiler and NVCC.

PyTorch 2.13 and newer require C++20 for extensions that include PyTorch/ATen
headers. When attempting to build FlashAttention against PyTorch 2.14 using the
existing C++17 configuration, the build encountered the PyTorch C++ standard
requirement:

```text
#error C++20 or later compatible compiler is required to use ATen
```

I changed the two NVIDIA build flags from C++17 to C++20. After this change,
FlashAttention 2.8.4 built and installed successfully against PyTorch
2.14.0+cu132.

No FlashAttention kernel implementation or mathematical behavior was modified.

## Environment

```text
Operating system: Linux x86_64
Python: 3.13
PyTorch: 2.14.0+cu132
CUDA runtime used by PyTorch: 13.2
CUDA toolkit / NVCC: 13.2
GPU: NVIDIA GeForce RTX 5090
Compute capability: 12.0
Compilation target: SM120
FlashAttention version: 2.8.4 from the current main branch
Build system: Ninja
```

## Current configuration

The NVIDIA build currently contains C++17 flags for both compiler paths:

```python
nvcc_flags = [
    "-O3",
    "-std=c++17",
    ...
]

compiler_c17_flag = [
    "-O3",
    "-std=c++17",
]
```

Because the standard is specified explicitly, PyTorch's extension build
machinery cannot select C++20 automatically.

## Local modification

I changed both C++17 flags:

```diff
 nvcc_flags = [
     "-O3",
-    "-std=c++17",
+    "-std=c++20",
     ...
 ]

-compiler_c17_flag = ["-O3", "-std=c++17"]
+compiler_cxx_flag = ["-O3", "-std=c++20"]
```

For the experiment, the variable name did not need to be changed; only the two
flag values were modified.

## Build procedure

I cloned the repository and initialized the NVIDIA CUTLASS submodule:

```bash
git clone --depth 1 \
    https://github.com/Dao-AILab/flash-attention.git

cd flash-attention

git submodule update \
    --init \
    --depth 1 \
    csrc/cutlass
```

I then replaced the two C++17 flags with C++20 and built using:

```bash
export BUILD_TARGET=cuda
export FLASH_ATTENTION_FORCE_BUILD=TRUE
export FLASH_ATTN_CUDA_ARCHS=120
export MAX_JOBS=6
export NVCC_THREADS=2

python -m pip install \
    . \
    --no-build-isolation \
    --no-deps \
    --force-reinstall \
    --verbose
```

`FLASH_ATTENTION_FORCE_BUILD=TRUE` was used to ensure that the modified source
was compiled instead of downloading a prebuilt wheel.

`--no-deps` was used to prevent pip from replacing the existing PyTorch 2.14
installation.

## C++20 propagation

The generated build commands confirmed that C++20 was used by all relevant
compiler stages.

Host compiler:

```text
c++ ... -O3 -std=c++20 ...
```

CUDA compiler:

```text
nvcc ... -O3 -std=c++20 ...
```

CUDA device compiler:

```text
cicc --c++20 ...
```

SM120 code generation was also correctly selected:

```text
-gencode arch=compute_120f,code=sm_120
-gencode arch=compute_120,code=compute_120
```

## Result

The complete extension compiled and linked successfully:

```text
Building wheel for flash_attn (pyproject.toml): finished with status 'done'
Successfully built flash_attn
Successfully installed flash_attn-2.8.4
```

The generated extension was:

```text
flash_attn_2_cuda.cpython-313-x86_64-linux-gnu.so
```

The generated wheel was:

```text
flash_attn-2.8.4-cp313-cp313-linux_x86_64.whl
```

The extension was linked against the PyTorch 2.14 libraries in the active
environment, including:

```text
libc10
libtorch
libtorch_cpu
libtorch_python
libc10_cuda
libtorch_cuda
libcudart
```

This confirms that changing the host and NVCC flags to C++20 allows the complete
FlashAttention extension to build and install with PyTorch 2.14 and CUDA 13.2.

## Requested change

Could the NVIDIA CUDA build be updated to use C++20 when compiling against
PyTorch 2.13 or newer?

A version-dependent implementation could be:

```python
torch_version = tuple(
    int(component)
    for component in torch.__version__.split("+")[0].split(".")[:2]
)

requires_cxx20 = torch_version >= (2, 13)
cxx_standard = "c++20" if requires_cxx20 else "c++17"

nvcc_flags = [
    "-O3",
    f"-std={cxx_standard}",
    ...
]

compiler_cxx_flags = [
    "-O3",
    f"-std={cxx_standard}",
]
```

It may also be useful to:

1. Rename `compiler_c17_flag` to a version-neutral name.
2. Keep the host and NVCC C++ standards synchronized.
3. Add a source-build CI job for PyTorch 2.13 and 2.14.
4. Test CUDA 13.x and SM120 in that CI matrix when suitable hardware is
   available.

## Scope of this report

This report concerns source-build compatibility only.

It does not claim that changing C++17 to C++20 improves FlashAttention runtime
performance. The CUDA kernel implementation was not changed; only its
compilation standard was updated.

zhang-keliang:
## Summary

The NVIDIA CUDA build hardcodes `-std=c++17` for both the host compiler and NVCC. PyTorch 2.13+ requires **C++20** for any extension that includes ATen headers, so building against torch 2.13/2.14 fails at compile time:

```
#error C++20 or later compatible compiler is required to use ATen
```

Because `setup.py` passes an explicit `-std` flag, PyTorch's `cpp_extension` machinery cannot upgrade the standard for us — the hardcoded C++17 forces the failing standard.

Fixes #2863.

## Change

- Select the C++ standard from the installed torch version: `c++20` for `torch >= 2.13`, `c++17` otherwise. Existing toolchains on older torch are unaffected (no regression).
- Applied to both `nvcc_flags` and the host-compiler flags, including the Windows `/std:` variant, so the two stay in sync.
- Renamed `compiler_c17_flag` → `compiler_cxx_flag` since it is no longer always C++17.

The ROCm/HIP build path already uses C++20, so this only aligns the CUDA path.

No kernel implementation or mathematical behavior is changed — this is a build-configuration fix only.

## Testing

- Verified the version-selection logic maps torch `2.12→c++17`, `2.13→c++20`, `2.14→c++20`, `3.x→c++20`.
- The change is confined to the NVIDIA CUDA branch of `setup.py`; `python -m py_compile setup.py` passes.
- Full end-to-end build against torch 2.14 + CUDA 13.2 / SM120 was reported working by the issue author (#2863); I do not have matching CUDA/GPU hardware to re-run the full compile here, so I validated the setup.py logic rather than the produced binary.

AI assistance was used to prepare this change.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -309,9 +309,16 @@ if not SKIP_CUDA_BUILD and not IS_ROCM:
     if FORCE_CXX11_ABI:
         torch._C._GLIBCXX_USE_CXX11_ABI = True
 
+    # PyTorch 2.13+ requires C++20 for extensions that include ATen headers
+    # (ATen raises "#error C++20 or later ... required"). Because we pass an
+    # explicit -std flag below, PyTorch's build machinery cannot upgrade the
+    # standard for us, so select it from the installed torch version. Older
+    # torch keeps C++17 to avoid requiring a newer toolchain unnecessarily.
+    cxx_standard = "c++20" if (TORCH_MAJOR, TORCH_MINOR) >= (2, 13) else "c++17"
+
     nvcc_flags = [
     "-O3",
-    "-std=c++17",
+    f"-std={cxx_standard}",
     "-U__CUDA_NO_HALF_OPERATORS__",
     "-U__CUDA_NO_HALF_CONVERSIONS__",
     "-U__CUDA_NO_HALF2_OPERATORS__",
@@ -330,11 +337,11 @@ if not SKIP_CUDA_BUILD and not IS_ROCM:
     # "-DFLASHATTENTION_DISABLE_LOCAL",
     ]
 
-    compiler_c17_flag=["-O3", "-std=c++17"]
+    compiler_cxx_flag = ["-O3", f"-std={cxx_standard}"]
     # Add Windows-specific flags
     if sys.platform == "win32" and os.getenv('DISTUTILS_USE_SDK') == '1':
         nvcc_flags.extend(["-Xcompiler", "/Zc:__cplusplus"])
-        compiler_c17_flag=["-O2", "/std:c++17", "/Zc:__cplusplus"]
+        compiler_cxx_flag = ["-O2", f"/std:{cxx_standard}", "/Zc:__cplusplus"]
 
     # Opt-in: disable building dropout and its dependent headers (ATen philox/RNG
     # headers) from the FA2 build. This flag must be shared across both cxx and nvcc
@@ -446,7 +453,7 @@ if not SKIP_CUDA_BUILD and not IS_ROCM:
                 "csrc/flash_attn/src/flash_fwd_split_align_hdim256_bf16_causal_sm80.cu",
             ],
             extra_compile_args={
-                "cxx": compiler_c17_flag + feature_flags,
+                "cxx": compiler_cxx_flag + feature_flags,
                 "nvcc": append_nvcc_threads(nvcc_flags + cc_flag + feature_flags),
             },
             include_dirs=[


setup.py
```diff
@@ -309,9 +309,16 @@ def get_ck_tile_bfloat16_supported_modes(ck_dir):
     if FORCE_CXX11_ABI:
         torch._C._GLIBCXX_USE_CXX11_ABI = True
 
+    # PyTorch 2.13+ requires C++20 for extensions that include ATen headers
+    # (ATen raises "#error C++20 or later ... required"). Because we pass an
+    # explicit -std flag below, PyTorch's build machinery cannot upgrade the
+    # standard for us, so select it from the installed torch version. Older
+    # torch keeps C++17 to avoid requiring a newer toolchain unnecessarily.
+    cxx_standard = "c++20" if (TORCH_MAJOR, TORCH_MINOR) >= (2, 13) else "c++17"
```
Johnsonms: It is supposed to  `(TORCH_MAJOR, TORCH_MINOR) >= (2, 14)` instead of 2.13， right? Please verify that, ref: https://github.com/pytorch/pytorch/pull/178150/changes

zhang-keliang: Confirmed — the C++20 header guard lands in `release/2.14`, not `release/2.13`. Verified directly against pytorch/pytorch sources:

- `release/2.12` `aten/src/ATen/ATen.h`: `__cplusplus < 201703L` (C++17)
- `release/2.13` `aten/src/ATen/ATen.h`: `__cplusplus < 201703L` (C++17)
- `release/2.14` `aten/src/ATen/ATen.h`: `__cplusplus < 202002L` (C++20)

Bumped the gate to `>= (2, 14)` and updated the comment (eabdfc6). Thanks for catching this.

Johnsonms (approved):

PR #2888, opened 2026-09-14
[CuTe, SM90] Fix swapped backward MMA layout and use 2-stage hdim256 config

SuperGoodGame:
## Summary

SM90 backward at head_dim 256 runs at ~300 TFLOP/s while hdim 128 is at parity with the C++ FA3 kernel (#2878). The cause is the hdim 256 config: it is the only SM90 backward config with `num_stages_Q=1`, so the Q/dO TMA loads and the WGMMAs are serialized every m-block. The C++ tile (64x80, 2-stage Q) does not fit next to the 64 KB smem dQaccum staging buffer, but 64x48 with `dKV_swapAB` does.

Using that config for GQA/MQA as well needs one real fix, and it turned up a second one:

1. **Swapped dK/dV postprocess layout.** `FlashAttentionBackwardPostprocess._get_tiled_mma` reversed the whole `(tile / atom_layout)` tiler for `swap_ab`, which asks `make_trivial_tiled_mma` for a WGMMA atom with M = hdim / num_wg. Every SM90 config with `dKV_swapAB` therefore fails in the GQA/MQA postprocess: `expects the M-mode to be 64, but got 192` for hdim in (128, 192] on `main` today (cf. #2087), `got 128` for `dQ_swapAB` above hdim 128. Build the tiled MMA the way the mainloop does, `tiler_mn=(64, tile / atom_layout on N)`; the accumulator register layout is unchanged.
2. **GQA dV accumulator rounding.** `dv_accum` was allocated with a fixed 32-element head-dim rounding while the kernel and the postprocess use `hdim_multiple_of` (64 with `dKV_swapAB`), so head_dim_v in {160, 200, 208, 224} reduced and read past each (batch, kv-head) region and dV came out wrong. Round it like `dk_accum`.
3. **hdim 256 config: 64x48, 2-stage Q, `dKV_swapAB`**, the same for MHA, GQA and MQA (no per-mode fallback). `dKV_swapAB` rounds `tile_hdim` to 64, so head_dim in (192, 256) (200/208/224/240, which currently abort inside the DSL compiler or fail in `get_smem_store_C`) now compiles and is correct. Block-sparse backward keeps the previous 64x64 tile because it derives its KV block size from `n_block_size`.

No new kernel or config parameters; forward and hdim <= 192 configs are untouched. 3 files, +40/-9 including tests.

## Results

[...]

Issue #2878, opened 2026-09-10
[FA4, Hopper, BWD] BWD Speed Regression on H200 when headdim=256


complexfilter:
flash-attn-4    version:      4.0.0b24

Supposely, the runtime of bwd of FA4 should be nearly 500ms given the runtime of headdim=128. Unfortunately, it takes almost twice the runtime. 

```
batch_size: 1,  seq_len: 86000, head_size: 32, dim: 128
runtime of fwd of FA4 at bf16 precision in milliseconds: 88.88246 (88.88246, 88.88246).
---------------------------------------------------------------------
runtime of bwd of FA4 at bf16 precision in milliseconds: 246.29193 (246.29193, 246.29193).
---------------------------------------------------------------------
batch_size: 1,  seq_len: 86000, head_size: 32, dim: 256
runtime of fwd of FA4 at bf16 precision in milliseconds: 166.24957 (166.24957, 166.24957).
---------------------------------------------------------------------
runtime of bwd of FA4 at bf16 precision in milliseconds: 965.84863 (965.84863, 965.84863).
---------------------------------------------------------------------
```

To reproduce, run the following code on H200 under 4.0.0b24:

```
import torch
import triton
from flash_attn.cute.interface import flash_attn_varlen_func as FA4_Varlen

def benchmark(f, job_name:str):
    warmup=20
    rep=100
    ms = triton.testing.do_bench(f, warmup=warmup, rep=rep, quantiles=[0.2, 0.5, 0.8])
    print(f"runtime of {job_name} in milliseconds: {ms[1]:.5f} ({ms[0]:.5f}, {ms[2]:.5f}).")
    return ms[1]

def test_once_varlen(batch_size, seq_len, num_heads, dim):
    print(f"batch_size: {batch_size},  seq_len: {seq_len}, head_size: {num_heads}, dim: {dim}")
    q, k, v = get_tensors(batch_size, seq_len, num_heads, dim)
    device = "cuda"
    cu_q = torch.tensor([0, seq_len >> 1, seq_len], device=device, dtype=torch.int32)
    cu_k = torch.tensor([0, seq_len >> 1, seq_len], device=device, dtype=torch.int32)
    max_seqlen_q = int((cu_q[1:] - cu_q[:-1]).max())
    max_seqlen_k = int((cu_k[1:] - cu_k[:-1]).max())
    fa4_fwd_bf16 = lambda: FA4_Varlen(
        q[0], k[0], v[0],
[...]

PR #2889, opened 2026-09-14
[CuTe, SM90] hdim256 backward: C++ FA3 dQ path and 2-CTA DSMEM dS exchange (373 → 530 TFLOP/s)

SuperGoodGame:
## Summary

This PR takes the CuTe SM90 backward at head_dim 256 from 373 to 530 TFLOP/s (dense bf16, S=16k, H200), 1.42x, past the C++ FA3 kernel (480 TFLOP/s on the same GPU), with unchanged numerics. It is the follow-up to #2888, which fixed the swapped postprocess layout and moved hdim 256 to a 2-stage 64x48 config (302 → 373 TFLOP/s). The two optimization commits are independently useful:

1. **C++ FA3's hdim256 dQ path** (`BwdConfig.dQacc_use_TMA=False`). No smem staging of dQ: the MMA warp groups reduce the dQ accumulator straight into `dQaccum` with `red.global.add.v4.f32`, partitioned by the same R2S tiled copy as the smem path, so the gmem layout and the postprocess are unchanged. dQ is computed in two hdim halves into the same registers (C++ `Slice_dQKV_Mma`; with two accumulators ptxas hoists the second WGMMA above the atomics and dV spills at tile_n=80). The freed 64 KB is what lets the C++ tile (64x80, 2-stage Q) fit. 373 → 471 TFLOP/s, C++ FA3 is 480 on the same GPU.
2. **FA4's 2-CTA dS exchange, on Hopper via DSMEM** (`BwdConfig.cluster_size=2`). With the atomics compiled out the 1-CTA kernel runs at 747 TFLOP/s, so what remains is the fp32 reduce traffic into dQaccum (`S² · d · 4 B / tile_n` per head; a single SM sustains only ~30 GB/s of `red.add`). The two CTAs of a cluster take KV blocks n and n^1 of one head, walk the union of their m-block ranges in lockstep and exchange their dS tiles through distributed shared memory (`cp.async.bulk.shared::cluster`, mbarrier full/empty handshake with remote arrive/expect_tx). CTA r then computes `dQ[:, 128r:128r+128] = dS_own @ K_own[:, half] + dS_peer @ K_peer[:, half]` (K = 2·tile_n) and reduces only its half: half the atomics and half the dQ registers at the same tile_n. Splitting by head-dim halves rather than query rows (FA4's Blackwell variant) needs only the peer's K half (20 KB) in smem, which is what makes 64x80 fit exactly (232,448 B). Q and dO are TMA-multicast to the pair. Odd block counts use a "ghost" partner that participates in every barrier with fully masked columns (dS = 0), flushes its dQ half, and skips the dK/dV epilogue. 471 → 530 TFLOP/s.

The pair's fixed cost only pays off once the m-block loop is long enough, so `_tile_size_bwd_sm90` selects the 2-CTA kernel for `seqlen_q >= 2048` (max_seqlen_q for varlen) and keeps the 1-CTA 64x80 config below (measured crossover, table below). Deterministic and block-sparse backward keep the semaphore-ordered smem-staged path, matching the C++ restriction `Deterministic && kHeadDim < 256`. Forward and hdim ≤ 192 are untouched.

A third commit rounds `seqlen_k_rounded` straight to `n_block_size * cluster_size` instead of fixing up the parity of the n-block count. The old form branched on `max_seqlen_k`, which callers may pass as a device tensor (HF/TE varlen): that synced, and raised `DataDependentOutputException` in the FakeTensorMode compile pass. The rounding is the same value and does not branch. The SM100 2-CTA backward has the same issue; the SM90 cluster config added here reaches it from the varlen path.

New surface: two `BwdConfig` fields (`dQacc_use_TMA`, `cluster_size`) and two `FlashAttentionBackwardSm90` constructor arguments of the same names, both selected in `_tile_size_bwd_sm90` (which gains `deterministic` and `seqlen_q` keyword arguments). No public API change.

## Results

H200 (132 SMs, 227 KB smem; CUDA 12.8, torch 2.10, nvidia-cutlass-dsl 4.7.1), bf16, backward, ms · TFLOP/s:

| | `main` (before #2888) | #2888 | commit 1 (1-CTA) | commit 2 (2-CTA) | C++ FA3 |
|---|---|---|---|---|---|
| dense S=16384 H=16 D=256 | 36.5 · 302 | 29.5 · 373 | 23.3 · 471 | **20.8 · 530** | 22.9 · 480 |
| dense causal S=16384 | 17.2 · 320 | 14.8 · 370 | 11.9 · 464 | **10.5 · 522** | 11.7 · 471 |
| varlen S=86000 H=32 (#2878) | 1931 · 314 | 1672 · 362 | 1314 · 461 | **1250 · 485** | 1277 · 474 |
| varlen causal S=86000 H=32 | — | 790 · 383 | 659 · 460 | **568 · 534** | 645 · 470 |
| dense S=65536 H=16 | 567 · 310 | 480 · 367 | 383 · 459 | **364 · 483** | 371 · 475 |

Other head dims are untouched (the same kernels compile; hdim 192 GQA works because of #2888). Same GPU, `main` and this branch back to back, S=16384 H=16, dense / causal:

| D | `main` | this PR |
|---|---|---|
| 128 | 8.86 / 4.48 ms · 620 / 614 TFLOP/s | 8.53 / 4.36 ms · 645 / 631 |
| 192 | 15.72 / 7.93 ms · 525 / 520 | 15.46 / 7.73 ms · 533 / 534 |
| 64 | 4.84 / 2.36 ms · 568 / 583 | 4.70 / 2.36 ms · 584 / 583 |

2-CTA vs 1-CTA (same 64x80 tile, H=16, ms ratio; `bench/crossover_*.log`):

| seqlen | 512 | 1024 | 2048 | 4096 | 8192 | 16384 |
|---|---|---|---|---|---|---|
| dense | 1.02 | 1.03 | 0.91 | 0.91 | 0.92 | 0.83 |
| causal | 1.02 | 1.02 | 0.95 | 0.92 | 0.93 | 0.88 |
| local (1024,1024) | 1.00 | 0.97 | 0.90 | 0.89 | 0.89 | 0.91 |
| local (256,0) | 1.02 | 1.01 | 1.00 | 1.00 | 1.00 | 1.01 |

## Correctness

- Max error vs an fp32 reference is unchanged from `main` on every path (S=2048: dq/dk/dv 1.33e-3 / 1.16e-3 / 9.14e-4 non-causal, 1.20e-2 / 1.21e-2 / 1.58e-2 causal). dQ stays fp32-accumulated; only the reduction order changes, as in the C++ kernel.
- Checked vs the fp32 reference on both the 1-CTA and the 2-CTA kernel: dense, causal, GQA 16/4 and 16/2, MQA, local (200,0)/(100,50)/(256,0)/(1024,1024), softcap, deterministic (bitwise-identical repeat, MHA and GQA; smem path), hdim_v=128, hdim 200/224/240 (MHA and GQA, per kv head), odd n-block counts (ghost CTA), ragged S=1000, S=4096, varlen S=86000.
- Varlen with the 2-CTA kernel, per batch vs the fp32 reference (`bench/varlen_gqa_cluster.py`): GQA/MQA/MHA with batches of 0, 1, odd and many KV blocks, causal and not, identical errors to the 1-CTA kernel (this caught a padded-offset mismatch between the kernel and the dK/dV accumulator layout during review; fixed in commit 2).
- Varlen callers that pass `max_seqlen` as a device tensor: same gradients as the host-int call, and the FakeTensorMode compile pass goes through (commit 3).
- `tests/cute/test_flash_attn.py::test_flash_attn_output -k -256-` (8560 ids: every head_dim 256 case and every seqlen 256 case, mha/mqa/gqa × causal × local × softcap × deterministic × sink): 5246 passed, 3312 skipped, 2 failed. Both failures are the `dSink` tolerance check at `[256-256-128-...-mqa]` (diff 0.0018 vs tolerance 0.0011, head_dim 128) and fail identically with `main`'s `flash_bwd_sm90.py` / `interface.py` on this GPU. `test_flash_attn_varlen_output` d=256 random sample (320 ids): 196 passed, 124 skipped, 0 failed. `test_flash_attn_sm90_hdim_padding*` and the empty-workload backward tests: 170 passed.
- `ruff check` / `ruff format --check` (0.11.13) clean on `flash_bwd_sm90.py`.

Related: #2878, #2888.


SuperGoodGame:
@codex review

PR #2891, opened 2026-09-16
Support seqused_q/seqused_k in SM100 HD256 backward

Mellonta:
## Summary
Add seqused_q/seqused_k support to Blackwell HD256 backward
Transformer Engine’s THD context-parallel KV all-gather path passes these arguments to FA4 backward in [context_parallel.py](https://github.com/NVIDIA/TransformerEngine/blob/110791f2dfa006335187943414c4fc9e1b03ef0f/transformer_engine/pytorch/attention/dot_product_attention/context_parallel.py#L4395-L4396).

## Test
```
FLASH_ATTENTION_CUTE_DSL_CACHE_ENABLED=1 pytest tests/cute/test_hd256_varlen_scheduler.py -k seqused_backward -x
```


Johnsonms (approved):
LGTM. Separating the physical offset (`cuseqlen_*`, from `cu_seqlens`) from the effective length (`seqlen_*_cur_batch`, overridable by `seqused`) is exactly what the hd256 forward already does, and every length-dependent site in both kernels (trip bounds, masks, LSE/sum_OdO predicates, dQ/dK/dV store bounds, `dQ_epilogue_write_zero`) already reads the effective length. The `is_valid_q` widening is needed so all four warp roles agree on `has_work` in the dense + `seqused_q` case. This also matches how TE's THD CP path calls FA4 (padded `cu_seqlens_*` + real `seqused_*`, pre-zeroed grads) — your `packed`/`qk` test case.

Validated on 8×B200 at `824fa31c`:
- `test_hd256_varlen_scheduler.py`: 101 passed / 0 failed.
- Re-enabled `test_flash_attn_varlen_output` d=256 `seqused` combinations incl. zero-length batches: 144 passed / 0 failed (48 skips are the pre-existing `unpad_q=False, unpad_kv=True` skip).
- No default-path regression: dense hd256 fwd/bwd/preprocess cubins are byte-identical to base; the packed (`cu_seqlens`-only) bwd cubin differs only by instruction order (same PTX/SASS instruction counts), timings within noise.
- Merges cleanly onto current `main`.

One note, non-blocking: the hd256 bwd can now read rows that belong to no sequence, so like the generic SM100 bwd it needs padding rows to be finite (neither path trims TMA loads by `seqused`; `0 × NaN` would propagate). Maybe worth a one-line comment.

PR #2894, opened 2026-09-16
ci: registry-free GPU job (runner-local SIF) + verify the provisioned overlay from a fresh session

Johnsonms:
Two CI infra fixes that interact, in one PR.

## Registry-free GPU job

The job logged in to Docker Hub on every run to pull a private image. With the credentials rotated, the
unconditional login step fails before any test runs, although the SIF is cached and the pull is skipped.
The image is just Ubuntu + Python + torch; `run_fa4_ci.py` installs cutlass-dsl / quack / FA4 at job time.

- `gpu-test`: new input `fa4_local_sif` — use that SIF, skip pull and prune, fail if the file is missing.
- `ci.yml`: Docker Hub login only when `vars.FA4_LOCAL_SIF` is empty. Unset = unchanged behaviour.
- `tools/ci/docker/build_local_sif.sh`: `docker build` the existing Dockerfile on the runner and
  `apptainer build` a SIF into `CI_WORK_DIR/local-sif/` (registry mode prunes `*.sif` directly under
  `CI_WORK_DIR`, hence the subdirectory).

## Verify the provisioned overlay from a fresh session

All 8 tests failed with `JIT session error: Symbols not found: [ CuteDSLRT_TVMFFISetRaisedCudaError ]`
after a passing floor check. The test session saw the SIF-baked cutlass-dsl 4.6.0 runtime (`lib/`) next to
the installed 4.8.0 (`cu13/lib/`): `apptainer exec` returns before its fuse2fs child has flushed the overlay
image, so the next session can read a stale view, and DSL discovery tries `lib/` first. Intermittent and
load-dependent; latent since #2685, fatal once #2798 moved the pin off the SIF's layout. The floor check
runs in the same session as the install, so it cannot see this.

`tools/ci/run_fa4_ci.py`: wait until nothing holds the overlay image between sessions; verify from a fresh
session (one distribution per package, `flash_attn.cute` from the checkout, runtime owned by the installed
cutlass-dsl), re-provision once, then fail with `OVERLAY VERIFY FAILED: ...`; pin the verified runtime via
`CUTE_DSL_LIBS`.

## Validation

- Action pull step run locally in all three states (local SIF selected / missing file → exit 1 / variable
  unset → old cached-SIF path).
- Verification fails on an `--overlay img:ro` mount (whiteouts ignored, i.e. the leaked state), passes on RW.
- End-to-end with the `ci.yml` filter on the current CI SIF (DSL 4.6.0 baked) and on a
  `build_local_sif.sh` image (4.8.0 baked): verify OK, Pass 1 8/8, Pass 2 8/8 on both.

Related: #2685, #2798. Supersedes #2895.


tools/ci/run_fa4_ci.py
```diff
@@ -178,12 +179,148 @@ def prepare_overlay(
...
+    start = time.monotonic()
+    while True:
+        holders = _pids_holding(real)
+        waited = time.monotonic() - start
+        if not holders:
+            if waited > 0.5:
+                print(f"(overlay hand-off: waited {waited:.1f}s for the previous session to release {overlay})")
+            return waited
+        if waited > timeout_s:
+            print(f"WARNING: {overlay} still held by pid(s) {holders} after {timeout_s:.0f}s; continuing")
```
drisspg: should we just fail on timeout?

tools/ci/run_fa4_ci.py
```diff
@@ -178,12 +179,148 @@ def prepare_overlay(
...
+    sys.exit(1)
+print(f"overlay verify OK: FA4 from {os.path.dirname(iface.__file__)}, runtime {runtimes[0]}")
+print("CUTE_DSL_LIBS=" + os.pathsep.join(libs))
+"""
+
+
+def verify_overlay(repo_root: Path, base_env: dict[str, str], sif: str, work_dir: str, overlay: str) -> str | None:
+    """Check the provisioned overlay from a fresh session; return CUTE_DSL_LIBS to pin, or None."""
+    print("=== Verify overlay (fresh session) ===")
+    inner = f"cd /tmp && python3 -c {shlex.quote(_VERIFY_OVERLAY_PY)} {shlex.quote(str(repo_root))}"
```
drisspg: Could we rerun assert_dsl_floor.py in this fresh session, the version check currently only runs in the provisioning session.

drisspg (approved):

tools/ci/run_fa4_ci.py
Johnsonms: Agreed — changed in abe60189: `wait_for_overlay_release()` now raises after the timeout instead of warning and continuing. Opening another session on a still-held image is exactly the stale-view case the wait guards against.

tools/ci/run_fa4_ci.py
Johnsonms: Done in abe60189: `verify_overlay()` now runs `assert_dsl_floor.py` in the fresh session before the consistency checks, so the floors are checked against what the test sessions actually see (the log shows the floor check twice: provisioning + verify).

PR #2895, opened 2026-09-16
ci: verify the provisioned overlay from a fresh session and pin the DSL runtime

Johnsonms:
## Problem

Every GPU job on `main` failed on 2026-09-16, all 8 tests at `cute.compile(..., "--enable-tvm-ffi")`:

```
JIT session error: Symbols not found: [ CuteDSLRT_TVMFFISetRaisedCudaError ]
```

right after provisioning had reported `DSL floor check OK: nvidia-cutlass-dsl=4.8.0.dev0`.

Inside the failing test container the SIF-baked cutlass-dsl 4.6.0 runtime (`nvidia_cutlass_dsl/lib/`) was
visible next to the freshly installed 4.8.0 (`cu13/lib/`), although `prepare_overlay()` had deleted it and the
overlay image's upper layer on disk had the correct whiteouts and opaque marker. `apptainer exec` returns
before the fuse2fs that owns the ext3 image has finished flushing, so a session opened right after can read a
stale view. DSL runtime discovery tries `lib/` before `cu{major}/lib/`, so the stale 4.6.0 runtime was linked.
Intermittent and load-dependent. Latent since #2685; fatal once #2798 moved the pin from `==4.6.0.dev0`
(same layout as the SIF) to `>=4.6.2` (new `cu13/` layout).

The floor check runs in the same session as the install, where the in-memory overlay state is always right,
so it cannot see a cross-session inconsistency.

## Change (`tools/ci/run_fa4_ci.py`)

1. `wait_for_overlay_release()`: after every session, wait until no process holds the overlay image
   (60 s cap, then warn). `sync` at the end of provisioning.
2. `verify_overlay()`: a fresh session asserts one visible distribution per runtime package, `flash_attn.cute`
   imported from the checkout, and the `libcute_dsl_runtime` resolved by `get_shared_libs()` owned by the
   installed cutlass-dsl. On failure: recreate + re-provision once, then fail with `OVERLAY VERIFY FAILED: ...`
   naming what leaked, instead of 8 opaque JIT link errors.
3. Pin the verified runtime into the test sessions via `CUTE_DSL_LIBS`.

`tools/ci/README.md` documents the step.

## Validation

- The verification snippet fails on an `--overlay img:ro` mount (read-only mounts ignore whiteouts, i.e. the
  leaked state): 5 duplicate distributions and `flash_attn.cute` resolving to the baked copy. Passes on RW.
- End-to-end with the `ci.yml` filter on the current CI SIF (bakes DSL 4.6.0) and on a fresh
  `tools/ci/docker/Dockerfile` build (bakes 4.8.0): verify OK, Pass 1 8/8, Pass 2 8/8 on both.

Related: #2685, #2798, #2894.


Johnsonms:
Folded into #2894 (same commit, cherry-picked onto that branch): the two changes interact — the registry-mode pull step prunes `*.sif` under `CI_WORK_DIR`, which is where the local SIF was going to live — so reviewing them together is clearer. Closing this one.

PR #2896, opened 2026-09-17
[CuTe, SM90] Fix wrong dQ in backward for head_dims padded to a non-multiple of 32

arvin-chou:
## Description

The SM90 CuTe backward sets `hdim_multiple_of = 64 if dKV_swapAB else 16`. With the default (non-swapped) dK/dV layout, head_dim 72/104 therefore round `tile_hdim` up to only 80/112, while interface.py allocates and zeroes `dq_accum` with head_dim rounded to 32 (96/128), and the preprocess/postprocess steps use the same granularity. The kernel never writes the tail of the padded accumulator (columns 80-95 / 112-127), and the postprocess readback carries that stale data into dQ. Forward, dK and dV are unaffected; dQ comes back silently wrong with a uniform four-mismatches-per-column fingerprint.

### Root cause

The kernel-level tiling granularity (16) disagrees with the interface-level allocation granularity (32) for the non-swapped layout.

Use 32 for the non-swapped layout as well, matching the interface allocation. One-line change; a no-op for head_dims that are already multiples of 32, and the swapAB path is unchanged.

## Examples

Repro cell (bf16, seqlen 512, 8 heads, head_dim 72, non-causal, varlen with a single sequence), dQ checked against an fp32 SDPA reference at atol=rtol=0.05:

```python
import torch
from flash_attn.cute.interface import flash_attn_varlen_func

torch.manual_seed(0)
q, k, v = (torch.randn(512, 8, 72, dtype=torch.bfloat16, device="cuda") for _ in range(3))
cu = torch.tensor([0, 512], dtype=torch.int32, device="cuda")
out = flash_attn_varlen_func(q, k, v, cu, cu, 512, 512, causal=False)
dq = torch.autograd.grad(out.float().pow(2).sum(), q, retain_graph=False)[0]
```

Affected head_dims: 72, 80, 104, 112 (16-aligned padding that is not a multiple of 32).

## Performance Data

No performance claim: this is a pure correctness fix. The change only alters tile padding granularity for the affected head_dims and is a no-op for head_dims that are already multiples of 32.

## Verification

A/B on H100 (SM90), bf16 inputs against an fp32 SDPA reference, head_dim in {72, 104} x causal in {False, True}, same seed/shapes/tolerances per cell. The stock build fails all four cells: three return dQ with more than half of all elements mismatched (uniform four-per-column fingerprint; fwd/dK/dV exact), and head_dim 104 non-causal additionally fails to compile with an MLIR partition error. With the patch all four cells pass, and every per-tensor max_abs_diff matches the same cells run on stock B100 (SM100), which does not have the bug. The analogous fix for the swapped layout landed upstream as [#2482](https://github.com/Dao-AILab/flash-attention/pull/2482).


LiRunGuo:
Independent validation on **H200 (SM90)**, main @ edb5c76, nvidia-cutlass-dsl 4.8.0, torch 2.11.0+cu128, bf16:

- Stock main: `tests/cute/test_flash_attn.py::test_flash_attn_varlen_backward_non_aligned_head_dim` fails all 4 cases. hdim 72/104 causal and non-causal show 53-57% of grad elements mismatched; hdim 104 non-causal fails to compile with `failed to partition '(112,80)' with tiled_mma`.
- With this one-line change: `-k "non_aligned_head_dim or sm90_hdim_padding"` plus `tests/cute/test_flash_attn_fast.py` gives **347 passed, 0 failed** (compile cache disabled, so every kernel was freshly JIT-compiled).

FWIW, FA4 CI only runs on B200, so SM90 regressions like this one aren't caught there.

PR #2899, opened 2026-09-19
Compile with c++20 for pytorch 2.13+

cih9088:
## Motivation
https://github.com/Dao-AILab/flash-attention/pull/2879 introduced compiling with `c++20` for pytorch2.14+ based on [header guard](https://github.com/pytorch/pytorch/pull/178150/changes) but pytorch2.13 has `c++20` feature as well.

For example, the first compilation error encountered while building flash attention2 with pytorch2.13 is [here](https://github.com/pytorch/pytorch/blob/release/2.13/c10/core/AutogradState.h#L88-L89) which is [default member initialisers for bit-fields](https://en.cppreference.com/cpp/language/bit_field)
> There are no [default member initializers](https://en.cppreference.com/cpp/language/data_members#Member_initialization) for bit-fields: `int b : 1 = 0;` and `int b : 1 {0}` are ill-formed.
(until C++20)

## Changes
- changed to compile the flash attention with `c++20` if pytorch version is greater or equal to `2.13`.


Johnsonms (approved):
Thanks @cih9088 

PyTorch 2.13 introduced C++20-only bit-field default initializers in `c10/core/AutogradState.h` (pytorch/pytorch#180767). GCC, Clang, and NVCC accept them under C++17 with warnings, but MSVC and `-Werror` builds fail. My test confirms:

* PyTorch 2.12 + C++17: passes
* PyTorch 2.13 + C++17 + `-Werror`: fails
* PyTorch 2.13 + C++20: passes without warnings

So 2.13 is the correct threshold, consistent with `torch.utils.cpp_extension`, which has used C++20 since 2.12.

Optional wording:

> PyTorch 2.13+ headers use C++20-only syntax (bit-field default member initializers in `c10/core/AutogradState.h`, pytorch#180767). GCC/Clang/NVCC only warn under C++17, while MSVC and `-Werror` builds fail. PyTorch 2.14+ additionally emits `#error` from `ATen/ATen.h`.

**Which compiler** produced the hard error—MSVC or a `-Werror` build? It may be worth noting in the PR description.

Also, #2872 touches the same block and should adopt this threshold after rebasing.

cih9088:
@Johnsonms I dug into the issue and here is what I have found.

When I created this PR, I didn't compile FA2 from the main but from the pypi.
```console
FLASH_ATTENTION_FORCE_BUILD=TRUE pip install flash-attn --no-build-isolation --force-reinstall --no-deps
```
and it throws the compilation error I mentioned.

After seeing your comment, I also confirmed the following.
- PyTorch 2.13 + C++17 (pypi): fails
- PyTorch 2.13 + C++17 (main before the PR): passes
- PyTorch 2.13 + C++17 + -Werror (main before the PR): fails

which made me wonder what have been changed since the source was published to pypi?

It turns out that https://github.com/Dao-AILab/flash-attention/pull/2669 was the source of the compilation success. https://github.com/Dao-AILab/flash-attention/pull/2669 removed `#include <ATen/cuda/CUDAGeneratorImpl.h>` (`AutogradState.h`) from `flash.h`. CUDA files included `flash.h` header file so after https://github.com/Dao-AILab/flash-attention/pull/2669, NVCC no longer needs to preprocess `AutogradState.h`.


NVCC with `-std=c++17` does not treat the bit-field default member initialiser as an extension and raises the error, on the other hand GCC with `-std=c++17` does treat it as an extension.

But doesn't NVCC use GCC or whatever under the hood? It is indeed, NVCC uses GCC under the hood to compile the host code but before that, NVCC needs to know which part of the code is host and device so it has its own parser to distinguish the host code from the device code. The compilation error is from that NVCC parser before even GCC gets a change to compile.

In the end, the PR fixes when FA2 is compiled with `-Werror`.

FYI, here are versions I used.
GCC 11.4
NVCC 12.1

PR #2901, opened 2026-09-20
[ROCm/CK] Fix num_splits heuristic, which could never return more than 1

Johnsonms:
Cause
-----
override_num_splits_if_necessary() passed num_splits_heuristic_ck() two wrong inputs, leaving split-KV dead code: num_splits=0 was bit-identical to num_splits=1 on every shape measured.

1. num_n_blocks was ceildiv(hdim_v, kN1) with kN1 = hdim_v, i.e. a constant 1. It counts the KV-sequence blocks a split is carved from, so it must come from seqlen_k, as set_params_splitkv() does on the CUDA side. Pinned at 1, `max_splits = min(max_splits, num_SMs, num_n_blocks)` clamped to 1 and the heuristic could only ever return 1. seqlen_k was not even a parameter.

2. The parallelism term used nhead. The splitkv kernel packs the query heads sharing a KV head into one workgroup, so the grid is batch * nhead_k * num_m_blocks; using nhead overstated parallelism by the GQA ratio and suppressed splitting where it pays most. At a fixed batch*nhead of 4096 the gain from splitting tracks batch*nhead_k: 1.02x at 4096, 1.29x at 256, 1.95x at 128, 3.21x at 64.

Fix
---
Pass seqlen_k and nhead_k through and derive both quantities correctly. num_splits_heuristic_ck() itself is unchanged: given correct inputs its choices match, shape for shape, the best policy found by fitting occupancy- and blocks-per-split-based alternatives against measured optima.

The heuristic is also capped at 8 splits, and the `num_splits <= 128` checks become `<= 8`. The CK splitkv combine kernel only reduces up to 8 splits correctly: on gfx950 with CK c56c6750, num_splits 1..8 match an fp32 reference to ~5e-3 relative error while num_splits >= 9 diverge (rel err 0.28..2.0), independently of batch, seqlen_k and causal. Callers passing more than 8 get an error instead of silently wrong output. Lifting that limit in the combine kernel is left to a separate change; it is what holds the win to ~2.2x rather than ~7x, since 25 of the 93 shapes measured have their optimum above 8.

varlen fwd calls the same helper and is updated identically.

Correctness
-----------
Changing num_splits changes the split-K reduction, so this was checked against a full rebuild of the extension (2413 instances, gfx950, 0 errors) rather than by reasoning about the heuristic alone.

The existing suite carries the strongest signal. test_flash_attn_kvcache is already parametrized over num_splits=[1, 0] and over mha/mqa/gqa, d in {32, 59, 64, 80, 128, 256}, new_kv, paged and non-paged KV, rotary (fraction 0/0.5/1.0, interleaved or not), causal, local, alibi and batch_idx. Because 0 always collapsed to 1, none of those cases had ever reached the split-KV path. They do now:

| test                                   | result        |
| -------------------------------------- | ------------- |
| tests/test_flash_attn_ck.py -k kvcache  | 152064 passed |
| tests/test_flash_attn_ck.py -k varlen   | 56774 passed  |

On top of that, against an fp32 reference:

| check                                          | result                 |
| ---------------------------------------------- | ---------------------- |
| auto path (num_splits=0), 12 decode shapes      | rel err 0.0019..0.0036 |
| every split count 1..8, same shapes             | rel err <= 0.0041      |
| num_splits of 9, 64, 128                        | rejected, not computed |
| varlen vs the dense kernel on identical inputs  | rel err 0.00000        |

The rejection cases are the reason for the 8-split cap. Before this change the API accepted up to 128 and the combine kernel silently produced wrong output above 8; the divergence is abrupt (rel err ~5e-3 at 8, 2.0 at 9) and does not depend on batch, seqlen_k, num_n_blocks or causal.

The cap itself is measured on gfx950 with CK c56c6750 only. Whether other architectures or CK revisions share the same limit is untested here, so it is a named constant rather than something derived.

Performance
-----------
93 decode/short-q shapes on MI350X (gfx950), bf16, d=128, batch 1..128, seqlen_k 1K..128K, GQA ratios 1:1..64:1:

| metric                              | value  |
| ----------------------------------- | ------ |
| geomean speedup                     | 2.16x  |
| max speedup                         | 7.3x   |
| geomean vs best reachable at ns<=8  | 1.030x |
| shapes made slower                  | 0      |

Per shape, from the end-to-end run against the rebuilt extension (`auto` is num_splits=0, the path this patch changes; `ns=1` is the old behaviour):

| batch | seqlen_k | q/kv heads | ns=1     | auto     | speedup | rel err |
| ----- | -------- | ---------- | -------- | -------- | ------- | ------- |
| 2     | 131072   | 64/8       | 6.435 ms | 0.988 ms | 6.51x   | 0.0029  |
| 1     | 65536    | 64/8       | 3.237 ms | 0.519 ms | 6.23x   | 0.0030  |
| 4     | 16384    | 64/1       | 0.801 ms | 0.138 ms | 5.80x   | 0.0027  |
| 1     | 16384    | 64/8       | 0.815 ms | 0.144 ms | 5.66x   | 0.0028  |
| 16    | 16384    | 64/1       | 0.836 ms | 0.154 ms | 5.43x   | 0.0030  |
| 8     | 32768    | 64/8       | 1.655 ms | 0.354 ms | 4.67x   | 0.0036  |
| 4     | 4096     | 64/4       | 0.217 ms | 0.051 ms | 4.27x   | 0.0030  |
| 64    | 4096     | 64/1       | 0.243 ms | 0.076 ms | 3.18x   | 0.0029  |
| 1     | 4096     | 64/64      | 0.225 ms | 0.071 ms | 3.15x   | 0.0035  |
| 16    | 4096     | 64/8       | 0.247 ms | 0.114 ms | 2.17x   | 0.0019  |
| 128   | 1024     | 64/8       | 0.191 ms | 0.191 ms | 1.00x   | 0.0021  |
| 64    | 4096     | 64/8       | 0.322 ms | 0.325 ms | 0.99x   | 0.0035  |

The last two rows are shapes with enough parallelism already; the heuristic correctly leaves them at one split. rel err is against an fp32 reference.


micmelesse:
@rocking5566 Can you take a look at this?

rocking5566 (commented):
LGTM, but could you remove or shorten the comment?

Johnsonms:
> LGTM, but could you remove or shorten the comment?

Trimed

micmelesse (approved):

PR #2902, opened 2026-09-20
[CuTe, SM120] Reject backward configurations that exceed shared memory

adenzhou1350:
SM120 backward with equal head dimensions of 192 requests 115712 bytes of shared memory, exceeding the 101376-byte limit. Forward succeeds, but backward reaches `CUDADialectError(cudaErrorInvalidValue)` during compilation. CUTLASS prints the resource counts separately; they are not in the Python exception.

Check the actual shared-storage declaration before emitting the backward kernel and raise a `ValueError` containing the required bytes, limit, tile, and padded head dimensions. Using the declaration includes P/dS, LSE/dPsum, and alignment padding omitted by the existing, unused `can_implement()` estimate. The change only improves rejection of unsupported configurations; it does not add d192 backward support or change supported kernels.

Validation against upstream `edb5c76ee329b18ed95d1f7ea9aa522a1331ab7d`:

- Actual CuTe shared-storage compilation confirms 115712 bytes for d192 and the new `ValueError`; supported d64, d128, and d192/v128 declarations still compile.
- RTX 5090 / SM120: four rejection tests passed (FP16/BF16 × causal/noncausal), with PyTorch 2.11.0+cu130 and CUTLASS DSL 4.6.2.
- Python syntax checks and `git diff --check` passed.
- In separate fresh baseline/candidate processes, 12 supported cases each passed output and dQ/dK/dV comparison to FP64 (with a same-input, output-dtype-rounded FP32 reference): d64/d64, d128/d128, d192/v128 × FP16/BF16 × causal/noncausal, B1/S128/H2. The d192/d192 case reproduced the original CUDA error and the new descriptive `ValueError`, respectively.

Reproduce the new regression test with:

```sh
pytest -q tests/cute/test_flash_attn.py -k sm120_backward_rejects_smem_overflow
```

Related to #2609, which wires the existing `can_implement()` predicates into the interface. Its SM120 predicate still returns `True` for this d192 configuration because it counts 98304 bytes of Q/K/V/dO tiles but not the complete 115712-byte allocation. This small kernel-local check covers that gap without depending on the broader interface change.

AI-assisted implementation and testing. This is a validation/diagnostic fix, not a performance claim.


adenzhou1350:
@tridao, could you review this small SM120 backward guard when you have time? It checks the actual shared-storage size; the rejection tests and supported forward/gradient checks pass on an RTX 5090. Related to #2330 and #2609.

PR #2904, opened 2026-09-21
[CuTe, SM90] Skip the causal/local mask on unmasked KV blocks in forward

LiRunGuo:
### Motivation

On H200 the FA4 SM90 causal forward trails the C++ FA3 forward by 8-20% at every sequence length, while the non-causal forward is at parity. Fitting kernel time against `seqlen_k` shows the per-KV-block cost of the *non-causal* kernels is identical (FA3 1.626 us vs FA4 1.620 us per block per SM), so the causal gap has to come from the causal kernel itself.

### Cause

The SM90 mainloop splits the KV iterations into blocks that need masking (the diagonal, or the window edges for local) and blocks that do not. The second loop ("The remaining iterations have no masking") still passes `mask_fn`, which is bound with `mask_causal=self.is_causal` / `mask_local=self.is_local`. So every unmasked block re-applies the causal mask element-wise, even though it is all-true there. In the SASS of the hdim-128 causal kernel the unmasked inner loop has 580 instructions with 66 `FSEL` and 8 `R2P`, against 457 instructions / 2 `FSEL` / 0 `R2P` in the non-causal kernel.

`flash_fwd_sm100.py` already handles this loop correctly: it only passes `mask_fn` when `mask_mod` is set. This PR does the same on SM90. With the change the unmasked inner loop drops to 477 instructions, 2 `FSEL`, 0 `R2P`; the masked loop is unchanged.

### Performance (H200, bf16, `benchmarks`-style shapes with 32k total tokens, CUDA-event timing)

| forward | FA3 | FA4 main | FA4 this PR |
|---|---|---|---|
| hdim128 causal 1k | 509 | 414 | 431 |
| hdim128 causal 4k | 691 | 607 | 643 |
| hdim128 causal 8k | 679 | 637 | 667 |
| hdim128 causal 16k | 682 | 613 | **660** |
| hdim64 causal 4k | 481 | 410 | 466 |
| hdim64 causal 16k | 514 | 463 | **531** |
| hdim128 non-causal 16k | 685 | 665 | 665 |

TFLOPS. Non-causal is untouched (same kernel). The remaining gap at short sequence lengths is the single-tile (non-persistent) scheduler on SM90, which is a separate issue (#2754 territory).

### Correctness

- Forward `out` and `lse` are bitwise identical to main for 9 configurations: causal with sq==sk, sq<sk, sq>sk, GQA, hdim 64/128/192/256, local windows (256,0) and (300,100), and non-causal.
- `tests/cute/test_flash_attn_fast.py` + a `test_flash_attn.py` subset (sm90 hdim padding, small head dim, pack_gqa, lse_grad, `test_flash_attn_output` at 113x203 / 1024x1023 / 4224x4224): 2010 passed; the 4 failures are the same `dSink` tolerance failures as on main.
- `tests/cute/test_mask_mod.py`: 1256 passed, 120 failed, the identical set as main (fixed by #2903).
- `tests/cute/test_flash_attn_varlen.py -k "2048 and dtype0"`: 972 passed, 0 failed.

All test runs on H200 with the compile cache disabled or in a fresh cache dir.


LiRunGuo:
@tridao @reubenconducts could one of you review this? It's a small change in the SM90 forward mainloop:
- @tridao, you wrote most of `flash_fwd_sm90.py`;
- @reubenconducts, #2754 touches the same `mma()` loop.

Suggested order:

1. **Land this one first.** It only changes the `mask_fn=` argument of one `mma_one_n_block` call, forward `out`/`lse` are bitwise unchanged, and the non-causal kernel is untouched.
2. **Then rebase #2754 on top.** The two are complementary. This PR fixes the long-sequence part of the causal gap to FA3 (16k: 613 -> 660 TFLOPS on H200, FA3 682). The persistent scheduler in #2754 targets the short-sequence part (1k: 431 vs FA3 509). I can re-run the FA3/FA4 table on H200 for the combined branch.
3. **Follow-up from me: SM80.** `flash_fwd.py` (SM80) has the same pattern in its unmasked loop. I'd send the same one-line change for it separately, validated on A800.

cc @drisspg for merge scheduling. FA4 CI runs only on B200, so I'm happy to keep providing H200 numbers for SM90 PRs.

drisspg (approved):
looks good; thanks

PR #2906, opened 2026-09-22
[CuTe, SM100] Write real zeros for block-sparse empty tiles

pashu-cohere:
A Q tile with no live KV blocks skips its QK/PV sequence entirely. The PV GEMM
carries zero_init on its first iteration and is the only writer of the O
accumulator, so for an empty tile the O region in TMEM is never written --
tcgen05.alloc reserves the columns without initialising them.

The empty-tile path relied on calling correction_epilogue with scale 0.0 to
turn that residue into zeros. But 0.0 * NaN and 0.0 * Inf are both NaN, and
infinities are routine content for attention TMEM, so the tile yields zeros
most of the time and NaN whenever a prior kernel left an infinity behind.

Add a constexpr zero_fill to correction_epilogue so an empty tile fills the
fragment directly and never reads TMEM. Being constexpr, the non-empty call
site codegens exactly as before.

SM90 needs no equivalent flag: acc_O lives in registers and is already zeroed
inline in the mainloop. Forward counterpart to #2776.


pashu-cohere:
cc @drisspg @tridao

pashu-cohere:
Any comments @drisspg?

drisspg (approved):
Looks good thanks for the fix!

